Issue #511, opened 2014-04-24
render_form_fields should check widget instead of field type for hidden fields


SpotlightKid:
The `render_form_fields` template macro in `templates/admin/lib.html` checks whether a form field is a hidden input by comparing the field's `type` attribute against the hardcoded field type name `'HiddenField'` (lines 119 and 129).

IMHO it would be better to check the field's widget type instead, e.g.:

   {% for f in form if f.widget.input_type == 'hidden' %}

This allows to use custom sub-classes of `HiddenField` or any other field class and have them rendered as a hidden input field by just setting the field's widget to `HiddenInput()`. Currently they cause an orphaned field label to be rendered among the other non-hidden fields.

PR #522, opened 2014-05-02
Check hidden fields for widget type instead of specific class

SpotlightKid:
See ticket #511 for an explanation.

### DIFF
--- a/flask_admin/templates/bootstrap2/admin/lib.html
+++ b/flask_admin/templates/bootstrap2/admin/lib.html
@@ -116,7 +116,7 @@
         {% if csrf_token %}
         <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
         {% endif %}
-        {% for f in form if f.type == 'HiddenField' %}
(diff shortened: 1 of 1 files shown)


jgonggrijp:
I just spent several hours to figure out that commit 32068b0 had broken [my solution for an sortable order-preserving Select2Multiple field](http://stackoverflow.com/a/24928528/1166087) (for interest: read bottom part of answer, starting at "Addition: fixing the ordering"). The custom field uses a custom widget which inherits from wtforms.widgets.HiddenInput, but it's obviously supposed to appear with the other visible form rules and have a proper label and so on.

I was able to fix my widget with two line insertions, one to override `input_type` so it would be unequal to "hidden" and one to explicitly set `kwargs['type']` to "hidden" again in `__call`. I understand the reasoning behind 32068b0 and I don't think it should necessarily be reverted. However, I think the issue does illustrate how even small changes like these can break existing code. Changes like these should be mentioned in the changelog, in my opinion.

Issue #245, opened 2013-06-23
Translated column filter results in TypeError: keys must be a string


ozon:
I try to translate the column filter. I proceed as follows:

``` python
column_filters = ( Client.connections, )
...
column_labels = dict(
    connections=lazy_gettext(u'Connections'),
)
...

```

Flask give a Traceback with `TypeError: keys must be a string`

I tried to reproduce the problem with an flask-admin example.
https://gist.github.com/ozon/d05bf06fbdf08521b7df#file-gistfile1-py-L109
Traceback addet as comment.

It result in the same error. Without filter translations works the translated app fine. What do i wrong?

manuelbua:
I encountered the same problem solved here, but can't find the fix applied here in the latest version: how is this supposed to work now? I implemented a lazystring-aware custom JSON encoder, but i would like to know the current solution on this, anyone?

xqliu:
This issue is also happens on latest Flask-Admin (1.2.0), and seems the patch applied earlier is not working any more.

Because of this issue, it's not able to translate field labels in UI like below

```
    column_labels = {
        'supplier': lazy_gettext('supplier'),
        # 'name': lazy_gettext('supplier'),
        # 'code': lazy_gettext('supplier')
    }
```

Is there any update on this?

Thanks.

xqliu:
The issue might be related to fix commit via 14dec4f5396c094fce6c726d7d1b7e0afee62a83

xqliu:
Tried a fix 

Change

```
results[key] = items
```
on line 872 to 

```
results[as_unicode(key)] = items
```
Resolve the issue temporary.

I think a fix for both this issue and #840 is needed since this quick fix will revert fix to #840.

PR #525, opened 2014-05-04
fix the filter_groups keys non-string BUG

lazygunner:
I met the same problem as issue [#245](https://github.com/mrjoes/flask-admin/issues/245), and I notice the fix. But the new 'filter_groups' stuff in commit [d851c420019df6212d576f31301519cfb12104dc](https://github.com/mrjoes/flask-admin/commit/d851c420019df6212d576f31301519cfb12104dc) seems did nothing to it's non-string keys for 'tojson' function in the [template/admin/model/list.html ](https://github.com/mrjoes/flask-admin/blob/master/flask_admin/templates/bootstrap3/admin/model/list.html).

this is my PR based on the previous fix, hope you can find a better way to solve this, Thanks.

### DIFF
--- a/flask_admin/model/base.py
+++ b/flask_admin/model/base.py
@@ -14,7 +14,7 @@ from flask.ext.admin.actions import ActionsMixin
 from flask.ext.admin.helpers import get_form_data, validate_form_on_submit, get_redirect_target
 from flask.ext.admin.tools import rec_getattr
 from flask.ext.admin._backwards import ObsoleteAttr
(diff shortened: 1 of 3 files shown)


pawl:
Looks like @torotil is also trying to fix this issue in #994.

torotil:
@pawl I don't think this fix will make babel translations work: There is still the issue with _LazyString being non-hashable - so we can't use it as key for the BaseModelView._filter_group in the first place.

PR #617, opened 2014-08-08
Correct format_errors for nested dictionaries

cybertoast:
In the case of errors on EmbeddedDocuments, the format_errors() function had
an exception due to the error being returned as a dictionary, instead of a
string.


pawl:
Looks like mongoengine has something built-in for this: https://github.com/MongoEngine/mongoengine/blob/f8890ca841deb6b0f016e7fcce8bbe4d00246660/mongoengine/errors.py#L127

pawl:
I think this was just fixed by @gordol in #1262

PR #619, opened 2014-08-11
prefix blueprint name with admin to prevent conflicts... 

alanhamlett:
...with user-defined blueprint names

### DIFF
--- a/flask_admin/base.py
+++ b/flask_admin/base.py
@@ -237,7 +237,7 @@ class BaseView(with_metaclass(AdminViewMeta, BaseViewClass)):
             self.name = self._prettify_class_name(self.__class__.__name__)
 
         # Create blueprint and register rules
-        self.blueprint = Blueprint('flask_admin_%s' % self.endpoint, __name__,
+        self.blueprint = Blueprint('admin.%s' % self.endpoint, __name__,
                                    url_prefix=self.url,
                                    subdomain=self.admin.subdomain,
                                    template_folder=op.join('templates', self.admin.template_mode),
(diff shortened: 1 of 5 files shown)


jmagnusson:
This is a must have imo. It could easily happen otherwise that the generated names collide with an existing blueprint otherwise.

deniszgonjanin:
:+1: would be nice to have this.

jmagnusson:
@alanhamlett Could you update your PR to reflect the latest code?
@mrjoes? Would it be possible to get this merged in? It's quite easy to get name conflicts with your own blueprints.

mrjoes:
I'd say this pull request is no longer necessary. Now it is possible to override automatic names in `BaseView`: https://github.com/flask-admin/flask-admin/blob/master/flask_admin/base.py#L209

For example:
```python
class MyBaseView(BaseView):
    def _get_endpoint(self, endpoint):
        if endpoint:
            return endpoint

        return 'admin.%s' % self.__class__.__name__.lower()
```

jmagnusson:
Thanks @mrjoes. Doesn't it make sense to be able to namespace them either way though?

mrjoes:
Yeah, but it'll break backwards compatibility for everyone who uses `url_for('myadminview.index')`

jmagnusson:
Figures... :smile: Perhaps one could optionally supply a blueprint prefix to `init_app`?

```python
admin = Admin()
...
admin.init_app(myapp, blueprint_prefix='myadmin')
```

mrjoes:
Yeah, but one would have to define their own classes anyway (security, maybe global customizations, etc). So that's better place to have this logic in.

alanhamlett:
So this pull request isn't needed now?

jmagnusson:
@mrjoes with your example all the URLs will be changed as well. I just want to change the name of the resource, and keep the current URL. How would I do that?

mrjoes:
Good catch, automatically generated URLs use endpoint name.

I'll move URL scaffolding logic to a separate, overridable, method later today.

jmagnusson:
Great! :+1:

PR #644, opened 2014-09-09
Add optional page_size parameter to get_list

tandreas:
Add an optional page_size parameter to get_list.  

If the optional parameter page_size is passed, it will use this instead of the one set on the model view.  If the passed value is set to false, there will be no limit on get_list.

This will make it possible to get all rows (with filters applied) without changing self.page_size on the modelview.  This is necessary for something like a CSV export of all rows.


gstf:
+1

pawl:
Moved these commits into #981 and resolved the merge conflict.

PR #661, opened 2014-10-15
fix pymongo boolean filter issues in list view

Raz0r:
https://github.com/mrjoes/flask-admin/commit/a5648e3d819faaf49df2e62d8fe143ef95edfab1 introduced problems with pagination (?flt0_1=True instead of ?flt0_1=1 on links - filter reseted) and selected option  (it was always "Yes")


pawl:
Closing this in favor of #952

Issue #755, opened 2014-12-29
Idea: Editable List View


pawl:
It would be useful to be able to bulk edit records from the list view, like this:
![editable_listview2](https://cloud.githubusercontent.com/assets/992533/5567370/15760168-8f09-11e4-959d-a2f170ff0f5e.png)

I've made some progress on getting the form fields to show on the list view:
https://github.com/pawl/flask-admin/compare/list_editable

However, I'm having trouble with wtform's FieldList and processing the submitted data. Is there a way to do this that uses the existing inline form code? Or, does anyone have any suggestions for how to implement this?

Here's the gist I've been using to test my branch (with a new column_editable_list property): https://gist.github.com/pawl/ec6c51d1a10ffa75bac3

gordol:
Nice! I have been working on something very similar but using gevent socketio to push changes in real-time to other connected clients as well. Good stuff!

gordol:
Nice! I have been working on something very similar but using gevent socketio to push changes in real-time to other connected clients as well. Good stuff!

gordol:
Found one minor UI issue...

http://node.gordo4.com/snaps/37e652d9f501d90df8e156bdecda1e1e.png

gordol:
Perhaps bump down default number of textarea rows from 7 to say... 4 or 5.

gordol:
https://vitalets.github.io/x-editable/docs.html#textarea

karimtabet:
I've been having an issue using this when my tables primary key is a UUID with a "-" character. It seems to only read a sub-string of the UUID; from the beginning of the UUID to the first "-" character.

The bottom of the stack trace is:
```
  DataError: (DataError) invalid input syntax for uuid: "edcda51d"
  LINE 3: WHERE products.product_uuid = 'edcda51d'
                                      ^
'SELECT products.product_uuid AS products_product_uuid, products.title AS products_title,      products.short_description AS products_short_description, products.long_description AS products_long_description, products.last_seen AS products_last_seen, products.price AS products_price, products.rating AS products_rating \nFROM products \nWHERE products.product_uuid = %(param_1)s' {'param_1': u'edcda51d'}
```
The UI goes a bit crazy too:
https://i.imgur.com/vAAUMzB.png

pawl:
The issue is here: https://github.com/flask-admin/flask-admin/blob/a964d3fff65ebbc8ab31ea4f8c3e8be1a5b9b296/flask_admin/model/fields.py#L146

[...]

PR #756, opened 2015-01-02
Add editable list view

pawl:
This pull request adds the new editable list view described in issue #755.

A demo is available here: http://104.236.24.168:5065/admin/model1/

It supports SQLA, MongoEngine, and Peewee. And, it also has tests and CSRF support.

most_recent


mrjoes:
I'll go over the pull request tomorrow. Thank you!

pawl:
Just fixed an issue with the new pull request + CSRF on mongoengine and peewee.

Here are gists of the code I'm using to test: https://gist.github.com/pawl/5f28c3af3c757551913b

pawl:
Also found and fixed an issue that occurred when other values on the same row had a "Required()" validator.

flask_admin/contrib/mongoengine/view.py
```diff
@@ -545,6 +565,37 @@ def update_model(self, form, model):
...
+            :param form:
+                Form instance
+        """
+        try:
+            model = self.model()
+            for field in form:
+                # FieldList's last_index will only be set if a field is submitted
+                # last_index will be the primary key of the updated record
+                if getattr(field, 'last_index', None):
+                    record = self.get_one(field.last_index)
```
mrjoes: What if record is not found?

flask_admin/contrib/peewee/view.py
```diff
@@ -382,6 +402,40 @@ def update_model(self, form, model):
...
+            :param form:
+                Form instance
+        """
+        try:
+            model = self.model()
+            for field in form:
+                # FieldList's last_index will only be set if a field is submitted
+                # last_index will be the primary key of the updated record
+                if getattr(field, 'last_index', None):
+                    record = self.get_one(field.last_index)
```
mrjoes: Same as above.

flask_admin/contrib/sqla/view.py
```diff
@@ -611,6 +614,27 @@ def scaffold_form(self):
...
+            The form is created using the existing get_form(),
+            but each field is wrapped in a WTForms FieldList. 
+        """
+        converter = self.model_form_converter(self.session, self)
+        form_class = form.get_form(self.model, converter,
+                                   base_class=self.form_base_class,
+                                   only=self.column_editable_list)
+        
+        # iterate FormMeta to get unbound fields
+        field_dict = {}
```
mrjoes: Was thinking if we can move this common code into some helper function that's called by all backends?

flask_admin/model/base.py
```diff
@@ -1242,11 +1291,45 @@ def _create_ajax_loader(self, name, options):
         raise NotImplementedError()
 
     # Views
-    @expose('/')
+    @expose('/', methods=('POST', 'GET'))
```
mrjoes: Quick question: form is submitted via AJAX POST or a normal form POST?

If it is AJAX POST, it is better to use separate endpoint. If it is normal post - can we use AJAX POST?

flask_admin/model/widgets.py
```diff
@@ -61,3 +61,81 @@ def __call__(self, field, **kwargs):
...
+        if not kwargs.get('pk'):
+            raise Exception('pk required')
+        kwargs['data-pk'] = str(kwargs.pop("pk"))
+
+        kwargs['data-csrf'] = kwargs.pop("csrf", "")
+
+        # get first entry from FieldList to determine field type
+        subfield = field.entries[0]
+
+        if subfield.type == 'StringField':
```
mrjoes: Can we do it via the mapping with separate functions returning various settings?

This way if can be configurable for custom fields.

pawl: I've been thinking about this too.

How about I made a form_args and form_widget_args used only in the editable list view form and widget? I could possibly call it "list_form_args" and "list_form_widget_args".

mrjoes: Probably not necessary - can be property of a widget. If someone will want to customize it, they can always subclass widget and override it.

pawl: What if someone is using a custom validator like "date must be greater than today"?

I left "field_args=self.form_args" out of scaffold_list_form: https://github.com/mrjoes/flask-admin/pull/756#discussion-diff-22463150R625

I figured it may cause some issues with how the field is displayed if the same form_args are applied, but I can test to see if this is the case.

pawl: Can you show me an example of what you mean by "do it via the mapping with separate functions returning various settings"?

I think I'm done implementing all the other suggestions except this one.

flask_admin/contrib/mongoengine/view.py
pawl: I ended up completely getting rid of this function. I found a way to use `update_model` instead of creating a new function.

flask_admin/contrib/sqla/view.py
pawl: Here's the helper function I created: https://github.com/mrjoes/flask-admin/pull/756/files#diff-e1ef986502cb727b91e5215c566f57c6R17

I noticed `scaffold_list_forms` needs to be implemented in each backend's view.py because get_form is implemented differently for each.

flask_admin/model/base.py
pawl: Here's the new endpoint: https://github.com/mrjoes/flask-admin/pull/756/files#diff-401640c72e6a6b2a48e711090af0534cR1494

flask_admin/model/widgets.py
mrjoes: Simplest way: have a method that calls other methods. If someone will want to change logic, then can override one of the methods.

ArtemSerga:
This is an incredibly cool feature. Applause!

flask_admin/model/widgets.py
pawl: I just pushed the change. Is this what you had in mind?: https://github.com/pawl/flask-admin/blob/e7d821abc1936917098ca4e16190dc6ef114d555/flask_admin/model/widgets.py#L101

flask_admin/model/widgets.py
```diff
@@ -61,3 +61,112 @@ def __call__(self, field, **kwargs):
...
+            raise Exception('pk required')
+        kwargs['data-pk'] = str(kwargs.pop("pk"))
+
+        kwargs['data-csrf'] = kwargs.pop("csrf", "")
+
+        # gets the first entry (subfield) from FieldList (field) and uses the
+        # subfield.type (`kwargs_` method) to determine kwargs
+        subfield = field.entries[0]
+        try:
+            kwargs = getattr(self, 'kwargs_' + subfield.type)(subfield, kwargs)
```
mrjoes: I was thinking about method that'll accept subfield and return `kwargs`. Inside it can have bunch of ifs, call child methods like you did, etc. As long as this method is overridable, it should do the trick.

pawl: I made the change you suggested. Now, to override the new `get_kwargs` method from the ModelView class, the developer would need to override `scaffold_list_form` and `wrap_fields_in_fieldlist` (to use a different field), then create the new field which uses a different widget.

It's not as easy as overriding other form fields, but I will plan to do a pull request in the future to fix that.

gordol:
anyone interested in support for tabbing through editable fields?

i've got something I've been working on for quick editing of listviews. but it did require a small hack to x-editable bootstrap code. i had to change triggerhandler to trigger, so the events would properly bubble up. that might not be needed here though, since in flask-admin we're serving the editable elements from server-side, but in my app i'm loading the data over websockets and doing data-binding.

anyway, if anyone is interested, let me know and i can post some example code soon. seems like it might be useful here. :)

pawl:
When you say "tabbing through editable fields", you mean every time you press tab it switches to the next field?

The most recent changes should make this easier: https://github.com/mrjoes/flask-admin/blob/master/flask_admin/model/base.py#L893

You can override the widget with your custom get_kwargs, then you could add the javascript code to form.js.

gordol:
> When you say "tabbing through editable fields", you mean every time you press tab it switches to the next field?

Yep.

PR #766, opened 2015-01-19
fix index_view csrf value rendering

pawl:
This fixes an issue caused by my pull request #763. ```{{ delete_form.csrf_token() }}``` was rendering the entire field rather than just the value.

I made sure this works for both flask_wtf and wtforms2.

### DIFF
--- a/flask_admin/templates/bootstrap3/admin/model/list.html
+++ b/flask_admin/templates/bootstrap3/admin/model/list.html
@@ -111,7 +111,7 @@
                             <input type="hidden" name="id" value="{{ get_pk_value(row) }}"/>
                             <input type="hidden" name="url" value="{{ return_url }}"/>
                             {% if delete_form.csrf_token %}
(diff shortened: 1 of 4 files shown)


flask_admin/templates/bootstrap3/admin/model/list.html
```diff
@@ -111,7 +111,7 @@
                             <input type="hidden" name="id" value="{{ get_pk_value(row) }}"/>
                             <input type="hidden" name="url" value="{{ return_url }}"/>
                             {% if delete_form.csrf_token %}
-                            <input type="hidden" name="csrf_token" value="{{ delete_form.csrf_token() }}"/>
+                            <input name="csrf_token" type="hidden" value="{{ delete_form.csrf_token() }}"/>
```
pawl: I just started my bootstrap3 application with the changes and noticed I forgot to change ```delete_form.csrf_token()``` to ```delete_form.csrf_token._value()``` like in the bootstrap2 version. I'm going to submit a pull request.

PR #767, opened 2015-01-19
Allow easier overriding of editable list view

pawl:
I added instructions for how to override the new editable list view FieldList's widget and made it easier to override.

This would be useful if someone didn't like the default kwargs set by XEditableWidget, or wanted to add kwargs for a custom field.

Also, the editable list view form now shares validators with the form through form_args. But, none of the other form_args are used by the editable list view.


flask_admin/model/base.py
```diff
@@ -853,10 +854,18 @@ def scaffold_form(self):
         """
         raise NotImplementedError('Please implement scaffold_form method')
 
-    def scaffold_list_form(self):
+    def scaffold_list_form(self, CustomFieldList, validators=None):
```
mrjoes: Is it possible to provide default value for the `CustomFieldList`? Also, lets use underscore variable name even if it is a class value.

pawl: Ok, I changed `CustomFieldList` to `custom_fieldlist` and added a default value.

PR #780, opened 2015-01-27
Added _search method for mongoengine and pymongo backends for easier override

gstf:


rturk:
Sounds interesting can you also add Documentation and Examples?

PR #795, opened 2015-02-22
Allow searching non-text SQLAlchemy column types

pawl:
I tried this with SQLAlchemy 0.7 and 0.9.8 on both sqlite and mysql. I also created a unit test for searching integer columns.

The reason for this change? It's more user friendly to allow searching integer columns too. In my application, I have a ModelView that shows usernames and a related number. Rather than requiring the user to create a filter for the numbers or do a search for the names - it's easier just to use the search for both.

### DIFF
--- a/flask_admin/contrib/sqla/view.py
+++ b/flask_admin/contrib/sqla/view.py
@@ -80,8 +80,7 @@ class ModelView(BaseModelView):
                                           'searchable_columns',
                                           None)
     """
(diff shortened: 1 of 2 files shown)


pawl:
I can make more tests for this if you want. I also tried searching on date and datetime columns, and it appeared to work. It doesn't seem like SQLAlchemy's ILIKE has any restrictions on column type?

PR #796, opened 2015-02-22
Add AppEngine ModelViews for both DB and NDB.

mikelambert:
Tested it works on flask_admin running on my local dev_appserver, and works much better than the built-in appengine admin site. More complex features still remain TBD (queries, as well as nested keys, and probably other obscure features)


pawl:
@mikelambert Would you mind converting tabs to spaces in this code and submitting a pull request?

An usage example would be nice too. Maybe something like: https://github.com/flask-admin/flask-admin/blob/master/examples/sqla/app2.py

Issue #357, opened 2013-11-03
Deleting all elements from a SelectMultipleField list does not work


cybertoast:
I'll write a test case for this, but wanted to report it as well ...
Env: contrib.mongo

Steps:
- Add 1 or more elements to a SelectMultipleList in a model that has a db.ListField
- Save
- Delete all the elements in the list field one at a time and save
- Notice that the last element is not deleted

This is true for both string and referenced values in a list.

gordol:
Is this on anyone's roadmap to fix?

gordol:
I updated my fix for this here: https://github.com/flask-admin/flask-admin/pull/586/ so it should merge cleanly now.

ElLorans:
Closing this issue due to inactivity and lack of progress. If there's still interest or new context to add, feel free to reopen or open a new issue with updated details.

PR #798, opened 2015-02-26
Fix: allow clearing multiple-select fields (Issue #357)

jnga:
This is a fix for #357, where multiple-select fields cannot be cleared.

If a multiple-select form field is cleared of its items, it no longer has any selections, so the browser will not even include that field in the POST data. Due to this absence, the model data are not cleared as expected, and the items reappear in the form field.

This fix detects any multiple-select fields that the user has cleared, then clears the corresponding model data. It has been tested with MongoEngine.

### DIFF
--- a/flask_admin/model/base.py
+++ b/flask_admin/model/base.py
@@ -1245,7 +1245,7 @@ class BaseModelView(BaseView, ActionsMixin):
                 form._fields[field].raw_data == []]
 
         # Clear the corresponding model data for each of these fields.
-        if model:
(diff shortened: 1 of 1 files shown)


rmoskal:
Please pull this in, it's a pretty annoying issue!

mrjoes:
Well, this one is hack and while it solves a problem it is not maintainable. Proper fix would be to handle this logic in the wtforms field and return empty array if nothing was sent from the client.

PR #799, opened 2015-02-28
maybe verbose name a better choce for column name

coolcoca:


mrjoes:
Sorry, won't accept pull request - it is very project specific.

For these kind of changes it is better to add it to your own base model class that you use in your project.

PR #805, opened 2015-03-06
Fix misleading documentation

jmagnusson:


mrjoes:
That's not correct. Only `ImageUploadField` enforces image extensions: https://github.com/jmagnusson/flask-admin/blob/docfix/flask_admin/form/upload.py#L357

jmagnusson:
Ah my bad. I put the "fix" in the wrong place!

PR #806, opened 2015-03-09
Python 3 fix: FilterConverter sub-classing bug

jmagnusson:
A bunch of hairs pulled later I finally found the source for my problems.
FilterConverter wouldn't work properly if instantiated more than once in Python 3 because the return value of a `map()` call in the decorator `flask_admin.model.filters.convert`, which now returns a run-once generator. This would make the second instantiation's `FilterConverter.conv_XYZ._converter_for` empty, leading to a `Exception: Unsupported filter type X` exception.

So now it should be possible to do sub-classes of `FilterConverter` in Py3. At least it does for me.

### DIFF
--- a/flask_admin/model/filters.py
+++ b/flask_admin/model/filters.py
@@ -266,7 +266,7 @@ def convert(*args):
         See :mod:`flask.ext.admin.contrib.sqla.filters` for usage example.
     """
     def _inner(func):
-        func._converter_for = map(str.lower, args)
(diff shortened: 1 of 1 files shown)


mrjoes:
Good one, thanks!

jmagnusson:
No worries, glad to be contributing to such a well-built project. Any chance for a new release?

mrjoes:
I published 1.1.0 to PyPI last week, but without this fix.

Issue #686, opened 2014-11-10
Cause possible data loss when form_edit_rules do not display some columns


Eliacy:
Take the following models as an example:

    class Image(db.Model):
        id = db.Column(db.Integer, primary_key=True) 
        path = db.Column(db.String(120), default='')
        user_id = db.Column(db.Integer, db.ForeignKey('user.id'))
        user = db.relationship('User', backref=db.backref('images', lazy='dynamic'), foreign_keys=[user_id])
        
    class User(db.Model):
        id = db.Column(db.Integer, autoincrement='ignore_fk', primary_key=True)
        username = db.Column(db.String(80), unique=True, default='')

I'd like to record who uploaded detailed images, but one user could relate to hundreds of images, and make the User edit form to be super slow to load. So I decide to hide 'images' column in User edit form, and write down the following code:

    class UserView(ModelView):
        form_edit_rules = ('username')

Then when I submit a modification of a user instance, all his/her 'images' value will be wiped to be '[]', and all these images instances will lose their original 'user_id' value. It is because the form param to 'update_model(self, form, model):' still keep an empty 'images' item.

I think columns not in form_edit_rules should not be modified, that will be a more straightforward mechanism. Thanks.

larkin:
@mrjoes I was thinking that "sensible logic" would be that any field that is not rendered in the page should also be excluded from the form so that it doesn't override the form with default values.  Similar to `form_columns`, a column that is excluded because of `form_edit_rules` should be excluded from the form and not rendered nor expected in the response.

Any reason why this shouldn't be so?

mrjoes:
It will be hard to do it in generic way - one can implement custom rule and even nest them. Probably it should be possible to implement following logic:

1. In rule constructor contribute column to list of rendered columns
2. Compare list of rendered columns with `form_columns` and show a warning if they don't match

larkin:
I may be missing something here, but couldn't this be resolved with a change to `BaseModelView.edit_view` and `BaseModelView.create_view`: after instantating a form, we can generate a list of visible fields based on `form_rules` and remove non-visible fields?

Here's a pseudo-patch for `BaseModelView.edit_view`-- can you explain to me what would be the downsides or problems with the following solution?  If it's a reasonable avenue for this, I'm happy to pursue it further.


```python
    @expose('/edit/', methods=('GET', 'POST'))
    def edit_view(self):
        """
            Edit model view
        """
        return_url = get_redirect_target() or self.get_url('.index_view')

[...]

PR #815, opened 2015-03-14
remove not-visible fields from forms

larkin:
This is a fix for #686:  it prevents accidental data loss that can be caused when form fields are not included in form_rules.

Each form rule needs to define a property for `visible_fields` which returns all fields that the rule exposes.  I've covered all the basic form rules.

@mrjoes how does this look?  I'm open to any / all feedback.

### DIFF
--- a/flask_admin/model/base.py
+++ b/flask_admin/model/base.py
@@ -5,6 +5,7 @@ from flask import (request, redirect, flash, abort, json, Response,
                    get_flashed_messages)
 from jinja2 import contextfunction
 from wtforms.fields import HiddenField
+from wtforms.fields.core import UnboundField
 from wtforms.validators import ValidationError, Required
 
 from flask_admin.babel import gettext
@@ -685,6 +686,10 @@ class BaseModelView(BaseView, ActionsMixin):
         # Form rendering rules
         self._refresh_form_rules_cache()
 
+        # Process form rules
+        self._validate_form_class(self._form_edit_rules, self._edit_form_class)
+        self._validate_form_class(self._form_create_rules, self._create_form_class)
+
     # Primary key
     def get_pk_value(self, model):
         """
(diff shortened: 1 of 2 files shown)


flask_admin/model/base.py
```diff
@@ -1477,6 +1477,10 @@ def create_view(self):
             return redirect(return_url)
 
         form = self.create_form()
+        if self._form_create_rules:
```
mrjoes: I'm thinking if it should be really dynamic. Can do as a part of the form scaffolding. If someone will want dynamic form - they can always do it later.

Also, please move this logic into separate method, so it is callable and overridable.

larkin: How a form comes to be: `self.edit_form` -> `self._edit_form_class` -> `self._refresh_forms_cache` -> `self.get_edit_form` -> `self.get_form` -> `self.scaffold_form`.

Where would you say scaffolding stops and dynamic begins-- in `_refresh_forms_cache`?

The problem I see is every point of scaffolding code beyond `_refresh_forms_cache` is implemented or overridden by code outside the base class; so it would require changing all the contrib modules to implement it-- seems bad for compatibility.

Without doing this dynamically, if anyone has existing form overrides, they're still susceptible to this bug, and we have no room to warn them.

In order to maintain separation between form generation and form rendering rules, I think doing this dynamically is best: it allows us to prevent data loss due to this bug no matter how someone implements form generation.  How you set up your form rendering rules should not cause you to lose data.

mrjoes: It is slightly different. In method that calls `edit_form` and getting form class, it can call something like `self._validate_form_class` (or whatever name could be). `_validate_form_class` will call `self._check_render_rules` twice with default implementation that'll filter invisible fields out of the forms. 

Here's rationale:
1. Form can be supplied by the developer, without using Flask-Admin scaffolding. It might contain custom fields that contribute to the logic of the application without changing model itself (you can do all kind of stuff in the `on_model_change`, as form object is passed there). These fields should not be filtered out automatically.

2. Keeping previous point in mind, it should be easy to change default behavior. I'm not yet convinced that straight form class filtering is right default behavior. But agree that at least it should issue a warning. 

3. If developer wants to filter form using list of visible fields - he should be able to do it from any part of the code. In lots of cases there's no need to filter form in views - it won't change in most of the cases. If it is necessary to filter form dynamically, it is always possible to do it in `get_form` or create/edit overloads.

Hope it all makes sense.

larkin:
@mrjoes 

I added a check to `_refresh_cache` that validates each form class and sets a flag to show that the specific form has been validated already.  Then the create/edit view validates the form only if it hasn't been validated already.

In the worst case, this is dynamic validation and in the best case this is validation during scaffolding.  It still allows flexibility in how for form scaffolding overrides are applied (`edit_form`, `get_edit_form`, `scaffold_form`, etc.) without allowing accidental data loss.

If a developer supplies a form via their own means (e.g. is not using Flask-Admin scaffolding), and is also using form rules, a warning when a field is not included in the form rules is helpful in answering the question, "why can't I see this new field I added?".

What do you think?

PR #816, opened 2015-03-16
Fix “object has no attribute 'message’” in py3

jmagnusson:


jmagnusson:
Would be nice to have!

Issue #802, opened 2015-03-04
Bug in exception handler of _on_model_change


cybertoast:
The current handler for dealing with number of arguments in `_on_model_change()` is incorrect, and obscures cases of the method being overridden.

The fix is to change the [method](https://github.com/mrjoes/flask-admin/blob/master/flask_admin/model/base.py#L1108) to:

```
    def _on_model_change(self, form, model, is_created):
        """
            Compatibility helper.
        """
        try:
            self.on_model_change(form, model, is_created)  
        except TypeError, exc:
            if not re.match(r"on_model_change.*takes exactly.*arguments", str(exc)):
                raise
            msg = ('%s.on_model_change() now accepts third ' +
                   'parameter is_created. Please update your code') % self.model
            warnings.warn(msg)

            self.on_model_change(form, model)
```

I can submit a pull request if necessary.

mrjoes:
It was backwards compatibility hack (sinace 1.0.4, I think) and no longer necessary. I plan to rename `_on_model_change` to `on_model_change` and remove compatibility layer.

Pull requests are always welcome.

moser:
@mrjoes Fast forward 3+ years and the compatibility layer is still there. No blame here, I am definitely guilty of keeping hacks (read "compatibility layers") for longer periods.
But as the pretty broad "catch TypeError" cost me a couple of hours in the last weeks, can you merge the PR (if it is still mergable) or just remove the hack finally in one of the next releases. I was just at the point of submitting a PR of the fashion myself and I am glad that I checked for old issues first.

mrjoes:
Yeah, I actually thought about removing the compatibility layer, but merging this PR is definitely easier to do.

moser:
Thanks, also generally for the great work on flask-admin :+1:

PR #826, opened 2015-03-26
swallow TypeError only if related to compatibility

nitishr:
Handling TypeError in _on_model_change without any additional checking for the cause causes an extremely puzzling error message in case a TypeError is raised by the code in on_model_change. Checking the error's message to see if it's actually caused by incompatible on_model_change signature give more readable error message and trace when the TypeError is caused by the user's code.


nitishr:
Fixes #802

PR #831, opened 2015-03-30
Redirect to edit view after creating new model

longhotsummer:
This is less confusing for the user because they don't return to a list view which may not actually include the newly created model.

Changes `view.create_model()` to return the newly created model, including for all concrete backends.

### DIFF
--- a/flask_admin/contrib/appengine/view.py
+++ b/flask_admin/contrib/appengine/view.py
@@ -56,7 +56,7 @@ class NdbModelView(BaseModelView):
 			model = self.model()
 			form.populate_obj(model)
 			model.put()
-			return True
+			return model
 		except Exception as ex:
 			if not self.handle_view_exception(ex):
 				#flash(gettext('Failed to create record. %(error)s',
@@ -137,7 +137,7 @@ class DbModelView(BaseModelView):
 			model = self.model()
 			form.populate_obj(model)
 			model.put()
-			return True
+			return model
 		except Exception as ex:
 			if not self.handle_view_exception(ex):
 				#flash(gettext('Failed to create record. %(error)s',
--- a/flask_admin/contrib/mongoengine/view.py
+++ b/flask_admin/contrib/mongoengine/view.py
@@ -543,7 +543,7 @@ class ModelView(BaseModelView):
         else:
             self.after_model_change(form, model, True)
 
-        return True
+        return model
 
     def update_model(self, form, model):
         """
--- a/flask_admin/contrib/peewee/view.py
+++ b/flask_admin/contrib/peewee/view.py
@@ -384,7 +384,7 @@ class ModelView(BaseModelView):
         else:
             self.after_model_change(form, model, True)
 
-        return True
+        return model
 
     def update_model(self, form, model):
         try:
--- a/flask_admin/contrib/pymongo/view.py
+++ b/flask_admin/contrib/pymongo/view.py
@@ -287,7 +287,7 @@ class ModelView(BaseModelView):
         else:
             self.after_model_change(form, model, True)
 
-        return True
+        return model
 
     def update_model(self, form, model):
         """
--- a/flask_admin/contrib/sqla/view.py
+++ b/flask_admin/contrib/sqla/view.py
@@ -892,7 +892,7 @@ class ModelView(BaseModelView):
         else:
             self.after_model_change(form, model, True)
 
-        return True
+        return model
 
     def update_model(self, form, model):
         """
--- a/flask_admin/model/base.py
+++ b/flask_admin/model/base.py
@@ -1227,7 +1227,7 @@ class BaseModelView(BaseView, ActionsMixin):
         """
             Create model from the form.
 
-            Returns `True` if operation succeeded.
+            Returns the model instance if operation succeeded.
 
             Must be implemented in the child class.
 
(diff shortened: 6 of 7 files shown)


mrjoes:
This one is questionable. I see where it is coming from: with "Submit" button, action was well defined. With "Save" button it is logical to stay on the same page and use other means to get back to the list view (for example - with help of Cancel button).

I'll keep this one open (you can change your base class and template to get desired for your behavior) and would like to collect more feedback first.

longhotsummer:
Great, feedback is welcome. We could also make this configurable (since to implement this behaviour without this patch you have to re-implement both `create_model` and `create_view` which duplicates a lot of code.)

My motivation is that it's jarring to go back to a listing view that doesn't (potentially) even contain your newly created item. A use case we see a lot is that users want to create an item and then double check it, either because some fields are only filled in after creation or because they want to save a draft and then keep editing. As it stands, they have to save the new item and then find/search for it in the listing view to keep editing it.

petrus-jvrensburg:
I also think this is a good idea. Any objections if I merge?

rochacbruno:
I always use 3 buttons (I include in my customized theme)

[Save, Save and continue editing, Cancel]

This is how Django admin does, I think it is very nice

fzumstein:
Guys, I think you got this and issue #884 pretty wrong. (And I would at least have expected a note about the change of behaviour in the 1.2.0 change log). The behaviour seems totally inconsistent to me.

**Current unintuitive user experience:**
* Add a single new record: `Create > Save > Cancel`. Adding a new record must never ever end with a click on `Cancel`.
* Add a new record: In this case, `Save` means: Just save and do nothing else (actually, it changes to edit view, but I bet that not a single user notices that. Also, what's the big difference between editing a new record vs. an existing one from the user's perspective).
* Add a new record when `can_edit = False`: Now `Save` means again go back to List view. (And users won't check if they have edit rights or not when creating a new record, so this seems inconsistent for them)
* Edit a record: After having learnt how the `save` button works for a new record, I would expect to stay on the edit page, but no, in this case, `Save` means: Leave the form and go back to list view.
* Create and edit mode have 3 buttons each, but the middle one is different (`Save and add` vs. `Save and continue`) which is a bad user experience per se. Every normal user expects these buttons to work consistently. But no, `Save and continue` is for edit mode what `Save` is in create mode.
* Another unexpected behaviour caused by the last point is that you click on `Save` while creating a new record ("just to save things for now"), then after that you want to click on `Save and add` - "huh - where did that button go?"

**How does Django admin do it? i.e. how should it be done correctly**?
* Consistency:  No matter whether in create or edit mode, django has always the same 3 buttons: `Save`, `Save and continue`, `Save and add`. They always perform the same action no matter what. Save closes out the dialog and goes back to list view and the others are self explanatory. You may have been misguided here by the fact that the `User` table shows a non-default behaviour in that `Save` doesn't go back to list view directly but goes to the edit-in-detail page.
* Now, since django admin doesn't have a `Cancel` or `Close` button, you always see the breadcrum menu list which allows you to back out. So for flask admin, either always show the `List` tab (also in edit mode) or introduce a `Cancel/Close` button somewhere which might end up in an overdose of buttons, though.

rochacbruno:
:+1: @fzumstein

mrjoes:
@fzumstein Sounds good, any chance for the pull request?

fzumstein:
Would love to, but I fear I am not very efficient at it :/

rochacbruno:
I can try to fix this, but I it will take some time because I am not able to work on this right now, maybe tomorrow or by this weekend.

longhotsummer:
+1 for harmonising the buttons across all the views, and making the buttons in the same positions do similar things across different views. Borrowing from Django works well for me.

petrus-jvrensburg:
@fzumstein I agree, the `Create > Save > Cancel` flow is not intuitive at all. Also, it would be great to have the same tabs on the `edit` page as on the `add` page. And harmonising the three buttons makes lots of sense.

Thanks for coming back to this issue. Looks like we might get a decent UX improvement out of it.

pawl:
@fzumstein I'm planning to add tabs to the edit_view like this: http://i.imgur.com/LZO6tnu.png

I also want to change the text on the buttons. I'd like to do something similar to django grappelli's buttons: http://i.imgur.com/dPUYjwK.png (Delete, Save, Save and add another, Save and continue editing)

For the create_view, we could have a cancel button instead of the delete button.

Does that sound good?

fzumstein:
@pawl I think this could work... is the `Edit` tab always visible or only when you click on the pencil to edit a record?

pawl:
@fzumstein It's only visible when editing a record. I thought it looked odd without it: http://i.imgur.com/cpbOSlt.png

I saw the issue by @petrus-jvrensburg (https://github.com/flask-admin/flask-admin/pull/831#issuecomment-115265197) after my comment yesterday.

It's a good summary of what needs to happen:
>It looks like the concensus is for three buttons on both the 'add' and 'edit' views: 
Save, Save and continue, Save and add.

>Furthermore, in stead of having an explicit cancel button, we'll display the same tabs from the 'add' view on the 'edit' view.

However, I would like to keep the cancel button since users are used to it, but I think we can keep it from looking crowded by aligning to the other side. I'd also like to change the text on the buttons.

fzumstein:
sounds good to me then!

Am Sonntag, 5. Juli 2015 schrieb Paul Brown :

ztane:
Redirecting to edit view after creating a model isn't idiot-proof however: one of my users got confused and used the *User edit form* to *"create 3 more user accounts"* - well, we thought there was a strange session sharing bug when one user logged in to see the another user's name within the app.

PR #833, opened 2015-03-30
Add CSS classes for columns

longhotsummer:
Adds a `column-{{name}}` CSS class to columns in list view, making it easy to style columns (eg. set widths) using CSS.

### DIFF
--- a/flask_admin/templates/bootstrap3/admin/model/list.html
+++ b/flask_admin/templates/bootstrap3/admin/model/list.html
@@ -61,7 +61,7 @@
                     {% endblock %}
                     {% set column = 0 %}
                     {% for c, name in list_columns %}
-                    <th class="column-header">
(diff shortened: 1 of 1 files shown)


longhotsummer:
This also removes the `column-header` CSS class which I couldn't see any use of, either in CSS or Javascript.

If someone needs to style all table headers they can apply styles to the `th` CSS selector.

mrjoes:
I'd keep column-header, as I know few projects where guys did some customizations around it. But I like idea of adding custom classes to columns. Having column-{{ name }} in addition to column-header is good idea.

longhotsummer:
I can add that back in. My concern is that there will be a name clash if there is a model attribute called `header`, which seems reasonably likely to me. It seems easier to work around the lack of a `column-header` attribute than to work around having a name clash.

PR #837, opened 2015-04-09
menu: unicode conversion

hachreak:
- Ensures that strings are unicode.

Signed-off-by: Leonardo Rossi leonardo.r@cern.ch


mrjoes:
Can't merge - it is not py3 compatible. There's `flask_admin._compat` and it has `text_type`. Please use it instead - https://github.com/mrjoes/flask-admin/blob/master/flask_admin/_compat.py#L38

hachreak:
I made the changes... it's better now? :)

PR #852, opened 2015-04-22
control inline model row control display (bootstrap3)

nitishr:
Currently, although the render_inline_fields macro takes an optional
check parameter, it's not passed from the default inline_field_list
templates, forcing applications to override the template just to control
the parameter.


nitishr:
Closing since this is included now in PR #853.

PR #853, opened 2015-04-22
control inline model row control display

nitishr:
Currently, although the render_inline_fields macro takes an optional
check parameter, it's not passed from the default inline_field_list
templates, forcing applications to override the template just to control
the parameter.


petrus-jvrensburg:
This looks fine, but can you add something to the documentation (or to one of the examples) about what the `check` parameter does for inline fields, and how to use it? I can't find anything on it.

nitishr:
Done. Also included the bootstrap3 file change in this PR. So, closing PR #852.

PR #854, opened 2015-04-22
Add config.MAPBOX_ACCESS_TOKEN

### LINKED ISSUE #842: Geo module doesn't display maps due to mapbox API upgrade
The geo module doesn't work with mapbox v4 API as api requests require mapbox access token. The v3 API currently used by flask-admin seems to be no longer supported (returns 401 Unauthorized status).

nitishr:
Fixes #842

### DIFF
--- a/flask_admin/static/admin/js/form-1.0.0.js
+++ b/flask_admin/static/admin/js/form-1.0.0.js
@@ -73,8 +73,8 @@
        * Process Leaflet (map) widget
        */
       function processLeafletWidget($el, name) {
-        if (!window.MAPBOX_MAP_ID) {
-          console.error("You must set MAPBOX_MAP_ID in your Flask settings to use the map widget");
+        if (!(window.MAPBOX_MAP_ID && window.MAPBOX_ACCESS_TOKEN)) {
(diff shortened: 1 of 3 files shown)


petrus-jvrensburg:
@nitishr Can you squash these commits?

flask_admin/static/admin/js/form-1.0.0.js
```diff
@@ -155,7 +155,7 @@
         }
 
         // set up tiles
-        L.tileLayer('http://{s}.tiles.mapbox.com/v3/'+MAPBOX_MAP_ID+'/{z}/{x}/{y}.png', {
```
petrus-jvrensburg: Has v3 API completely stopped functioning? Or will these changes break existing functioning code running on the v3 API?

abkfenris: The v3 API is still accessible for maps created before mid March, but for anything newer, only v4 is supported. Earlier maps will work just fine with the v4 API, people will have to add MAPBOX_ACCESS_TOKEN if they upgrade flask-admin.

nitishr:
Commits squashed

PR #860, opened 2015-04-27
Give users the option to use mapbox v4

kespindler:
You can't create a new Mapbox account & use the v3 of the API. This gives users the ability to use the v4 API.

### DIFF
--- a/flask_admin/templates/bootstrap2/admin/lib.html
+++ b/flask_admin/templates/bootstrap2/admin/lib.html
@@ -187,6 +187,10 @@
   {% if config.MAPBOX_MAP_ID %}
   <script>
   window.MAPBOX_MAP_ID = "{{ config.MAPBOX_MAP_ID }}";
+  {% if config.get(MAPBOX_API_VERSION) == 4 %}
+  window.MAPBOX_ACCESS_TOKEN = "{{ config.MAPBOX_ACCESS_TOKEN }}";
+  window.MAPBOX_API_VERSION = "{{ config.MAPBOX_API_VERSION }}";
+  {% endif %}
   </script>
   <script src="{{ admin_static.url(filename='vendor/leaflet/leaflet.js') }}"></script>
   <script src="{{ admin_static.url(filename='vendor/leaflet/leaflet.draw.js') }}"></script>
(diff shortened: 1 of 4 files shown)


mrjoes:
This is duplicate of #854?

kespindler:
In functionality, yes. This gives users a flag to switch between v3 and v4, which maintain backwards compatibility for existing users.

flask_admin/templates/bootstrap2/admin/lib.html
```diff
@@ -187,6 +187,10 @@
   {% if config.MAPBOX_MAP_ID %}
   <script>
   window.MAPBOX_MAP_ID = "{{ config.MAPBOX_MAP_ID }}";
+  {% if config.get(MAPBOX_API_VERSION) == 4 %}
+  window.MAPBOX_ACCESS_TOKEN = "{{ config.MAPBOX_ACCESS_TOKEN }}";
+  window.MAPBOX_API_VERSION = "{{ config.MAPBOX_API_VERSION }}";
```
nitishr: Either the double quotes need to be removed, or the comparison in the js file has to be non-strict or with a string

flask_admin/templates/bootstrap3/admin/lib.html
```diff
@@ -169,7 +169,7 @@
   <link href="{{ admin_static.url(filename='vendor/select2/select2.css') }}" rel="stylesheet">
   <link href="{{ admin_static.url(filename='vendor/select2/select2-bootstrap3.css') }}" rel="stylesheet">
   <link href="{{ admin_static.url(filename='vendor/bootstrap-daterangepicker/daterangepicker-bs3.css') }}" rel="stylesheet">
-  {% if config.MAPBOX_MAP_ID %}
+  {% if config.MAPBOX_MAP_ID and config.MAPBOX_ACCESS_TOKEN %}
```
nitishr: This will prevent inclusion of the leaflet assets even for the v3 API, since MAPBOX_ACCESS_TOKEN isn't expected in the config in that case.

flask_admin/templates/bootstrap3/admin/lib.html
```diff
@@ -179,9 +179,13 @@
 {% endmacro %}
 
 {% macro form_js() %}
-  {% if config.MAPBOX_MAP_ID %}
+  {% if config.MAPBOX_MAP_ID and config.MAPBOX_ACCESS_TOKEN %}
```
nitishr: This will prevent inclusion of the leaflet assets even for the v3 API, since MAPBOX_ACCESS_TOKEN isn't expected in the config in that case.

PR #861, opened 2015-04-28
Coerce MultiPolygon to Regular polygon to enable leaflet.draw editing

kespindler:
```
Coerce MultiPolygon to Regular polygon to enable leaflet.draw editing

Leaflet.draw does not currently support multipolygon features.
This coerces a MultiPolygon to be a collection of individual Polygon
features so that they can be edited successfully.
```

### DIFF
--- a/flask_admin/static/admin/js/form-1.0.0.js
+++ b/flask_admin/static/admin/js/form-1.0.0.js
@@ -72,26 +72,19 @@
       // this function fixes the lack of support for multipolygon in Leaflet.draw
       function splitMultiPolygonIntoPolygonFeatures(geojson) {
         var splitFeatures = [];
-        if (!geojson.features) {
-          return geojson;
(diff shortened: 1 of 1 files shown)


mrjoes:
@pawl can you review this pull request?

petrus-jvrensburg:
@kespindler thanks for this fix. But I'm getting `leaflet TypeError: Cannot read property 'lat' of undefined` in my browser console when trying this out locally, editing a regular `POINT` record.

Could you merge master into your branch and check what's broken? 

Also, could you use the `/examples/geo_alchemy/app.py` and check expand the fix to include `MULTIPOINT` and `MULTILINE` geometry types in addition to `MULTIPOLYGON`?

petrus-jvrensburg:
closing this, this shouldn't be an issue with the latest leaflet update

PR #862, opened 2015-04-28
Revert 860 ks mapbox v4

nitishr:
The earlier mapbox v4 fix in PR #860 didn't work due to the version number being assigned as a string but strictly compared as a number in javascript. This PR fixes that issue as well as simplifies the code, not requiring an additional MAPBOX_API_VERSION config variable. That variable was misleading anyway, since it was acting more as a boolean flag to check if the v4 API is to be used.


mrjoes:
Thank you.

PR #863, opened 2015-04-28
fix mapbox-v4

kespindler:
Resolves issues from #860 and #862 .

### DIFF
--- a/flask_admin/static/admin/js/form-1.0.0.js
+++ b/flask_admin/static/admin/js/form-1.0.0.js
@@ -78,10 +78,6 @@
           return false;
         }
 
-        if (!window.MAPBOX_ACCESS_TOKEN) {
-          console.warn("You must set MAPBOX_ACCESS_TOKEN in your Flask settings to use the map widget with mapbox v4 API");
-        }
-
         var geometryType = $el.data("geometry-type")
         if (geometryType) {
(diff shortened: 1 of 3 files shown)


flask_admin/static/admin/js/form-1.0.0.js
```diff
@@ -78,10 +78,6 @@
           return false;
         }
 
-        if (!window.MAPBOX_ACCESS_TOKEN) {
```
mrjoes: If there's a way to check version on client-side and spit out warning if access token is missing?

kespindler:
It's a good point - that's the philosophical problem with removing the MAPBOX_API_VERSION field. For full clarity, I think re-introducing that is a good thing, and I can re-add to PR.

mrjoes:
No, don't add server-side MAPBOX_API_VERSION configuration setting. I bet there's way to check on client-side what's mapbox version and only then check if token is provided.

nitishr:
I apologize for the copy and paste errors in my earlier patch. Will be a lot more careful next time.

I've verified this fix locally and it works fine. Thanks @kespindler. Can this be merged, since without this, the map integration remains broken for mapbox API v4 users.

kespindler:
@nitishr Yup, it works fine on my end too. Merge at will!

PR #864, opened 2015-04-30
Allow markup in field help-text

cybertoast:
This is to allow help_text in the model to have HTML markup, and render in the forms.


pawl:
@mrjoes any objections to merging this? It seems unlikely that someone would be populating this with user input, and it seems very likely that someone would want to do additional styling.

pawl:
This should also be on the bootstrap2 template.

pawl:
Closing this until bootstrap 2 lib.html change is added.

PR #865, opened 2015-05-01
Default to case-insensitive filtering and searching (MongoEngine)

jnga:
The ability to do case-insensitive filtering and searching with MongoEngine was added with pull request #177. Preceding the search term with an asterisk made the query case insensitive.

This change is to make case insensitive the default. The aim is to match the usual behavior of search boxes on web sites and operating systems.

In order to do this while keeping case-sensitive capability, the asterisk now means the opposite: case sensitive.

### DIFF
--- a/flask_admin/contrib/mongoengine/tools.py
+++ b/flask_admin/contrib/mongoengine/tools.py
@@ -1,17 +1,17 @@
 def parse_like_term(term):
     """
         Parse search term into (operation, term) tuple. Recognizes operators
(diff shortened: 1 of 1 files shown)


gordol:
seems bad to change default behavior after all these years, but it should have been case insensitive by default. however, using asterisk to mean the query is case sensitive is really weird to me... surely we should have used something else as a flag, so as to remain backwards compatible instead if literally inverting the expected behavior.

PR #877, opened 2015-05-19
FileUploadInput fix

vovanz:
If you edit a model where file field is required you will get a validation error if file field is set but wasn't changed.
You should pass filename in form data if it is already set.


vovanz:
Closed because of misprint, recreated: https://github.com/flask-admin/flask-admin/pull/878

PR #881, opened 2015-05-21
add trove classifiers to setup.py

jamesturk:
will make it so https://caniusepython3.com stops showing this as a blocker

based on the versions currently tested in Travis


Mause:
According to the documentation, the one that is actually required is;
`Programming Language :: Python :: 3`

As such, it's still shown as a blocker, for example, for [airflow](https://caniusepython3.com/project/airflow)

Mause:
Erm, disregard that, it seems the caching is broken on https://caniusepython3.com

PR #888, opened 2015-06-01
Adds Leaflet Search to Map

yefim:
Adds a configurable (just set the `MAPBOX_SEARCH` flag) geocoding search to the default mapbox view. This is very useful for placing pins at precise locations (a home, an office, etc.).

@stefanocudini, could you chime in on what you think of adding this to Flask Admin?

Here's what the final product looks like below:

screenshot 2015-06-01 12 09 36


yefim:
Thanks @mrjoes :)

stefanocudini:
@mrjoes this is very interesting!!  congrants!

I suggest you ti use a git submodules for any leaflet plugins in your project, this allow to maintain updated all version of plugins in simple way, like as this my project that use more plugins:
https://github.com/stefanocudini/osm-contributors-by-area


If you want promote your project you can add it on this list of use cases:
https://github.com/stefanocudini/leaflet-search/wiki/Websites-that-use-Leaflet-Control-Search

dkerkow:
@yefim does not work with a recent version of Flask-Admin. MAPBOX_MAP_ID, MAPBOX_ACCESS_TOKEN and MAPBOX_SEARCH = True are all set, and the map is showing, but not the search element. Am I missing something or is it a bug? Flask-Admin version is 1.4.0

PR #896, opened 2015-06-09
Fix custom field classes with Bootstrap3

### LINKED ISSUE #894: WYSIWYG example fails with 'template_mode=bootstrap3'
Running WYSIWYG example with `template_mode=bootstrap3` results in non-functioning example. This is because the `class` property is specified twice for the textarea widget:
```
<textarea class="form-control" class="ckeditor" id="text" name="text"></textarea>
```

plaes:
Currently the custom class is created inside the call to `field()` so we
can't actually check in the template whether the field or widget has set
any custom classes.

With Bootstrap3, the `render_field` macro overwrites the `class` anyway,
so we have to define the widget's CSS class in the Python class level
and look it up before we actually create the object.

Fixes #894

### DIFF
--- a/examples/wysiwyg/app.py
+++ b/examples/wysiwyg/app.py
@@ -20,9 +20,7 @@ db = SQLAlchemy(app)
 
 # Define wtforms widget and field
 class CKTextAreaWidget(widgets.TextArea):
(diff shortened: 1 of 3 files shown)


pawl:
This was fixed in #927

Issue #920, opened 2015-06-30
Cannot configure select2js


sebleblanc:
I would like to set the attribute [`closeOnSelect`](https://select2.github.io/select2/#documentation) in select2 on a specific field, but there is no easy way to achieve this.

PR #921, opened 2015-06-30
Add 'data-no-autoclose' attribute to select2 widget

sebleblanc:
By setting the 'data-no-autoclose' attribute on the Select2Widget
object, the Select2 JavaScript widget will not automatically close
itself after an object has been selected.

You can use it this way:
    class MyModelView(ModelView):
        form_widget_args = dict(
            field = {'data-no-autoclose': u'1'} # any value will also do
            )

IMO the `u'1'` part is counter-intuitive (setting it to any value has the same behaviour) but I am respecting the pattern used by other attributes of the widget (e.g. 'data-allow-blank').

This commit references issue #920. I based it off v1.1.0 simply because my project uses this version and I could not test it with the latest. It should be easy to merge though.


pawl:
The new select2 4.0 allows setting options through data attributes. It might just be better to upgrade select2 instead of adding more hacks.

sebleblanc:
In the meantime this fix works for me.

pawl:
Unfortunately, the bootstrap theme for Select2 4.0 is still a work in progress: https://github.com/select2/select2-bootstrap-theme

I tried testing 4.0 without the bootstrap theme for a few minutes and noticed a few problems:
* Select2 input fields not the same size as normal input fields: https://i.imgur.com/BAKdbTF.png
* More size issues: https://i.imgur.com/LG2h0JW.png
* X button here doesn't work: https://i.imgur.com/l5V0WOH.png

petrus-jvrensburg:
Closing this. But please re-open if it's still an issue.

PR #924, opened 2015-07-03
add query string versioning to assets

pawl:
The problems:
* When we upgrade assets, we need the url to change so it busts cache and gets the newest version of the asset.
* If a developer using flask-admin completely overrides a template block containing scripts and the script file name changes in the future - upgrading to a version of flask-admin with different asset file names will break their site. I think it's common for developers to do this with base.html. However, this example shows the correct way to override blocks with assets: https://github.com/flask-admin/flask-admin/blob/master/examples/layout_bootstrap3/templates/layout.html#L5

The solution in this pull request: 
* Use a query string with the version as part of the asset URL, like "jquery.min.js?v=2.1.4". This will bust cache + prevent completely overridden templates from getting 404 errors.

There's some controversy over using query strings for assets: http://www.stevesouders.com/blog/2008/08/23/revving-filenames-dont-use-querystring/

> I wasn’t aware of any performance difference, but in a meeting this week a co-worker, Jacob Hoffman-Andrews, mentioned that Squid, a popular proxy, doesn’t cache resources with a querystring.

But this article describes why that isn't the case anymore: http://bizcoder.com/caching-resources-with-query-strings

> That was in August 2008.  Earlier that year, in May 2008, Squid released version 2.7 which changed their default behaviour to no longer refuse to cache URLs that contained a query string.


pawl:
This will be a breaking change for people who have overridden template blocks with scripts and still have the old file names in their code. But, I think this solution is better in the long term than changing file names every time we update an asset (to solve only the caching issue), like in this pull request: https://github.com/flask-admin/flask-admin/pull/918 (jquery-2.1.1.min.js -> jquery-2.1.4.min.js)

PR #935, opened 2015-07-08
Validate form instance of edit form should be against _form_edit_rules

xqliu:
Validate form instance of edit form should be against _form_edit_rules rather than _form_create_rules.

If all the validation is against _form_create_rules then the _form_edit_rules is actually not take effective during validation.


petrus-jvrensburg:
thanks @xqliu

PR #938, opened 2015-07-10
model: filter: fixed float overflow

xiaost:
see: 
https://en.wikipedia.org/wiki/Double-precision_floating-point_format

test code:
```python
>>> int(float("6169453081680413441"))
6169453081680413696
```

### DIFF
--- a/flask_admin/model/filters.py
+++ b/flask_admin/model/filters.py
@@ -109,7 +109,7 @@ class BaseIntFilter(BaseFilter):
         Base Int filter. Adds validation and changes value to python int.
     """
     def clean(self, value):
-        return int(float(value))
+        return int(float(value) if '.' in value else value)
 
 
 class BaseFloatFilter(BaseFilter):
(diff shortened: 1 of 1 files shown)


pawl:
Nice find!

It might be better to use this solution instead?: http://stackoverflow.com/a/379966

They mention a few edge cases that can cause problems when using this solution: http://stackoverflow.com/a/379968

pawl:
It would also be good to add a test for this, but we would need to figure out how to add a working integer filter to the MockModelView here: https://github.com/flask-admin/flask-admin/blob/master/flask_admin/tests/test_model.py#L39

pawl:
I'm also not completely sure why it needs to be anything except just ```int(value)```.

The idea might have been to allow people to enter things like "10.0" into the filter and still get results. Here's the relevant pull request: https://github.com/flask-admin/flask-admin/pull/718

pawl:
I think the best solution is to just use ```int(value)```. Especially since the code is for filtering integer fields anyway.

In the solution you suggested, the issue persists if the user uses decimals:
```
>>> value = "6169453081680413441.50"
>>> int(float(value) if '.' in value else value)
6169453081680413696
```

PR #942, opened 2015-07-12
add read-only details view and use separate templates for modals

pawl:
I've been working on a fix for https://github.com/flask-admin/flask-admin/issues/334

I'm submitting this pull request to get feedback. Also, it looks like I need to add ```scaffold_list_columns``` to pymongo and would appreciate any ideas for implementing it.

Here are some screenshots (FYI, the "Filter" will search the table and hide the rows which don't contain the filter text):

Bootstrap 3 (non-modal)
Bootstrap 3 (non-modal)

Bootstrap 3 list view
Bootstrap 3 list view

Bootstrap 3 (modal)
Bootstrap 3 (modal)

Bootstrap 2 (non-modal)
Bootstrap 2 (non-modal)

Bootstrap 2 list view
Bootstrap 2 list view

Bootstrap 2 (modal)
Bootstrap 2 (modal)

### DIFF
--- a/flask_admin/model/base.py
+++ b/flask_admin/model/base.py
@@ -157,7 +157,7 @@ class BaseModelView(BaseView, ActionsMixin):
         If set to `None`, will get them from the model.
     """
 
-    details_exclude_list = None
+    column_details_exclude_list = None
     """
         Collection of fields excluded from the details view.
     """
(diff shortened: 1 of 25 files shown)


pawl:
This is completed and I added tests for it.

If anyone would like to test it out and give feedback, here's the command you need to run to install it from my branch:
```pip install -e git+https://github.com/pawl/flask-admin.git@add_view_modal#egg=flask-admin```

pawl:
Testing it out will require setting ```can_view_details = True``` in a model view. To test the modal, you can use ```details_modal = True```.

pawl:
Just made another change to this.

I've been worried that I've been making the CRUD templates more cluttered with the modal logic, so I separated the modals into their own templates under /templates/model/modals/ and added attributes to the base modelview for ```edit_modal_template```, ```create_modal_template```, ```details_modal_template```.

Also fixed a quirk with the create_view's "save and continue" button. It was "btn-large" when it needed to be "btn-default" to match the edit_view.

I think it's ready to be merged, but would appreciate if anyone could look it over.

nirenyang:
Is there a way to implements relationship(* to many)? I'm looking around. can't figure out it. Thanks!!

PR #947, opened 2015-07-15
Add custom date example for column_type_formatter

plaes:
Took a bit headscratching to actually implement, so why not document it ;)


pawl:
Works great, just tried it out on the sqla example: 
example of column_type_formatter

PR #961, opened 2015-07-26
Make fields without inline edit html safe by default

xqliu:
Make fields in list page html safe by default, for the ability to render field in different style, only assume it's safe if the field is configured without inline edit.


mrjoes:
Can't accept this one - secure defaults should stay.

If you want to expose unsafe HTML, you can always wrap it with `jinja2.Markup` class in your python app.

Issue #514, opened 2014-04-28
localize filters


moulinux:
I read in the french admin.po that filters have been translated, for exemple contains is used in flask_admin/contrib/sqla/filters.py:52 :+1: 

return gettext('contains')

running flask-admin in french, the localization is completed, except for filters : contains, equals, .. are not localized

dont't understand why...

tahajahangir:
@mrjoes Filters are *cached* at *import time*, so their language will stick on default language. This prevents localizitation of filters.

PR #963, opened 2015-07-27
use lazy_gettext with pymongo filters

pawl:
Finishes adding lazy_gettext to all the filters to fix the issue @tahajahangir mentioned in #514: "Filters are cached at import time, so their language will stick on default language. This prevents localizitation of filters."


tahajahangir:
Thanks for fix

PR #964, opened 2015-07-27
SQLAlchemy: Don’t require data for text fields that have a empty string as default value

jmagnusson:
Current behavior by Flask-Admin is to treat all fields with `nullable=False` as required even if they have `default=''` set. This PR changes this behavior so that one does not have to enter a value if both of these conditions are met.

It might not be totally clear from the formatting but the only difference is the added `and not column.default` to the if-statement.

### DIFF
--- a/flask_admin/contrib/sqla/form.py
+++ b/flask_admin/contrib/sqla/form.py
@@ -205,7 +205,11 @@ class AdminModelConverter(ModelConverterBase):
 
                 optional_types = getattr(self.view, 'form_optional_types', (Boolean,))
 
-                if not column.nullable and not isinstance(column.type, optional_types):
+                if (
+                    not column.nullable
+                    and not isinstance(column.type, optional_types)
+                    and not column.default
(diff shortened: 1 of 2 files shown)


boolbag:
@pawl Thank you, you are right, it is related. Sorry, I hadn't seen it.

One difference between the two requests is that I don't see `default=''` as being a condition for a `NOT NULL` field to accept empty string values. Let's say the `NOT NULL` field `favorite_gift` has a default value of `flowers`. You enter a new record for user Anna, not entering a value. Fine, she gets

pawl:
I moved the commits in this pull request to #1013 (which also checks server_default). Closing this pull request.

@boolbag Looks like this case is already handled. It still allows saving empty strings. But, if you create a new record, it will have "flowers" in the field initially. It's a bit hard to explain, so you may want to test it for yourself with ```pip install git+https://github.com/pawl/flask-admin.git@test_fix_default_not_null```

PR #972, opened 2015-08-03
Hide create button in edit_view unless user has permission

pawl:
Similar to #971, thanks for finding the issue @emve-dev 

Also when .edit_view is shown, the "create" button in the navtab is displayed even if model view has can_create = False.


emve-dev:
Happy to contribute to this awesome extension :+1:

Issue #979, opened 2015-08-12
Details view tab not visible in edit tab


litnimax:

pawl:
I'm thinking it may also make sense to change the tab's name from "View Record" to "View Details", since you're technically viewing a record when you're editing it. Any opinion on that?

litnimax:
It's ok for me :-)

litnimax:
Need to add translation also.

PR #980, opened 2015-08-12
Fix bug with details view tab.

litnimax:
Fix for #979


pawl:
You would also need to make the same change to the bootstrap 2 version: https://github.com/flask-admin/flask-admin/blob/master/flask_admin/templates/bootstrap2/admin/model/edit.html

pawl:
@litnimax Can you add another commit with the changes to this file too?: https://github.com/flask-admin/flask-admin/blob/master/flask_admin/templates/bootstrap2/admin/model/edit.html

litnimax:
@pawl Done.

PR #981, opened 2015-08-12
Allow overriding page_size in get_list

pawl:
This resolves merge conflicts in @tandreas's #644 and adds the change to mongoengine, pymongo, and peewee too. 

Here's more info about the problem this is trying to solve:

Currently, if you want to create a .csv from a table using ```get_list```, it requires this hack: 

* Set ```self.page_size``` to a number equal to or greater than the number of records in the table. (This affects the entire ModelView! If anyone were to visit the page during this time, it would load all records.)
* Use ```get_list``` to get all records from the table.
* Set ```self.page_size``` back to its original value.

### DIFF
--- a/flask_admin/contrib/mongoengine/view.py
+++ b/flask_admin/contrib/mongoengine/view.py
@@ -454,7 +454,7 @@ class ModelView(BaseModelView):
         return query.filter(criteria)
 
     def get_list(self, page, sort_column, sort_desc, search, filters,
-                 execute=True):
+                 execute=True, page_size=None):
         """
             Get list of objects from MongoEngine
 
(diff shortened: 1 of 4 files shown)


petrus-jvrensburg:
How about handling this through the existing `page` parameter on `get_list`, rather than an additional `page_size` parameter?

I would prefer sending `page=None` to `get_list()` to get the whole, unpaginated list returned. Then there's no need for an extra parameter, and the logic can be a little cleaner.

During normal usage, `page` is always an integer: https://github.com/flask-admin/flask-admin/blob/master/flask_admin/model/base.py#L1428. So, for this issue, we can check explicitly `if page is None` to override paging.

pawl:
That would only allow for turning pagination on or off. But, that wouldn't allow us to do things like allow users to adjust their own pagination: Example

pawl:
@petrus-jvrensburg forgot to mention you in the comment above.

> That would only allow for turning pagination on or off. But, that wouldn't allow us to do things like allow users to adjust their own pagination

pawl:
@mrjoes any objections to merging this one too?

mrjoes:
Nope, thanks.

PR #983, opened 2015-08-12
Call _on_model_change() before session.add()

guillaumep:
_on_model_change() should be called before the object is added to the session, to allow things like populate a primary key before the INSERT statement is sent to the database.

### DIFF
--- a/flask_admin/contrib/sqla/view.py
+++ b/flask_admin/contrib/sqla/view.py
@@ -982,8 +982,8 @@ class ModelView(BaseModelView):
         try:
             model = self.model()
             form.populate_obj(model)
-            self.session.add(model)
             self._on_model_change(form, model, True)
+            self.session.add(model)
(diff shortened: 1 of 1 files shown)


pawl:
The insert statement is only sent to the database on ```self.session.commit()```.

You can still make changes to the ```model``` object after ```session.add(model)``` runs. I think you can already change the primary key (from within ```_on_model_change()```) before the INSERT happens.

Here's an example of someone changing ```model``` after ```session.add(model)``` occurs: https://github.com/alexander-emelyanov/forex-advisor/blob/3f24f8b5a7a69be93b779e629a2da271ec21d2c4/admin_models.py#L9

Have you already tried that?

guillaumep:
Hum, you're right. on_model_change is indeed called before the insert.

This closing this PR was the good thing to do.

PR #984, opened 2015-08-13
Bugfix - Non pk ObjectFields were ignored

AlmogCohen:
Now, only the pk field is ignored if `column_display_pk` is not True. The rest of the `ObjectField`s are shown.


petrus-jvrensburg:
Closing this. But if it's still an issue, please re-open, and link to a github issue that describes the problem.

PR #986, opened 2015-08-13
Column sort follows ForeignKeys

ptbrowne:
You have two models:

```python
from peewee import ForeignKey, PrimaryKey, CharField, Model

class Purchase(Model):
     user = ForeignKey(User)

class User(Model):
    id = PrimaryKey()
    name = CharField()
```

Before you could not write

```python
class PurchaseView(ModelView):
    column_sortable_list = (('user', 'user.name'), )
```

because it would fail saying that 'user.name' is not a field of Purchase.

I tried with:

```python
class PurchaseView(ModelView):
    column_sortable_list = (('user', User.name), )
```

but I have an exception

```
 OperationalError: (1066, "Not unique table/alias: 't2'")
```

What I do is splitting on the dot and following foreign keys.

### DIFF
--- a/flask_admin/contrib/peewee/view.py
+++ b/flask_admin/contrib/peewee/view.py
@@ -288,7 +288,9 @@ class ModelView(BaseModelView):
 
     def _order_by(self, query, joins, sort_field, sort_desc):
         if isinstance(sort_field, string_types):
-            field = getattr(self.model, sort_field)
+            field = self.model
+            for f in sort_field.split('.'):
+                field = getattr(field.rel_model if isinstance(field, ForeignKeyField) else field, f)
             query = query.order_by(field.desc() if sort_desc else field.asc())
         elif isinstance(sort_field, Field):
             if sort_field.model_class != self.model:
--- a/flask_admin/tests/peeweemodel/test_basic.py
+++ b/flask_admin/tests/peeweemodel/test_basic.py
@@ -65,13 +65,14 @@ def create_models(db):
 
     class Model2(BaseModel):
         def __init__(self, char_field=None, int_field=None, float_field=None,
-                     bool_field=0):
+                     bool_field=0, model1=None):
             super(Model2, self).__init__()
 
             self.char_field = char_field
             self.int_field = int_field
             self.float_field = float_field
             self.bool_field = bool_field
+            self.model1 = model1
 
         char_field = peewee.CharField(max_length=20)
         int_field = peewee.IntegerField(null=True)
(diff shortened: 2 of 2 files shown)


pawl:
Would you be up for adding a test for this?

The SQLAlchemy tests have a good example: https://github.com/flask-admin/flask-admin/blob/master/flask_admin/tests/sqla/test_basic.py#L1537

ptbrowne:
OK I'll do this. Thanks for the pointer :)

Le ven. 14 août 2015 7:05 AM, Paul Brown <notifications@github.com> a
écrit :

boolbag:
Oddly, I'm able to set a list's **default sort** on a foreign key, but I'm unable to make the foreign column click-sortable. 

```
class SubscriptionAdmin(sqla.ModelView):
    # works. uid is a foreign key.
    column_default_sort = 'uid'      
    
    # neither of these work: 
    # column_sortable_list = ('uid', ) 
    # column_sortable_list = (psmodels.Person.uid,)
```

ptbrowne:
What I have to have the column click-sortable (here it is the "user" column)

```python
    column_default_sort = 'pur_cart.cart_user.user_name'
    column_sortable_list = [('user', 'pur_cart.cart_user.user_name'),
'pur_total']
```

where my models are

```python
class Purchase(Model):
    pur_cart = ForeignKey(Cart)
    pur_total = IntegerField()

class Cart(Model):
    cart_user = ForeignKey(User)

class User(Model):
     user_name = CharField()
```

ptbrowne:
What I have to have the column click-sortable (here it is the "user" column)

```python
    column_default_sort = 'pur_cart.cart_user.user_name'
    column_sortable_list = [('user', 'pur_cart.cart_user.user_name'),
'pur_total']
```

where my models are 

```python
class Purchase(Model):
    pur_cart = ForeignKey(Cart)
    pur_total = IntegerField()

class Cart(Model):
    cart_user = ForeignKey(User)

class User(Model):
     user_name = CharField()
```

ptbrowne:
What I have to have the column click-sortable (here it is the "user" column)

```python
    column_default_sort = 'pur_cart.cart_user.user_name'
    column_sortable_list = [('user', 'pur_cart.cart_user.user_name'),
'pur_total']
```

where my models are 

```python
class Purchase(Model):
    pur_cart = ForeignKey(Cart)
    pur_total = IntegerField()

class Cart(Model):
    cart_user = ForeignKey(User)

class User(Model):
     user_name = CharField()
```

boolbag:
@ptbrowne Thank you so much for sharing the syntax that works for you! That tuple cleared it up for me, it's working great now.
`column_sortable_list = (('person', psmodels.Person.uid), )`
Btw for me it only works without the quotes on the second element, probably a different setup somewhere.

pawl:
I tested this on the peewee example by changing:
```column_sortable_list = ('title', ('user', User.email), 'date')```
to
```column_sortable_list = ('title', ('user', 'User.email'), 'date')```

Here's the error I got: https://gist.github.com/anonymous/d058f256d9c6ad997e89

It would be better to abstract this login into something that could be used for searching, sorting, filters, list columns, etc. Like SQLA does: https://github.com/flask-admin/flask-admin/blob/master/flask_admin/contrib/sqla/view.py#L335

ptbrowne:
Why would you want to use a string there ?

mrjoes:
SQLAlchemy backend has this logic abstracted into a method: https://github.com/flask-admin/flask-admin/blob/master/flask_admin/contrib/sqla/view.py#L324

pawl:
@ptbrowne Now I'm confused about what this fixing.

* Sorting already works on foreign keys with this syntax: ```column_sortable_list = ('title', ('user', User.username), 'date')```
* Your message says "what is the advantage of using a string there?", but your test uses the foreign key field name + the field in the other model as a string like: "model1.test1"
* When I create a gist similar to the test case (but using column_sortable_list), I get this error: ```OperationalError: no such column: t2.username``` (https://gist.github.com/pawl/ce519c38308fc8c16fc2#file-test_peewee_sortable-py-L43)

pawl:
I think this one was fixed by https://github.com/flask-admin/flask-admin/issues/1192

PR #991, opened 2015-08-14
Change buttons on edit/create views, change redirect behavior

### LINKED ISSUE #884: Buttons on create view
Should we default to having three buttons [Save, Save and continue editing, Cancel] on the default create model view?
Thanks @rochacbruno
https://github.com/flask-admin/flask-admin/pull/831#issuecomment-105062373

pawl:
Fixes #884 (see discussion in #831 for more background info)

Create View: new create view

Edit View:  new create view

Here's the list of changes:
* Save button now redirects back to list view. But, there is now a "Save and Continue Editing" button on the ```.create_view```.
* Both ```.edit_view``` and ```.create_view``` have the same 3 buttons now.
* Changed "Save and Add" to "Save and Add Another"
* Changed "Save and Continue" to "Save and Continue Editing"
* Aligned form buttons to the right.
* Removed the cancel button to prevent crowding (users can use the "List" button in the upper left)
* Changed bootstrap2 form buttons to be the same size as the bootstrap3 form buttons and added an ```<hr>``` between the form buttons and form.
* Changed "View Record" tab on ```.edit_view``` to say "Details". (only shown when ```can_view_details = true```)

Removing the cancel button is probably the most controversial change here. But, it looks kinda crowded with that button there: https://i.imgur.com/p1ePE7i.png

### DIFF
--- a/babel/admin.pot
+++ b/babel/admin.pot
@@ -9,14 +9,14 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Flask-Admin VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2015-07-28 00:55-0500\n"
+"POT-Creation-Date: 2015-08-14 19:32-0500\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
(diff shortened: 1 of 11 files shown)


petrus-jvrensburg:
This is great. But I'm having second thoughts on taking out the 'cancel' button. There's something reassuringly simple in a big button that says cancel.

What if we keep all the regular buttons left-aligned, and float the 'cancel' button to the right? Would that still bee too cluttered?

Issue #988, opened 2015-08-13
column_filters is not compatible with babel


litnimax:
Using current master.
When adding:

```
column_labels = dict(
        created=lazy_gettext('Created'),
        message=lazy_gettext('Message'),
    )
column_filters = ['created', 'message']
```

get the following error:

```
127.0.0.1 - - [13/Aug/2015 19:13:17] "GET /logbook/?sort=1 HTTP/1.1" 500 -
Traceback (most recent call last):
  File "/home/max/tmp-work/astbarrier/env/lib/python2.7/site-packages/flask/app.py", line 1836, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/max/tmp-work/astbarrier/env/lib/python2.7/site-packages/flask/app.py", line 1820, in wsgi_app
    response = self.make_response(self.handle_exception(e))
  File "/home/max/tmp-work/astbarrier/env/lib/python2.7/site-packages/flask/app.py", line 1403, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/max/tmp-work/astbarrier/env/lib/python2.7/site-packages/flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/max/tmp-work/astbarrier/env/lib/python2.7/site-packages/flask/app.py", line 1477, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/max/tmp-work/astbarrier/env/lib/python2.7/site-packages/flask/app.py", line 1381, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/max/tmp-work/astbarrier/env/lib/python2.7/site-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/max/tmp-work/astbarrier/env/lib/python2.7/site-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/max/tmp-work/astbarrier/env/lib/python2.7/site-packages/flask_admin/base.py", line 68, in inner
    return self._run_view(f, *args, **kwargs)
  File "/home/max/tmp-work/astbarrier/env/lib/python2.7/site-packages/flask_admin/base.py", line 359, in _run_view
    return fn(self, *args, **kwargs)
  File "/home/max/tmp-work/astbarrier/env/lib/python2.7/site-packages/flask_admin/model/base.py", line 1636, in index_view
[...]

PR #994, opened 2015-08-16
Fix & test for #988: column_filters in not compatible with babel.

torotil:
Filters currently have two issues with translatable column labels:

 1. _LazyString is not hashable and can't be used as a dict key.
 2. It can't be JSON serialized.

This patch creates an explicit FieldGroup class to store the label of the FieldGroup and it's filters. Now we can use the untranslated FieldGroup label as dict-key while not losing the translatability of the label.

### DIFF
--- a/flask_admin/model/base.py
+++ b/flask_admin/model/base.py
@@ -22,7 +22,8 @@ from flask_admin.helpers import (get_form_data, validate_form_on_submit,
                                  get_redirect_target, flash_errors)
 from flask_admin.tools import rec_getattr
 from flask_admin._backwards import ObsoleteAttr
-from flask_admin._compat import iteritems, OrderedDict, as_unicode, csv_encode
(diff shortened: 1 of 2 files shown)


flask_admin/model/base.py
```diff
@@ -61,6 +61,26 @@ def clone(self, **kwargs):
         return ViewArgs(**kwargs)
 
 
+class FilterGroup:
```
mrjoes: New style class please.

jmagnusson:
Thanks for this fix @torotil. Did it for me at least.

@mrjoes Could we have this merged?

mrjoes:
I'd love to, but it breaks the build.

torotil:
That's due to the changes in flask-admin:master that happened in the meanwhile. Looking into it.

torotil:
Now it builds.

jmagnusson:
Nice @mrjoes. Could we have a new PyPI release with this?

PR #992, opened 2015-08-15
updates to 'create' & 'edit' view buttons 

petrus-jvrensburg:
This builds on @pawl's PR #991.

I put back the cancel button, and right-aligned it. While keeping the rest of the buttons left-aligned.

PR #995, opened 2015-08-17
Redirecting to details page after create/edit (v3)

pawl:
This is a version of #992 with the form buttons combined and without right alignment (which builds on PR #991).

Bootstrap 3 create view: 
Bootstrap 3 edit view: 
Bootstrap 3 modal: 

Bootstrap 2 create view: 
Bootstrap 2 edit view: 
Bootstrap 2 modal:


petrus-jvrensburg:
:+1:

PR #996, opened 2015-08-17
Make it possible to override return url after form has been saved

plaes:
Based on suggestion from Paul Brown (pawl):
https://github.com/flask-admin/flask-admin/issues/985


mrjoes:
Good one, thanks.

lud4ik:
why was it reverted? last version do not have possibilty to override.

Issue #1000, opened 2015-08-20
security risk: need html.escape


boolbag:
When database fields contain html, at the moment the raw html is output in the list [edit: in the case of editable fields]. For instance, `<h1>hello</h1>` will be rendered in the browser as a heading (unlike how it is rendered in this sentence). 

![escape](https://cloud.githubusercontent.com/assets/7551671/9376865/be7c7a10-4766-11e5-943c-8fa42ffbe518.jpg)

In my view, this is not a desired behavior and a security risk: think of raw JS stored in the database and rendered, raw, in the browser.

It seems to me that the output of any database field should go through `html.escape` before hitting the browser. 

As a temporary fix, you can bypass this problem with `column_formatters`, but there is a [related issue](https://github.com/flask-admin/flask-admin/issues/1001) that I will post separately.

**Edit: about the free cruise grand prize...** Please let me know how to collect my prize for posting issues #1000 and #1001!

Congratulations for making it this far, and huge thanks for the mammoth effort to everyone involved.

pawl:
I'm having trouble reproducing this in chrome: ![](https://i.imgur.com/NkFCQoV.png)

Using this gist: https://gist.github.com/pawl/a70fd8ff2094bf43c6fb

boolbag:
@pawl 
That was fast!
But add this to your gist (commenting out your add_view line):

```
class CarAdmin(sqla.ModelView):
    column_editable_list = ('desc', )

# later...
# admin.add_view(sqla.ModelView(Car, db.session))
admin.add_view(CarAdmin(Car, db.session))
```

Scary results... Thanks for all the alerts! :)

PR #1002, opened 2015-08-20
fix xss vulnerability - escape html in column_editable_list values

pawl:
Fixes the XSS vulnerability that @boolbag found in #1000 by escaping HTML before it's passed to WTForm's HTMLString. Nice catch @boolbag!

Here's a picture from after the fix is added:

### DIFF
--- a/flask_admin/model/widgets.py
+++ b/flask_admin/model/widgets.py
@@ -1,4 +1,5 @@
 from flask import json
+from jinja2 import escape
 from wtforms.widgets import HTMLString, html_params
 
 from flask_admin._compat import as_unicode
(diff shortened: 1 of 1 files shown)


boolbag:
Nice work, @pawl ! 
Works like a charm. 
I've applied the patch manually, but do you have a sense for when it will be world-available via `pip install --upgrade flask-admin` ?

pawl:
@mrjoes are you ok with merging this one? I tested it to make sure it didn't break the editable list view.

If you have time, it might be a good idea to do a release soon.

mrjoes:
@pawl How we're in terms of completeness of current features? Do you feel we might need to wait for some critical fixes?

pawl:
@mrjoes nothing comes to mind as far as critical fixes. And, I'm not currently working on anything. 

We've made a lot of great progress since the last release. Some ~50 pull requests closed since then.

mrjoes:
Great, I'll prepare release tomorrow. I think it should be 1.3.0 version, due to number of new features in the release. Thoughts?

pawl:
@mrjoes 1.3.0 makes sense, sounds good

PR #1005, opened 2015-08-21
remove first column from list view when it is empty anyway

hugojanruiter:
In the list, when view_details, edit and delete are disabled, the first column is empty. This commit ads an extra if statement to check that case and stops further output of the th and td to remove the column completely. 

The use case: the model that I use has many fields, but I do only want N of them to be able to be edited and just inline editing via column_editable_list. In that case there is no need for the column with the view/edit/delete icons.

### DIFF
--- a/flask_admin/templates/bootstrap2/admin/model/list.html
+++ b/flask_admin/templates/bootstrap2/admin/model/list.html
@@ -62,7 +62,9 @@
                     </th>
                     {% endif %}
                     {% block list_row_actions_header %}
-                    <th class="span1">&nbsp;</th>
+                        {% if admin_view.can_view_details or admin_view.can_edit or admin_view.can_delete %}
+                        <th class="span1">&nbsp;</th>
+                        {% endif %}
                     {% endblock %}
                     {% set column = 0 %}
                     {% for c, name in list_columns %}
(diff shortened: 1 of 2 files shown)


pawl:
There's some discussion about this in these old pull requests: #453 and #452

@mrjoes talks about the trade-off:
* It will negatively affect projects that have all the built-in row actions disabled and have only custom row action buttons. In this case, the developer is going to need to override the changes in this pull request to unhide the row action buttons header.

Personally, I like the idea and I've overriden list.html in all my projects to add this. I don't think it would negatively affect many projects. It's also an easy fix for the projects it does break.

hugojanruiter:
Ah I see, I couldn't find anything on this. Is there any discussion on simply adding a property to the ModelView, similar to column_display_pk, so for example column_display_actions?

pawl:
@hugojanruiter That sounds like a good compromise to me. ```column_display_row_actions``` might be a better name to prevent confusion with the bulk actions under the "with selected" dropdown.

hugojanruiter:
Ok, is it worth creating a pull request for it, or should I just use it for my own projects?

pawl:
Just opened another pull request for this in #1069

Issue #1003, opened 2015-08-21
Issue with Nullable Enum fields: forms don't let you leave them empty


boolbag:
It's quite common to have Enum fields that are Nullable.
At present, when you have a nullable enum field, the forms don't allow you to leave the value empty.

If you run [this gist](https://gist.github.com/boolbag/d1abb807bc401f92a001), in the sample data you will see that an empty value is legit for `gender_if_known` (the Fifi row). 

However, if you try to edit Fifi or to add a record (either directly in the list or by clicking Edit/Create, you will see that you cannot leave `gender_if_known` empty.

PR #1009, opened 2015-08-22
Fix nullable enums and allow blanks in editable list view select fields

pawl:
Fixes the issue @boolbag found in #1003 with nullable enum fields not having a blank option.

Also noticed that "allow_blank" wasn't working with the select fields in the editable list view. That's fixed too.

Thanks for finding this @boolbag!


boolbag:
@pawl 
Thank you!
That works like a charm. :)

PR #1017, opened 2015-08-24
responsive tables for screens smaller than 800px wide Fix #1012 

### LINKED ISSUE #1012: Responsive Tables in ModelViews
Hi, A user of QuokkaCMS opened this issue https://github.com/quokkaproject/quokka/issues/82

I think this is a problem we need to fix in Flask-Admin so I researched for options and found this:

http://bootsnipp.com/snippets/featured/no-more-tables-respsonsive-table

http://jasonbradley.me/bootstrap-responsive-tables/

http://stackoverflow.com/questions/28046506/bootstrap-responsive-table-change-rows-with-column

Is that something opened to discussion?

https://camo.githubusercontent.com/43d28380b77a806b00c51c31ddaac41fc171ab6e/68747470733a2f2f662e636c6f75642e6769746875622e636f6d2f6173736574732f313437323732382f313530373330302f64643436393639342d343937332d313165332d383731342d6533643661343939643863312e706e67

rochacbruno:
Related to #1012 
- B2 use no-more-tables approach
- B3 use built-in responsive table

B2
nomoretables

B3
responsive_b3

### DIFF
--- a/flask_admin/static/admin/css/bootstrap2/admin.css
+++ b/flask_admin/static/admin/css/bootstrap2/admin.css
@@ -89,3 +89,53 @@ table.filters tr td {
 .admin-form .controls {
     margin-left: 110px;
 }
+
+@media only screen and (max-width: 800px) {
+
+    /* Force table to not be like tables anymore */
+    #no-more-tables table,
+    #no-more-tables thead,
+    #no-more-tables tbody,
+    #no-more-tables th,
+    #no-more-tables td,
+    #no-more-tables tr {
+        display: block;
+    }
+
+    /* Hide table headers (but not display: none;, for accessibility) */
+    #no-more-tables thead tr {
+        position: absolute;
+        top: -9999px;
+        left: -9999px;
+    }
+
+    #no-more-tables tr { border: 1px solid #ccc; }
+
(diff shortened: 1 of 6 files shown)


petrus-jvrensburg:
Thanks @rochacbruno, I think you're right to give attention to the display on small screens. 

We'll need to put in some work to round off the `no-more-tables` solution, that you implemented here for Bootstrap 2: 

1. We need to figure out what to do with that dangling checkbox and action buttons. Perhaps including them as regular buttons below each record would make the most sense. The tiny icon-buttons don't make a lot of sense on mobile, and it's weird to have them on top of the record.
2. The headers take up a lot of space on the screen. It would be great if we could put more emphasis on  the data.

For the scrolling table, that you implemented for Bootstrap 3:

1. We could add a configuration parameter on the ModelView class for excluding specific columns on mobile. Maybe `column_hide_mobile`?
2. Again, we could replace the icon-buttons with regular buttons that are easier to tap.
3. We could replace the checkbox by highlighting selected rows, and listen to tap evens to toggle selection.

jmagnusson:
Why do we still support Bootstrap 2? Version 4 (!!) is soon in beta...

PR #1026, opened 2015-08-26
Editable list view fixes (order of choices & relations bug)

### LINKED ISSUE #1019: Issue with Inline Form Choices with Enum Field
In a database, it's normal not to know in what order `enum` fields are shown because they are sorted by their index value. However, using `form_choices`, it is possible to constrain the order of presentation of an `enum` field within the `CREATE` form.

Sadly, this is not working for inline editing. On [this gist](https://gist.github.com/boolbag/ec2bdd25ac5de0c12b63), if you try to edit the `favorite_day` field **inline**, you get a "random" order of days. In contrast, on the `CREATE` field, the options set in `form_choices` are respected.

pawl:
* Fixes #1019 by putting inline editable choices in a list instead of a dict.
* Fixes a bug caused by #1009 that added an extra blank field to inline editable fields (only on relations) when allow_blank was enabled.
* Fixed issue with unicode strings in options breaking x-editable.

### DIFF
--- a/flask_admin/model/widgets.py
+++ b/flask_admin/model/widgets.py
@@ -2,7 +2,7 @@ from flask import json
 from jinja2 import escape
 from wtforms.widgets import HTMLString, html_params
 
-from flask_admin._compat import as_unicode
+from flask_admin._compat import as_unicode, text_type
 from flask_admin.babel import gettext
 from flask_admin.helpers import get_url
 from flask_admin.form import RenderTemplateWidget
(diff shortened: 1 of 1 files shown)


flask_admin/model/widgets.py
```diff
@@ -135,20 +141,21 @@ def get_kwargs(self, subfield, kwargs):
...
         elif subfield.type in ['QuerySelectField', 'ModelSelectField']:
+            # QuerySelectField and ModelSelectField are for relations
             kwargs['data-type'] = 'select'
 
-            choices = {}
+            choices = []
             for choice in subfield:
                 try:
-                    choices[str(choice._value())] = str(choice.label.text)
+                    choices.append({'value': str(choice._value()),
```
mrjoes: Missed this last time. Why not `text_type` from compat?

pawl: @mrjoes Fixed. Also had to serialize with json.dumps to get rid of the "u" that's prepended to unicode strings in choices.

boolbag:
@pawl Beautiful... Really nice to have the fields display in the right order.
Big thanks!
Hope you have a relaxing weekend, you deserved it.

PR #1032, opened 2015-08-31
add a way to add multiple views or links with a single line of code

pawl:
This pull request adds the ability to add multiple views at once, like this:
```
admin.add_views([view1, view2, view3, view4])
```

Here's the old way:
```
admin.add_view(view1)
admin.add_view(view2)
admin.add_view(view3)
admin.add_view(view4)
```

Also does the same for ```add_link``` by adding an ```add_links```.

### DIFF
--- a/flask_admin/base.py
+++ b/flask_admin/base.py
@@ -540,6 +540,22 @@ class Admin(object):
 
         self._add_view_to_menu(view)
 
+    def add_views(self, *args):
+        """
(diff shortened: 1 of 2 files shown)


mrjoes:
We can also use `*args` to avoid explicit array/tuple.

pawl:
@mrjoes Good idea. I just pushed the changes.

PR #1034, opened 2015-08-31
Support editing of JSONType

jmagnusson:


pawl:
You're sure json fields don't require additional validation? Is that handled at the database level?

jmagnusson:
@pawl You're right. This PR would make the items always saved as strings. We would probably have to do `json.loads()` on the input data before saving also. I'll re-open if I get the time to make a proper PR.

PR #1035, opened 2015-08-31
Remove Flask-SQLAlchemy dependency

jmagnusson:
It should be noted that the declarative base still has to be configured like this:

```python
session = scoped_session(...)

class MyBase:
    query = session.query_property()

Base = declarative_base(cls=MyBase)
```


jmagnusson:
Would be nice to have this merged in, @mrjoes. This PR shouldn't cause any problems.

jmagnusson:
This should be good to merge now.

PR #1038, opened 2015-09-01
Implement CSV export for BaseModelView.

tandreas:
This pull request implements a csv export in BaseModelView.

* New attribute can_export enables an Export button and the /export/csv endpoint.  Default is False.

* New attribute export_max_rows will truncate the CSV to a maximum number of rows.
  Necessary for very large tables.  Default is None.

* New attribute column_formatters_export allows custom column formatters for exports.
  This was done in case the current column_formatters inject HTML or use a Jinaj2 Macro.
  Defaults to column_formatters if not set.

* New attribute column_type_formatters_export is by default the same as column_type_formatters
  except that it removes the bool_formatter which injects HTML, called EXPORT_FORMATTERS.

* get_list_value was refactored to allow different formatters and to make having a Jinja2 context optional.
  A new method _get_list_value was implemented that does most of the logic.
  get_list_value and the new method get_export_value call _get_list_value.

* CSV export was implemented in export_csv at the endpoint /export/csv.
  Uses a generator to stream the CSV as to use less memory in the case of large lists.

* csv_encode was added to _compat for Python 2 and Python 3 compatibility with unicode.

* In order to generate an error if a macro is used with column_formatters before streaming 
  occurs, a check is made to throw an error if the function name is "inner".
  This is the name of the function flask_admin.model.template.macro returns.
  I'm open to better ideas to determine if the column_formatter is a macro.

* Changes to test_model.py include making it easier to add your own dict of data to a 
  view and use page_size with get_list.

* Adds tests for CSV export.

Thanks to @pawl for helping!

### DIFF
--- a/doc/introduction.rst
+++ b/doc/introduction.rst
@@ -298,6 +298,12 @@ To **manage related models inline**::
 These inline forms can be customised. Have a look at the API documentation for
 :meth:`~flask_admin.contrib.sqla.ModelView.inline_models`.
 
+To **enable csv export** of the model view::
+
+    can_export = True
+
+This will add a button to the model view that exports records, truncating at :attr:`~flask_admin.model.BaseModelView.max_export_rows`.
+
 Adding Your Own Views
 =====================
 
--- a/flask_admin/_compat.py
+++ b/flask_admin/_compat.py
@@ -31,6 +31,10 @@ if not PY2:
 
         return str(s)
 
+    def csv_encode(s):
+        ''' Returns unicode string expected by Python 3's csv module '''
+        return as_unicode(s)
+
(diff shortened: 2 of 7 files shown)


pawl:
:+1:

klinkin:
Not bad

mrjoes:
This is great and I was thinking about it, but was not sure it should be part of the base model - it is getting huge.

Can it be implemented as a mixin? So anyone can mix this functionality into his base model?

On other hand we don't have machinery to add buttons/dropdown actions for mixed in code.

mrjoes:
Actually, nevermind - it is not that big of a chance.

Good work!

PR #1039, opened 2015-09-01
Closes #1037 -- add support for ``is_created`` to inline forms

### LINKED ISSUE #1037: Add support for `is_created` in `on_model_change` for inline fields
Hi I need to send an email when a certain inline model is created and the only way I could figure out to do that is to use `on_model_change`. Unfortunately for inline fields `is_created` isn't supported. Is there a better way to go about firing that email and/or would you accept a PR that essentially duplicated what you have done in the migration path for adding the `is_created` flag?

jschneier:
Corresponding pull request for my issue. It works fine for me on a project using SQLAlchemy. Not certain what tests you want me to add for things that are in `contrib`, please provide some guidance on that.

### DIFF
--- a/flask_admin/contrib/mongoengine/fields.py
+++ b/flask_admin/contrib/mongoengine/fields.py
@@ -36,13 +36,14 @@ class ModelFormField(InlineFormField):
 
     def populate_obj(self, obj, name):
         candidate = getattr(obj, name, None)
-        if candidate is None:
+        is_created = candidate is None
+        if is_created:
             candidate = self.model()
             setattr(obj, name, candidate)
 
         self.form.populate_obj(candidate)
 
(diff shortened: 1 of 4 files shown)


pawl:
Looks like there aren't any tests for inline models for any of the backends. That would be a great addition.

Maybe something like this, but for inline models?: https://github.com/flask-admin/flask-admin/blob/f9583efc48ee2029f20e28477ca7372468f379c7/flask_admin/tests/sqla/test_basic.py#L1503

pawl:
@jschneier I just tested this locally. Works great.

Also, I was wrong about there not being tests for inline models: https://github.com/flask-admin/flask-admin/blob/master/flask_admin/tests/sqla/test_inlineform.py

PR #1040, opened 2015-09-02
Closes #1033 - pass on form_extra_fields while scaffolding an inline …

### LINKED ISSUE #1033: form_extra_fields not considered in an inline form
[flask_admin.form.upload](http://flask-admin.readthedocs.org/en/latest/api/mod_form_upload/#module-flask_admin.form.upload) related fields are not rendered correctly in an inline form. 

Looks like the *extra_fields* kwarg is not set when invoking *get_form* from the [InlineModelConverter.contribute](https://github.com/flask-admin/flask-admin/blob/f774a37620f9efa9606a7f85b8a60c51e2918f2a/flask_admin/contrib/sqla/form.py#L610) method.  

Can you please point me to the contribution guidelines ?  I can maybe submit a PR for this.

(Note: I posted this on [StackOverflow](http://stackoverflow.com/questions/32300837/using-imageuploadfield-in-an-inlineformadmin-form) after which I looked into the source)

late-warrior:
This change allows form_extra_fields in inline forms

### DIFF
--- a/flask_admin/contrib/sqla/form.py
+++ b/flask_admin/contrib/sqla/form.py
@@ -612,7 +612,8 @@ class InlineModelConverter(InlineModelConverterBase):
                                   only=info.form_columns,
                                   exclude=exclude,
                                   field_args=info.form_args,
-                                  hidden_pk=True)
+                                  hidden_pk=True,
+                                  extra_fields=info.form_extra_fields)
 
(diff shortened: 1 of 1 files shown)


pawl:
Just tested this, works great! Thanks @late-warrior

PR #1052, opened 2015-09-16
Add customization options for export columns

radioprotector:
This is based on what is used for the details view (`get_details_columns`/`column_details_list`/`column_details_exclude_list`) with the added complication that `get_export_columns` also checks
`column_list` before attempting to invoke `scaffold_list_columns`.

The internal `_export_columns` collection is only used at the time of generation so that filters, sorting options, and so on will still use the standard column indices expected for these operations.

### DIFF
--- a/flask_admin/model/base.py
+++ b/flask_admin/model/base.py
@@ -168,6 +168,17 @@ class BaseModelView(BaseView, ActionsMixin):
         Collection of fields excluded from the details view.
     """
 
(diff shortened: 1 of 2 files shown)


radioprotector:
Sorry about all of the build issues. It appears that when I run the nosetests locally, it doesn't incorporate the endpoint configured for each `MockModelView` instance. Is this a known issue or something specific to my environment?

PR #1054, opened 2015-09-18
Change export_max_rows default to unlimited

pawl:
Currently, the csv export will only export the first 20 rows by default. (unless the developer increases export_max_rows)

This probably isn't expected behavior. So, this pull request makes the export unlimited by default.

I also added tests for export_max_rows into each backend's tests, because export_max_rows relies on each backend's implementation of ```get_list()```.


pawl:
@mrjoes are you ok with this change?

mrjoes:
Duh, I thought I merged this one.

PR #1059, opened 2015-09-25
Make QuerySelectMultipleField fields editable on the list view

chase-seibert:
For QuerySelectField and ModelSelectField, a select data-type is used.
This element does not support select multiple, but the checklist element
does. The same code for compiling the choices is used.

### DIFF
--- a/flask_admin/model/widgets.py
+++ b/flask_admin/model/widgets.py
@@ -141,9 +141,11 @@ class XEditableWidget(object):
         elif subfield.type in ['FloatField', 'DecimalField']:
             kwargs['data-type'] = 'number'
             kwargs['data-step'] = 'any'
-        elif subfield.type in ['QuerySelectField', 'ModelSelectField']:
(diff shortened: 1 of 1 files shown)


mrjoes:
Thanks for pull request. Saw your blog post and can't agree it is looking ugly by default... plain, but not ugly :-)

pawl:
@chase-seibert did you test this? I get an error message when I try it: https://i.imgur.com/nkPJoob.png

Form data in the checklist field's post request is an array like this:
```
tags-15[]: 5
tags-15[]: 6
```

But, I don't think the ListEditableFieldList (https://github.com/flask-admin/flask-admin/blob/master/flask_admin/model/fields.py#L129) is handling arrays. There might be other issues too.

Also, it's not pre-selecting fields correctly. We need to fix this too.

chase-seibert:
Ok, will add test and re-submit.

PR #1063, opened 2015-10-02
On adding a text filter, set input focus to allow immediate typing

jnga:
Upon choosing a text filter to add, it would be useful to be able to immediately type into the text box rather than first having to click in the box.


jnga:
The TravisCI build failed upon submitting this pull request, but the failure does not appear to have to do with the pull request. One test failed: flask_admin.tests.test_form_upload.test_image_upload_field

jnga:
Will resubmit this request in order to clean up its commits.

PR #1064, opened 2015-10-02
On adding a text filter, set input focus to allow immediate typing

jnga:
Upon choosing a text filter to add, it would be useful to be able to immediately type into the text box rather than first having to click in it.

### DIFF
--- a/flask_admin/static/admin/js/filters.js
+++ b/flask_admin/static/admin/js/filters.js
@@ -151,7 +151,10 @@ var AdminFilters = function(element, filtersElement, filterGroups, activeFilters
     $('a.filter', filtersElement).click(function() {
         var name = ($(this).text().trim !== undefined ? $(this).text().trim() : $(this).text().replace(/^\s+|\s+$/g,''));
         
-        addFilter(name, filterGroups[name], false, null);
+        // allow immediate typing in text filter box
+        if ($filterField.attr('type') === 'text') {
+            $filterField.focus();
+        }
         
         $('button', $root).show();        
     });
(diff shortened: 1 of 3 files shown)


jnga:
The CI build failed on submitting this pull request, but the failure appears unrelated to this pull request.

One test failed: flask_admin.tests.test_form_upload.test_image_upload_field

pawl:
Looks like the new Pillow 3.0.0 introduced the bug. The error I get when I try to save images looks like this: "Invalid image: cannot identify image file <FileStorage: u'copyleft.tiff' ('image/tiff')>"

 One option is to set Pillow in the requirements to 2.9.0, but I'm looking into another fix.

pawl:
@jnga can you also bump filters.js version to 1.0.1 here?: https://github.com/flask-admin/flask-admin/blob/e8b1b6b5aca047a4f83d685686d2632ce6206233/flask_admin/templates/bootstrap3/admin/model/list.html#L197

and here: https://github.com/flask-admin/flask-admin/blob/e8b1b6b5aca047a4f83d685686d2632ce6206233/flask_admin/templates/bootstrap2/admin/model/list.html#L197

pawl:
@jnga Looks like this was already added in https://github.com/flask-admin/flask-admin/pull/1015

Your change removes "addFilter" and prevents the filter field from getting added at all.

You might need to install the newest version of flask-admin?: ```pip install -e git+https://github.com/flask-admin/flask-admin.git@master#egg=flask-admin```

jnga:
@pawl Oops, I mistakenly dropped the addFilter line while trying to clean up my prior pull request #1063. I'll try the newer version, though--thanks very much.

PR #1069, opened 2015-10-04
add column_display_actions option to ModelView

pawl:
Adds the column_display_actions option to ModelView to allow for hiding the row actions column without modifying list.html. Useful for when you don't want the blank column to appear when you have can_edit, can_delete, and can_view_details disabled.

column_display_actions=True (default): 
true

column_display_actions=False:
false

This approach preserves backwards compatibility for projects that have can_edit + can_delete disabled and have custom row actions.

Discussion about this is here: #1005

### DIFF
--- a/flask_admin/model/base.py
+++ b/flask_admin/model/base.py
@@ -432,6 +432,20 @@ class BaseModelView(BaseView, ActionsMixin):
         Controls if the primary key should be displayed in the list view.
     """
 
+    column_display_actions = True
+    """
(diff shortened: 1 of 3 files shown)


pawl:
Probably a good idea to review this one with whitespace differences disabled: https://github.com/flask-admin/flask-admin/pull/1069/files?w=1

mrjoes:
@pawl can you resolve conflicts?

pawl:
@mrjoes Done

PR #1073, opened 2015-10-06
Fix 1072: Use mapper.primary_key_from_instance in QueryAjaxModelLoader

### LINKED ISSUE #1072: form_ajax_refs to model with primary key attribute not matching the column in the DB
When I have the following models:

    class Country(db.Model):

        name = db.Column(db.Text)
        alpha2 = db.Column('iso3166alpha2', db.Text)
        alpha3 = db.Column('iso3166alpha3', db.Text)
        numeric = db.Column('iso3166numeric', db.Integer, primary_key=True, autoincrement=False)

    class Customer(db.Model):
        id = db.Column('customer_id', db.Integer, primary_key=True, nullable=False)
        name = db.Column(db.Text)
        country_number = db.Column('country', db.Integer, db.ForeignKey('country.iso3166numeric'), nullable=False)
        country = db.relationship('Country', foreign_keys=country_number)

And I try to register Customer model in administration like that

    class CustomerModelView(StaffModelView):
        form_ajax_refs = {
            'country': {
                'fields': ['alpha3', 'name'],
                'page_size': 10
            }
        }

I receive an error:

    AttributeError: 'Country' object has no attribute 'iso3166numeric'

The problem is obviously that attribute name of the primary key of the Country model does not match the name of the column. But it is supported by the SQLAlchemy.

The problem occurs in QueryAjaxModelLoader.format method https://github.com/flask-admin/flask-admin/blob/v1.3.0/flask_admin/contrib/sqla/ajax.py#L50-L54

I found that in SQLAlchemy there is a way to get the value of primary key by using `primary_key_from_instance` method of the `Mapper`

I will try to provide a pull request to show what I mean.

vstoykov:
fixes #1072

### DIFF
--- a/flask_admin/contrib/sqla/ajax.py
+++ b/flask_admin/contrib/sqla/ajax.py
@@ -22,9 +22,11 @@ class QueryAjaxModelLoader(AjaxModelLoader):
         if not self.fields:
             raise ValueError('AJAX loading requires `fields` to be specified for %s.%s' % (model, self.name))
 
+        self._mapper = model._sa_class_manager.mapper
+
         self._cached_fields = self._process_fields()
 
-        primary_keys = model._sa_class_manager.mapper.primary_key
+        primary_keys = self._mapper.primary_key
         if len(primary_keys) > 1:
             raise NotImplementedError('Flask-Admin does not support multi-pk AJAX model loading.')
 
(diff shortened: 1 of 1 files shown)


pawl:
Closing this pull request in favor of this one: https://github.com/flask-admin/flask-admin/pull/1111

Great work figuring this one out! Thank you for your contribution.

PR #1085, opened 2015-10-15
Fix DeprecationWarning

mikelambert:
flask_admin/model/base.py:949: DeprecationWarning: Required is going away in WTForms 3.0, use DataRequired
(though in this case, I believe the correct choice is InputRequired, though that only matter if an object has id=0)

### DIFF
--- a/flask_admin/model/base.py
+++ b/flask_admin/model/base.py
@@ -10,7 +10,7 @@ from flask import (request, redirect, flash, abort, json, Response,
 from jinja2 import contextfunction
 from wtforms.fields import HiddenField
 from wtforms.fields.core import UnboundField
-from wtforms.validators import ValidationError, Required
+from wtforms.validators import ValidationError, InputRequired
 
 from flask_admin.babel import gettext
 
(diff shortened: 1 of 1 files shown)


flask_admin/model/base.py
```diff
@@ -10,7 +10,7 @@
 from jinja2 import contextfunction
 from wtforms.fields import HiddenField
 from wtforms.fields.core import UnboundField
-from wtforms.validators import ValidationError, Required
+from wtforms.validators import ValidationError, InputRequired
```
pawl: http://wtforms.readthedocs.org/en/latest/whats_new.html says: "wtforms.validators.Required is now wtforms.validators.DataRequired"

Shouldn't this be DataRequired instead?

mikelambert: Yes, it does say that, and I'm not sure why it prescribes a single choice. From the docs for InputRequired:
"""
class wtforms.validators.InputRequired(message=None)
...
Note there is a distinction between this and DataRequired in that InputRequired looks that form-input data was provided, and DataRequired looks at the post-coercion data.
"""

In particular, if the input is for record id=0, the post-coercision data will evaluate to False. And so DataRequired would fail on this input, whereas InputRequired would succeed. Since we should allow operations on records with id=0, I believe InputRequired is the correct validator.

This stackoverflow clarified the difference better for me:
http://stackoverflow.com/questions/23982917/flask-wtforms-difference-between-datarequired-and-inputrequired

pawl: Good point: https://i.imgur.com/1vZTGbE.png

Also tested InputRequired and it worked for id=0. I'll merge it.

PR #1087, opened 2015-10-18
Add filter option to apply to the sqla ajax loader results

dzhou:
We saw a lot of cases where we needed some predefined filters on the ajax autocomplete rather than always searching across the entire table. This is useful for applying additional business logic on the autocomplete and make it more user friendly. 

A simple example here is to filter out entries that are marked as inactive or deleted.

``` python
    form_ajax_refs = {
        'active_user': QueryAjaxModelLoader('user', db.session, User,
                                     filters=["is_active=True", "id>1000"])
    }
```


petrus-jvrensburg:
Thanks @dzhou 👍

PR #1089, opened 2015-10-19
Do not disclose the file system full path

porduna:
For example, when uploading a file called "foo/bar.txt" twice, right now it shows the full directory (e.g., "File /your/system/directory/foo/bar.txt already exists"), instead of only showing "File foo/bar.txt already exists".


pawl:
Just tested this out with examples/file/app.py, but didn't see the issue because the example uses a relative path. But, I was able to reproduce the issue after changing it to an absolute path: https://i.imgur.com/sBl9pnU.png

Here's what it looks like after the changes: https://i.imgur.com/wBKakhT.png

Notice it says "Successfully saved file" because we don't throw an exception in _save_form_files... I'll submit a pull request to fix that.

PR #1102, opened 2015-11-05
Fix #1101 form.js addInlineField for new nested elements

### LINKED ISSUE #1101: form.js addInlineField creates incorrect id for new nested elements
The problem appears with rendering new nested elements for MongoEngine.
When you create a new nested element for another new element (not yet saved) form.js puts an extra "-" in the id and this results in the parent element saved empty (the nested elements fail).

For example I use mapping for some elements with a Mongo field like this:
ListField(ListField(StringField())).

I solved this problem by adding additional check that elID exists (it doesn't exist yet for the new parent element):
https://github.com/ngr/flask-admin/commit/36d4b33bae4f45b4af131a32737edadbfcea2e81

Is this OK and should I create a pull request?

ngr:
The problem appears with rendering new nested elements for MongoEngine.
When you create a new nested element for another new element (not yet saved) form.js puts an extra "-" in the id and this results in the parent element saved empty (the nested elements fail).

For example I use mapping for some elements with a Mongo field like this:
ListField(ListField(StringField())).

I solved this problem by adding additional check that elID exists (it doesn't exist yet for the new parent element)

### DIFF
--- a/flask_admin/static/admin/js/form.js
+++ b/flask_admin/static/admin/js/form.js
@@ -484,7 +484,10 @@
         var $parentForm = $el.parent().closest('.inline-field');
 
         if ($parentForm.hasClass('fresh')) {
-          id = $parentForm.attr('id') + '-' + elID;
+          id = $parentForm.attr('id');
+          if (elID) {
+            id += '-' + elID;
+          }
         }
 
         var $fieldList = $el.find('> .inline-field-list');


pawl:
Just tested this, confirmed the new nested inline forms are saving now. And it doesn't appear to break non-nested inline forms.

The name is missing on new nested inline fields: https://i.imgur.com/jzidgQN.png But, at least it's saving now.

PR #1110, opened 2015-11-14
Add nowrap to action icons table cell

mdanse:
This was added in the past to the bootstrap2 css but not to the bootstrap3 css.


pawl:
Looks better to me.

Before: 
Before

After: 
After

PR #1114, opened 2015-11-20
fix inet types converter, search by non-text fields

kalombos:
Hello, i've added 2 features
1) Fix AdminModelConverter for some postgresql types. If don't do it then fields with type Mac, Inet will not be shown in edit form view.
2) Added cast any type of column to unicode. I think it would be nice to search by any non-text field.

### DIFF
--- a/flask_admin/contrib/sqla/view.py
+++ b/flask_admin/contrib/sqla/view.py
@@ -7,6 +7,8 @@ from sqlalchemy.orm import joinedload, aliased
 from sqlalchemy.sql.expression import desc, ColumnElement
 from sqlalchemy import Boolean, Table, func, or_
 from sqlalchemy.exc import IntegrityError
+from sqlalchemy.sql.expression import cast
+from sqlalchemy import Unicode
 
 from flask import flash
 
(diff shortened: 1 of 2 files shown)


pawl:
Looks like tests are failing on this one due to an unrelated issue: https://github.com/coleifer/peewee/issues/765

flask_admin/contrib/sqla/view.py
```diff
@@ -828,11 +830,11 @@ def _apply_search(self, query, count_query, joins, count_joins, search):
                                                                                    inner_join=False)
 
                 column = field if alias is None else getattr(alias, field.key)
-                filter_stmt.append(column.ilike(stmt))
+                filter_stmt.append(cast(column, Unicode).ilike(stmt))
```
pawl: My first reaction to this was "that looks bad for performance". But, I checked and MySQL isn't using an index for these searches anyway, because we're using `LIKE lower("%value%")`. I'm not noticing a speed difference before/after the change when doing a search on a datetime column with 50k rows.

I confirmed it prevents the exception when searching INET columns (and other columns?), so it seems like a good change.

PR #1133, opened 2015-12-10
make form_widget_args working for inline_models

immon:
f.name for inlined_models is something like parent-idx-slave, so we
should use f.short_name to get proper widget_args


mrjoes:
Nice catch, thanks!

PR #1139, opened 2015-12-12
add support for new version of flask-login and flask-security

### LINKED ISSUE #1116: Flask-Login's UserMixin.is_* are now properties
Flask-Login's UserMixin.is_* are now properties, so in the [Flask-Login auth example](https://github.com/flask-admin/flask-admin/blob/master/examples/auth-flask-login/app.py),

`login.current_user.is_authenticated()`

should be

`login.current_user.is_authenticated`

pawl:
Fixes #1116 

* Changes is_active() and is_authenticated() to properties
* Removes specific versions specified in requirements.txt files

### DIFF
--- a/examples/auth-flask-login/app.py
+++ b/examples/auth-flask-login/app.py
@@ -95,7 +95,7 @@ def init_login():
 class MyModelView(sqla.ModelView):
 
     def is_accessible(self):
-        return login.current_user.is_authenticated()
+        return login.current_user.is_authenticated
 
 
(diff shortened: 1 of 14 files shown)


mrjoes:
We need to pin version (using >=) because we broke compatibility with older Flask-Login version: .is_authenticated is method and boolean comparison will always return True.

pawl:
@mrjoes Good catch. Just changed it.

PR #1142, opened 2015-12-14
Friendly choices for editable boolean field.

iurisilvio:
`Yes / No` choices are better than `True / False`.

### DIFF
--- a/flask_admin/model/widgets.py
+++ b/flask_admin/model/widgets.py
@@ -109,7 +109,10 @@ class XEditableWidget(object):
         elif subfield.type == 'BooleanField':
             kwargs['data-type'] = 'select'
             # data-source = dropdown options
-            kwargs['data-source'] = {'': 'False', '1': 'True'}
+            kwargs['data-source'] = json.dumps([
+                {'value': '', 'text': gettext('No')},
+                {'value': '1', 'text': gettext('Yes')}
+            ])
             kwargs['data-role'] = 'x-editable-boolean'
         elif subfield.type == 'Select2Field':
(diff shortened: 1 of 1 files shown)


flask_admin/model/widgets.py
```diff
@@ -109,7 +109,10 @@ def get_kwargs(self, subfield, kwargs):
         elif subfield.type == 'BooleanField':
             kwargs['data-type'] = 'select'
             # data-source = dropdown options
-            kwargs['data-source'] = {'': 'False', '1': 'True'}
+            kwargs['data-source'] = json.dumps([
+                {'value': '0', 'text': gettext('No')},
```
pawl: better to use lazy_gettext here, like this: https://github.com/iurisilvio/flask-admin/blob/translate_editable_boolean/flask_admin/model/filters.py#L98

iurisilvio: The `lu'string'` is not JSON serializable, so the `json.dumps` crashes if I use the `lazy_gettext`, so I have to evaluate the lazy string before serialization.

pawl: Oops, looks good then.

Any idea why it's using "0" now instead of ""? Does that cause any issues?

iurisilvio: The `"0"` does not work the way I expected. Going back to the empty string.

iurisilvio:
@pawl I reverted to use empty string instead of `"0"`.

It still has a bug, the current value is not selected. It already exists, because the widget value is an HTML string to render the icon. I don't know yet how to fix this.

PR #1147, opened 2015-12-18
Fix db.Property index check.

mikelambert:


mikelambert:
Ping?

pawl:
I was looking at the docs and saw examples of both indexed and _indexed:
* indexed is used here: https://cloud.google.com/appengine/docs/python/ndb/properties
* _indexed is used here: https://cloud.google.com/appengine/docs/python/ndb/modelclass

I wanted to test the change, so I started making an example for appengine based off the sqla/app2.py example. But, I didn't finish: https://gist.github.com/pawl/7d6064aebac5ed9eb2b8

Can you give an example of how I can test this? Or maybe an explanation about why the change is necessary?

mikelambert:
indexed is the argument name used when calling the initializer, but _indexed is the property name used when accessing post-initialization. The documentation is unclear in that regard with respect to 'indexed', but isn't explicitly wrong, since I can read it as referring to the property initializer options, in which 'indexed' is correct.

For reference, here is the actual NDB code:
https://github.com/GoogleCloudPlatform/datastore-ndb-python/blob/master/ndb/model.py#L909

mrjoes:
But we're checking property value, so it should stay `_indexed`?

mikelambert:
Woops thanks @mrjoes . Sorry, @pawl and I got confused here with db vs ndb.

@pawl looked into ndb.Property's code and documentation, which I got distracted by as well. My previous comment proves that yes, ndb.Property should remain as _indexed, like you said.

But my change was actually for *db.Property*, for which we need to use "indexed". You can see here that db.Property uses "indexed" for both the property initializer as well as the object's property name post-construction:
https://code.google.com/p/googleappengine/source/browse/trunk/python/google/appengine/ext/db/__init__.py#571

PR #1152, opened 2015-12-21
File admin refactoring to support multiple backends

ArthurPBressan:
This is the first step for adding support for different kinds of file back ends, say for example, Amazon S3, Dropbox and so on.

We already have a working implementation of a S3 storage object, but we first want to know if there's any interest in this refactoring.

### DIFF
--- a/flask_admin/contrib/fileadmin.py
+++ b/flask_admin/contrib/fileadmin.py
@@ -1,11 +1,11 @@
+from datetime import datetime
 import os
 import os.path as op
 import platform
 import re
 import shutil
-
-from datetime import datetime
 from operator import itemgetter
+
 from werkzeug import secure_filename
 
 from flask import flash, redirect, abort, request, send_file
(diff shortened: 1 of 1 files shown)


mrjoes:
This is very cool. Thank you!

ArthurPBressan:
Yeah, I wasn't sure how to handle that one way or the other. FileAdmin is kinda tight with the file system for obvious reasons, and for the local files it makes a lot of sense to have the base path.

The main issue that I have with the `base_path` is that I couldn't remove it from the constructor without breaking existing code.

mrjoes:
Probably building new version of the file admin as a separate package is better option (and include alternative backends). But since pull request didn't really break anything (hopefully - I was not able to find any problems so far + all tests passed) - I don't see a why not to merge the pull request.

ArthurPBressan:
Sure, we'll build a separate package then.

Thank you! :smile:

mrjoes:
Another option would be creation of the base file admin class without file-specific constructor options (`base_path`, etc), hide `base_path` logic inside of file storage backend and make `FileAdmin` backwards-compatible by having something like this:

```python
class FileAdmin(BaseFileAdmin):
    storage_provider = LocalFileStorage

    def __init__(self, base_path, ....):
        self.storage = self.storage_provider(base_path, ...)
```

This way there's no need to create new package.

PR #1154, opened 2015-12-22
Adds S3 storage backend

ArthurPBressan:
Adds a working backend implementation for AWS S3.


mrjoes:
Can you put it in a separate file because of additional dependencies?

pawl:
@ArthurPBressan can you also add an s3 backend example to the "Managing Files & Folders" section of the docs?: http://flask-admin.readthedocs.org/en/latest/advanced/#managing-files-folders

This is a really great idea, thanks for adding it. Might want to add your name to AUTHORS file too.

Next up, file admin pagination?

flask_admin/contrib/fileadmin/s3.py
```diff
@@ -0,0 +1,161 @@
...
+            path += self.separator
+        for key in self.bucket.list(path, self.separator):
+            if key.name == path:
+                continue
+            if isinstance(key, Prefix):
+                name = _remove_trailing_slash(_strip_path(key.name, path))
+                key_name = _remove_trailing_slash(key.name)
+                directories.append((name, key_name, True, 0, 0))
+            else:
+                last_modified = int(dateutil.parser.parse(key.last_modified)
```
pawl: Can this be done with time.strptime and without the dateutil dependency?

flask_admin/contrib/fileadmin/s3.py
```diff
@@ -0,0 +1,161 @@
...
+from flask_admin.babel import gettext
+
+
+class S3Storage(object):
+    """
+        Storage object representing files on an Amazon S3 bucket.
+
+        Usage::
+
+            from flask_admin.contrib.fileadmin import BaseFileAdmin,
```
pawl: extra comma at the end of this line that can be removed

flask_admin/contrib/fileadmin/s3.py
```diff
@@ -0,0 +1,161 @@
...
+            from flask.admin.contrib.fileadmin.s3 import S3Storage
```
pawl: `flask.admin` should be `flask_admin` like in the line above

pawl:
I'm getting "Not Found" when I try to rename or delete files, but uploading works fine. Here's a gist of the code I'm using to test: https://gist.github.com/pawl/717fb9edde7bb6748427

mrjoes:
Also, I'd suggest adding convenience class, like current FileAdmin - S3FileAdmin. This way users won't have to create storage instance and use BaseFileAdmin class.

pawl:
@mrjoes :+1: was just about to say that. It would make usage even simpler.

flask_admin/contrib/fileadmin/s3.py
pawl: Here's an example that uses strptime: https://github.com/jordansissel/s3cleaner/blob/5969afaf5edf0223f20e13a21aa190d20c6b958a/s3cleaner.py#L46

ArthurPBressan:
@pawl Should be working now. :+1:

mrjoes:
@ArthurPBressan excellent, thank you.

Merged and will be in a new Flask-Admin release, which I was delaying for this feature.

ArthurPBressan:
Glad to help. :smile:

PR #1156, opened 2015-12-24
Add a demo app for appengine (and fix one more bug)

mikelambert:
@pawl requested some additional cleanup in the appengine library here https://github.com/flask-admin/flask-admin/pull/796 , which I've written up in this pull request.


pawl:
Just tried out the example, works great! This will be really helpful for new users + testing changes. Merging now.

PR #1159, opened 2015-12-30
SQLA get_count_query uses get_query() instead of model

fasouto:
Instead of override the get_count_query method every time by default it gets the model query. I use soft delete in my models and this was definitely useful.

I can write some test if you want.


mrjoes:
There's a reason why it is done this way. See: https://github.com/flask-admin/flask-admin/pull/198

PR #1164, opened 2016-01-04
Prevent showing 'save and ~' buttons when actions disabled

### LINKED ISSUE #1163: "Save and Add Another" button should not be shown if can_create = False
If I create a model view and set can_create = False then I wouldn't expect to see the button 'Save and Add Another' since creates are not allowed.

pawl:
Fixes #1163

Will prevent showing "Save and Add Another" and "Save and Continue Editing" when their respective actions are disabled.

Also moves `extra()` to lib.html while keeping backward compatibility.

@billmccord Good catch!


billmccord:
@pawl Wow, so fast! Thanks!

PR #1165, opened 2016-01-05
Update details.html

mdanse:
{{ super() }} is not called in the tail block of details.html. It should be as currently any base_template code in the tail block will be disregarded.


pawl:
@mdanse can you also add it to the bootstrap 2 details.html?

PR #1174, opened 2016-01-17
SQLA - allow using column objects in column_list

### LINKED ISSUE #1036: SQLA: Support passing in model attributes to `column_list`
I.e. I want to be able to do the following:

```python
class UserView(ModelView):
    column_list = (
        db.User.name,
        db.UserEmail.email,
        db.Role.name,
    )
```

This works for `column_searchable_list` and `column_filters` so I see no reason for it not to work for `column_list`.

pawl:
Fixes #1036

I didn't get it working with model attributes from related models, but I made a unit test for it (and commented it out): https://github.com/flask-admin/flask-admin/compare/master...pawl:fix_sqla_column_list?expand=1#diff-527d8cf625fb5fc58b22130d4ded2201R274

### DIFF
--- a/flask_admin/contrib/sqla/view.py
+++ b/flask_admin/contrib/sqla/view.py
@@ -471,19 +471,56 @@ class ModelView(BaseModelView):
                 if isinstance(c, tuple):
                     column, path = tools.get_field_with_path(self.model, c[1])
                     column_name = c[0]
-                elif isinstance(c, InstrumentedAttribute):
-                    column, path = tools.get_field_with_path(self.model, c)
-                    column_name = str(c)
                 else:
                     column, path = tools.get_field_with_path(self.model, c)
                     column_name = c
 
+                if path:
+                    # column is in another table, use full path as column_name
+                    column_name = text_type(c)
+                    self._sortable_joins[column_name] = path
+                else:
+                    # column is in same table, use only model attribute name
+                    column_name = column.key
+
+                # column_name must match column_name used in `get_list_columns`
                 result[column_name] = column
+            return result
+
+    def get_list_columns(self):
+        """
+            Returns a list of tuples with the model field name and formatted
+            field name. If `column_list` was set, returns it. Otherwise calls
+            `scaffold_list_columns` to generate the list from the model.
+        """
+        if self.column_list is None:
+            columns = self.scaffold_list_columns()
+
+            # Filter excluded columns
+            if self.column_exclude_list:
+                columns = [c for c in columns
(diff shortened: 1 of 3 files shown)


pawl:
I saw the TODO that was related to this: `# TODO: Use SQLAlchemy "path-finder" to find exact join path to the target property`

We might be able to consolidate a good amount of the SQLA code if we had a way to find the relative path from the ModelView's model to any given column (even if it's in a different model). Know of an easier way to do this?

mrjoes:
Not really - for now it is required to use strings and specify full path. Like `Model.foo.bar`.

pawl:
Justed tried a few different cases with this model setup:
```
class Car(db.Model):
    __tablename__ = 'cars'
    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
    desc = db.Column(db.String(50))

class Tyre(db.Model):
    __tablename__ = 'tyres'
    car_id = db.Column(db.Integer, db.ForeignKey('cars.id'), primary_key=True)
    tyre_id = db.Column(db.Integer, primary_key=True)
    car = db.relationship('Car', backref='tyres')
    desc2 = db.Column(db.String(50))
```

|                        | 'desc2'            | 'car.desc'         | 'Tyre.car.desc' | Tyre.car.desc | Tyre.desc2                                   | Car.desc               |
|------------------------|--------------------|--------------------|-----------------|---------------|----------------------------------------------|------------------------|
| column_list            | :white_check_mark: | :white_check_mark: | :x:             | :x:           | :white_check_mark: (after these changes)     | :x:                    |
| column_sortable_list   | :white_check_mark: | :white_check_mark: | :x:             | :x:           | :white_check_mark:                           | :x: (appears sortable) |
| form_columns           | :white_check_mark: | :x:                | :x:             | :x:           | :x:                                          | :x:                    |
| column_searchable_list | :white_check_mark: | :white_check_mark: | :x:             | :x:           | :white_check_mark:                           | :white_check_mark:     |
| column_filters         | :white_check_mark: | :white_check_mark: | :x:             | :x:           | :x: (filter appears, but returns no results) | :white_check_mark:     |

It's not a suprise that 'Tyre.car.desc' and Tyre.car.desc don't work. But I think we need to fix these:
* form_columns doesn't work with 'car.desc'
* form_columns  and column_filters don't work with Tyre.desc2

pawl:
@mrjoes Is this one OK to merge too? I just fixed the merge conflict caused by #1191 and updated the documentation.

jmagnusson:
Nice! Can we get a release for this?

PR #1175, opened 2016-01-17
Add separate form_columns for create and edit forms

### LINKED ISSUE #1044: Separate rules for create and edit.
To remove fields from the create and edit forms:

```
form_excluded_columns = ['last_name', 'email']
```

I want to label edited rows with comments. For example we create row with columns: name, count, payment. After we edit payment column, we MUST append comment about why we change this.

Something like

```
create_form_columns = ('name', 'count', 'payment')
edit_form_columns = ('name', 'count', 'payment', 'comment')

```

How to apply such behavior in flask-admin?

Thanks

[question on stackoverflow](http://stackoverflow.com/questions/32416893/how-to-exclude-column-from-create-but-validate-in-edit-in-flask-admin)

pawl:
Fixes #1044

It will still default to self.form_columns if self.create_form_columns or self.edit_form_columns aren't set.

### DIFF
--- a/flask_admin/contrib/appengine/view.py
+++ b/flask_admin/contrib/appengine/view.py
@@ -31,7 +31,7 @@ class NdbModelView(BaseModelView):
         #TODO: implement
         pass
 
-    def scaffold_form(self):
+    def scaffold_form(self, form_columns=None):
         return wt_ndb.model_form(self.model())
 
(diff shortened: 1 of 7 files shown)


pawl:
Oops, the way I did this isn't backward compatible. Going to do it a different way.

cancan101:
was this ever re-submitted?

PR #1181, opened 2016-01-26
new:support mongenine ReferenceField filter

likaiguo:


flask_admin/contrib/mongoengine/filters.py
```diff
@@ -221,7 +236,21 @@ def operation(self):
         return lazy_gettext('not between')
 
 
+class ReferenceFilter(BaseMongoEngineFilter):
+
+    def clean(self, value):
```
pawl: `validate` will return `false` if `clean` raises a ValueError, but it looks like invalid ObjectIds will return this`InvalidId` exception instead: https://gist.github.com/anonymous/4fc86f0798e481095d24

You're also going to need to override the `validate` method to handle `InvalidId` exceptions.

Here's the validate method you need to override: https://github.com/flask-admin/flask-admin/blob/7445f6a486556ab1faf92fc9180af15ab24f5fd6/flask_admin/model/filters.py#L56

flask_admin/contrib/mongoengine/filters.py
```diff
@@ -221,7 +236,21 @@ def operation(self):
...
+        return ObjectId(value.strip())
+
+    def apply(self, query, value):
+        flt = {'%s' % self.column.name: value}
+        return query.filter(**flt)
+
+    def operation(self):
+        return lazy_gettext('equals')
```
pawl: How about changing this to `ObjectId equals`? Currently, it's hard to tell that it's expecting an ObjectId as input.

likaiguo:
I am a little confused to solve the conflicts,please help me.

likaiguo:
support flask_admin display django model.

pawl:
I resubmitted a pull request for the original topic of this pull request in #1249.

The django model backend looks very interesting. Can you submit a pull request for each change except the new mongoengine referencefield filter?

PR #1184, opened 2016-01-28
Fallback to default location if it was not received from external service

t4ec:
If end-user uses adblock or other extension and access to ip-api.com is blocked map will apper gray and it will not be possible to do anything with it. 

If that happens we need to set default location and zoom out for better view.


pawl:
Works for me. Great work @t4ec

Before:


After:

jmagnusson:
Nice! Can we see a new release for this @pawl?

PR #1190, opened 2016-02-04
The S3 connection cost some money. Lazy connect to S3.

iurisilvio:


pawl:
@iurisilvio  Doesn't this still keep the connection open after it's first needed?

pawl:
Looks like some projects are using contextlib's `closing` for this: https://github.com/fezhou2/tempest/blob/c763c9fad04b28759a2c6d7cbac0ea69b7ed736b/thirdparty/boto/test_s3_objects.py#L44

iurisilvio:
I did this pull request to avoid the connection on start. In some cases, I don't setup my AWS credentials and crash the application.

I don't think I must explicitly close the connection. Your example is a test case, there is a good place to close the connection.

Amazon S3 timeout the connection after some time and Boto must handle it gracefully.

Looks like it is a good practice to close the connection to avoid `CLOSE_WAIT` connections: http://stackoverflow.com/questions/18383839/python-s3-boto-connection-close-causes-an-error

It is easy enough to recycle connection based on time or call counting. Do you want this implemented here?

PR #1191, opened 2016-02-07
SQLA - Allow using column objects in form_columns

### LINKED ISSUE #1176: SQLA - form_columns does not allow '<relation>.<column name in related model>'
See comment here: https://github.com/flask-admin/flask-admin/pull/1174#issuecomment-172404695

form_columns doesn't allow specifying a column in a related model in the same way as column_list, column_filters, column_sortable_list (`'<relation>.<column name in related model>'`)

It gives the following error message: `ValueError: Invalid model property name <class '__main__.Tyre'>.car.desc`

The fix might require using `rec_getattr` in tools.py.

pawl:
Fixes #1176

The main change in this pull request happens in `contrib/sqla/form.py`, which changes form scaffolding to use `_get_field_with_path` like filter and search scaffolding do. This allows us to use model objects in `form_columns` like in `column_filters` and `column_searchable_list`. It will also throw an exception telling you to use `inline_models` if you try to use a column that's in another table.

This required moving `_get_columns_for_field`, `_need_join` and `_get_field_with_path` to `contrib/sqla/tools.py` to allow using `_get_field_with_path` in `form.py`.

### DIFF
--- a/flask_admin/contrib/sqla/form.py
+++ b/flask_admin/contrib/sqla/form.py
@@ -9,13 +9,14 @@ from flask_admin.model.form import (converts, ModelConverterBase,
 from flask_admin.model.fields import AjaxSelectField, AjaxSelectMultipleField
 from flask_admin.model.helpers import prettify_name
 from flask_admin._backwards import get_property
-from flask_admin._compat import iteritems
+from flask_admin._compat import iteritems, text_type
 
 from .validators import Unique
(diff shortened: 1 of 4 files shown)


pawl:
Tests for this passed on my fork: https://travis-ci.org/pawl/flask-admin/builds/107526070

Not sure why they didn't run here.

PR #1198, opened 2016-02-11
Added support for filters with multiple columns

vToMy:
Added support for filters with multiple columns (such as composite filters) while retaining backward-compatibility.

### DIFF
--- a/flask_admin/contrib/sqla/filters.py
+++ b/flask_admin/contrib/sqla/filters.py
@@ -1,3 +1,5 @@
+import collections
+from flask_admin._compat import text_type
 from flask_admin.babel import lazy_gettext
 from flask_admin.model import filters
 from flask_admin.contrib.sqla import tools
@@ -116,6 +118,38 @@ class FilterNotInList(FilterInList):
 
     def operation(self):
         return lazy_gettext('not in list')
+
+
+# Composite filters
+class FilterAndEqual(FilterEqual):
+    def __init__(self, child_filters, name, delimiter=',', options=None, data_type=None):
+        column = [filter_.column for filter_ in child_filters]
+        super().__init__(column, name, options, data_type)
+        self.filters = child_filters
+        self.delimiter = delimiter
+
+    def convert_aliases(self, aliases):
+        is_aliases_sequence = isinstance(aliases, collections.Sequence) and not isinstance(aliases, text_type)
+        aliases = [aliases] * len(self.filters) if not is_aliases_sequence else aliases
+        return aliases
+
+    def apply(self, query, values, aliases=None):
+        aliases = self.convert_aliases(aliases)
+        for filter_, value, alias in zip(self.filters, values, aliases):
+            query = filter_.apply(query, value, alias)
+        return query
+
+    def clean(self, value):
+        result = value.split(self.delimiter)
+        if len(result) != len(self.filters):
+            names = [filter_.name for filter_ in self.filters]
+            value_format = self.delimiter.join(names)
+            raise ValueError('Value format should be: {value_format}.'.format(value_format=value_format))
+        return result
+
+    def get_column(self, aliases):
+        aliases = self.convert_aliases(aliases)
+        return [filter_.get_column(alias) for filter_, alias in zip(self.filters, aliases)]
 
 
 # Customized type filters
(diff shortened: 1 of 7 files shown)


flask_admin/contrib/sqla/view.py
```diff
@@ -853,27 +854,37 @@ def _apply_filters(self, query, count_query, joins, count_joins, filters):
...
-
-                if count_query is not None:
-                    count_query, count_joins, count_alias = self._apply_path_joins(
-                        count_query,
-                        count_joins,
-                        path,
-                        inner_join=False)
+                aliases = []
+                count_aliases = []
+                is_multiple_columns = isinstance(flt.column, collections.Sequence) and not isinstance(flt.column, str)
```
pawl: Probably need to use `text_type` here instead of `str` for python 3 compatibility? text_type comes from `from flask_admin._compat import text_type`

pawl:
Can you add a test for this?

vToMy:
I added a test and a basic composite filter.

flask_admin/contrib/sqla/filters.py
```diff
@@ -118,6 +120,38 @@ def operation(self):
         return lazy_gettext('not in list')
 
 
+# Composite filters
+class FilterAndEqual(FilterEqual):
+    def __init__(self, child_filters, name, delimiter=',', options=None, data_type=None):
```
pawl: This would be more user-friendly with `data_type='select2-tags'` (to indicate to users that they need to use a comma separated list).

flask_admin/contrib/sqla/filters.py
```diff
@@ -118,6 +120,38 @@ def operation(self):
...
+        for filter_, value, alias in zip(self.filters, values, aliases):
+            query = filter_.apply(query, value, alias)
+        return query
+
+    def clean(self, value):
+        result = value.split(self.delimiter)
+        if len(result) != len(self.filters):
+            names = [filter_.name for filter_ in self.filters]
+            value_format = self.delimiter.join(names)
+            raise ValueError('Value format should be: {value_format}.'.format(value_format=value_format))
```
pawl: This message doesn't make it to the UI. It's caught by `validate` and it says something like this instead `Invalid Filter Value: test`. This is an important message too, because there's nothing else indicating how many items are required.

flask_admin/contrib/sqla/filters.py
```diff
@@ -118,6 +120,38 @@ def operation(self):
         return lazy_gettext('not in list')
 
 
+# Composite filters
+class FilterAndEqual(FilterEqual):
```
pawl: This filter is essentially the same as adding two filters from the user interface? If that's the case, using the two seperate filters seems like it would be much easier to understand.

vToMy: It's the same as adding "n" filters, but basically yes.
I think it's sometimes easier for the user to use one filter if they have some kind of identifier they could paste, instead of opening up multiple filters.

flask_admin/contrib/sqla/filters.py
vToMy: I see. This is actually a deeper problem, but I have an idea of how to solve it.

flask_admin/contrib/sqla/filters.py
vToMy: Sounds good, thanks for the suggestion.

vToMy:
Incorporated said changes.

pawl:
The commits to fix hybrid properties need to be on a separate branch (and in a separate pull request). I didn't see any errors with hybrid properties after the changes in #1209, but I was only testing with column_list and column_sortable_list.

I'm going to defer the decision to merge this one to @mrjoes. I'm a bit worried about the increase in the complexity of the filter code and how difficult it is to use the added functionality (vs something like a menu link to multiple filters already set, or just having the user set multiple filters).

vToMy:
Sorry about that. Should you decide to not merge this one I'll have another PR for those fixes. It was just difficult testing the changes when things suddenly broke.
I agree about the increase in complexity - I didn't really liked the idea of using a "list" return type for thing like "get_column" but I couldn't think of anything better. However this is useful in cases where, for instance, you have some composite PK and the user wants to query for it in a single field, instead of creating multiple filters, which is clumsy and error-prone - the user is sure to mistake the order of the different parts of the composite PK with the different filters.

PR #1203, opened 2016-02-15
Tablib export

iurisilvio:


mrjoes:
Maybe we should use http://docs.python-tablib.org/en/latest/? It supports export to all major spreadsheet formats without any third-party dependencies.

iurisilvio:
Yes, it is a good option. I'm an almost inactive tablib core commiter. 

I didn't used it because it does not support generators, like the CSV export works. 

For XLSX, it is not easy to do like a generator and tablib is a good fit, but other formats can work like export_csv. 

I can fallback to tablib for formats not defined as export_xxx. I'll rewrite this pull request to use tablib.

iurisilvio:
Done. Export using tablib. Set the `export_type` and it'll just work for any tablib type.

I'm not sure if tablib must be a hard dependency.

flask_admin/model/base.py
```diff
@@ -2040,6 +2045,28 @@ def export_csv(self):
...
+        return count, data
+
+    @expose('/export/<export_type>/')
+    def export(self, export_type):
+        if export_type == 'csv':
+            return self.export_csv()
+        else:
+            return self._export_tablib(export_type)
+
+    @expose('/export/csv/')
```
pawl: This `expose` isn't necessary anymore. Can you remove this line?

flask_admin/model/base.py
```diff
@@ -2040,6 +2045,28 @@ def export_csv(self):
...
+    def export_csv(self):
+        """
+            Export a CSV of records.
+        """
+        return_url = get_redirect_target() or self.get_url('.index_view')
+
+        if not self.can_export:
```
pawl: This "permission denied" code is in both `export_csv` and `_export_tablib`, you can consolidate in `export` now.

pawl:
I submitted a pull request to @iurisilvio's branch with some changes (and the ability to add multiple export_types): https://github.com/iurisilvio/flask-admin/pull/1

Great work! @iurisilvio

iurisilvio:
Thanks @pawl! I merged your PR and it is already here. You implemented my next steps. :+1:

PR #1205, opened 2016-02-16
Allow adding sub-categories using bootstrap v2

bepetersn:
If this is of general use to others, I would be happy to adapt it to meet your project's requirements... add tests, etc. I found it useful.

### DIFF
--- a/flask_admin/base.py
+++ b/flask_admin/base.py
@@ -9,7 +9,7 @@ from flask_admin._compat import with_metaclass, as_unicode
 from flask_admin import helpers as h
 
 # For compatibility reasons import MenuLink
-from flask_admin.menu import MenuCategory, MenuView, MenuLink
+from flask_admin.menu import MenuCategory, MenuView, MenuLink, SubMenuCategory
 
 
 def expose(url='/', methods=('GET',)):
--- a/flask_admin/base.py
+++ b/flask_admin/base.py
@@ -557,6 +557,25 @@ class Admin(object):
         for view in args:
             self.add_view(view)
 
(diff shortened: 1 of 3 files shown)


pawl:
Can you give an example of how this would be used?

bepetersn:
Here's a screenshot:
screenshot from 2016-02-22 12-39-47

And to get that effect:

```python 

admin.add_view(LocationVerticalView(category='Locations'))
...
admin.add_sub_category(name='Potential', parent_name='Locations')
admin.add_view(PotentialLocationView(category='Potential'))
admin.add_view(PotentialLocationMessageView(category='Potential'))
```

pawl:
It's a good idea, but I think this needs to be implemented in a way that also allows for sub-sub-menus too. Maybe something like the way Flask-menu allows specifying a path to a menu item?: https://flask-menu.readthedocs.org/en/latest/#flask_menu.register_menu

bepetersn:
Actually this method does let you add arbitrarily nested sub-categories:

screenshot from 2016-04-08 16-30-29

The only additional lines of coded added from above were:

``admin.add_sub_category(name='Something', parent_name='Suggestable')
admin.add_view(AdminFraudDetails(category='Something')) ``

petrus-jvrensburg:
Could someone expand this so it works on Bootstrap3?

bepetersn:
Maybe some snippet like this could do it: https://www.gaslampmedia.com/multilevel-dropdown-menus-bootstrap-3-x/

bepetersn:
Context: https://github.com/twbs/bootstrap/issues/424

PR #1209, opened 2016-02-19
SQLA - fix hybrid property support after column object support changes

### LINKED ISSUE #1206: Recent commits broke support for hybrid properties expressions
For instance:
```
@hybrid_property
def foo(self):
    return some_value

@foo.expression
    return select(...)
```

No longer works. It seems contrib/sqla/view/get_list_columns (and maybe related methods) can't handle the "select" type that is being returned by the expression. This used to work. I suggest adding a test for it.

To be more specific, it worked before: https://github.com/flask-admin/flask-admin/commit/459010b0def9da74836f1379126434c5f922ba0c

pawl:
Fixes #1206

`column.key` is `None` for hybrid properties and the change in #1174 wasn't expecting that.

This pull request changes it to fail over to the column name in column_list (and column_sortable_list) if `column.key` is `None`.

### DIFF
--- a/flask_admin/contrib/sqla/view.py
+++ b/flask_admin/contrib/sqla/view.py
@@ -484,7 +484,10 @@ class ModelView(BaseModelView):
                                     "Failed on: {0}".format(c))
                 else:
                     # column is in same table, use only model attribute name
-                    column_name = column.key
+                    if getattr(column, 'key') is not None:
+                        column_name = column.key
+                    else:
+                        column_name = text_type(c)
 
                 # column_name must match column_name used in `get_list_columns`
                 result[column_name] = column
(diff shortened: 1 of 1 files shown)


vToMy:
Changing the lines to:
`column.key if hasattr(column, 'key') and column.key else text_type(c)`
Fixes the issue for me.

pawl:
Just made this change, @mrjoes can you merge?

vToMy:
You should use `hasattr` (or `getattr` with a default None value) since, as I said, sqlalchemy's `Select` object doesn't even have a key attribute, which will cause an exception in `getattr`

pawl:
@vToMy I ended up making that change here: https://github.com/flask-admin/flask-admin/pull/1240/files

PR #1211, opened 2016-02-20
resolve full path macro, used by column_formatters

nypgand1:
Resolve full path macro. (used by column_formatters)

### DIFF
--- a/flask_admin/model/template.py
+++ b/flask_admin/model/template.py
@@ -1,16 +1,27 @@
 def macro(name):
-    '''
-        Jinja2 macro list column formatter.
+    """
+        Resolve macro in a Jinja2 context
 
+        :param context:
+            Jinja2 context
         :param name:
-            Macro name in the current template
-    '''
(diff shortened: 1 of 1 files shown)


pawl:
@nypgand1 is there a way you can use the existing rec_getattr code for this?

https://github.com/flask-admin/flask-admin/blob/29f8e0d4de9e1490989b3fc96d53765af6acb44b/flask_admin/tools.py#L67

flask_admin/model/template.py
```diff
@@ -1,16 +1,27 @@
 def macro(name):
-    '''
-        Jinja2 macro list column formatter.
+    """
+        Resolve macro in a Jinja2 context
 
+        :param context:
```
pawl: This param doesn't exist.

pawl:
Nevermind my previous comment about rec_getattr. Can you explain the use case for this? Is for using macros within other macros? Or macros in other templates?

PR #1213, opened 2016-02-24
Fixed bugs

vToMy:
- changed date filter value type to date - this caused incorrect filtering in date filters.
- fixed test_base on windows.

### DIFF
--- a/flask_admin/model/filters.py
+++ b/flask_admin/model/filters.py
@@ -160,7 +160,7 @@ class BaseDateBetweenFilter(BaseFilter):
         Apply method is different for each back-end.
     """
     def clean(self, value):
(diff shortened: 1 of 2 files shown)


pawl:
I'm not having any issues with date range filters: http://i.imgur.com/l129TLA.png

Here's the gist I'm using: https://gist.github.com/pawl/cd3b172ae832ffb0b791

Any idea how to reproduce the issue?

vToMy:
When I tested it worked fine for dates inside the range, but failed for boundary dates.
In your example - if you had a record with 2016-03-03 it would not be displayed (tested with sqlite).

pawl:
I see the issue now. It happens when the end boundary is the same as the record's date. I verified this pull request fixes the issue.

Screenshots: http://imgur.com/a/eW5DI

PR #1214, opened 2016-02-25
Added hybrid properties filters support

vToMy:
- added hybrid properties filters support.
- fixed hybrid properties support (hybrid properties with `select` expressions wouldn't work).

### DIFF
--- a/flask_admin/contrib/sqla/view.py
+++ b/flask_admin/contrib/sqla/view.py
@@ -601,7 +601,7 @@ class ModelView(BaseModelView):
 
                 column = columns[0]
 
-            # Join not needed for hybrid attributes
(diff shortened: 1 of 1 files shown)


pawl:
Made this change in #1209, thanks for the assistance.

vToMy:
This PR includes hybrid properties filters support, the fix was just a side thing because it wouldn't otherwise work. Therefore I would like to request to reopen it...

pawl:
Just tested this with the hybrid property example here: https://github.com/flask-admin/flask-admin/blob/master/examples/sqla-hybrid_property/app.py

The example worked before/after the change. However, scaffolding with only the column's name as a string now works too. Nice improvement.

PR #1229, opened 2016-03-14
Append inline model only after populating it

bepetersn:
Is there a reason it wasn't done this way before? I'm curious.

I'm trying to listen to the SQLAlchemy 'append' event, however at the time the model is appended in Flask-Admin, it has none of its fields populated. To listen for the model being populated using SQLAlchemy events but WITHOUT this, I had to listen for multiple 'set' events, and check that it was done populating in each listener.

### DIFF
--- a/flask_admin/contrib/sqla/fields.py
+++ b/flask_admin/contrib/sqla/fields.py
@@ -287,9 +287,9 @@ class InlineModelFormList(InlineFieldList):
                     continue
             else:
                 model = self.model()
-                values.append(model)
 
             field.populate_obj(model, None)
+            values.append(model)
 
             self.inline_view.on_model_change(field, model)
 


mrjoes:
I never checked it, but would SQLAlchemy ignore `append` event if model was in the list already?

bepetersn:
Like if the model was a duplicate entry in the list? It does fire the append event.

mrjoes:
Just confirming - it won't duplicate entry?

bepetersn:
Well, it does fire the `append` event, and it actually appends the duplicate entry too. I added a unique constraint because I noticed that behavior. In my listener, I check to see if it's a duplicate before taking action.

mrjoes:
I'm closing it, because it solves special case and will create duplicate entries.

PR #1236, opened 2016-03-22
Add default type formatters for Enum

singingwolfboy:


pawl:
@singingwolfboy is there a way to make this work in python 2 also?

singingwolfboy:
@pawl: sure, just install the enum backport package! https://pypi.python.org/pypi/enum34

flask_admin/model/typefmt.py
```diff
@@ -44,13 +48,27 @@ def list_formatter(view, values):
...
 }
+if Enum:
+    BASE_FORMATTERS[Enum] = enum_formatter
 
 EXPORT_FORMATTERS = {
     type(None): empty_formatter,
     list: list_formatter,
 }
+if Enum:
+    BASE_FORMATTERS[Enum] = enum_formatter
```
pawl: Needs to be `EXPORT_FORMATTERS` instead of `BASE_FORMATTERS` here. Also, maybe put them under the same IF statement?

flask_admin/model/typefmt.py
```diff
@@ -44,13 +48,27 @@ def list_formatter(view, values):
...
+    """
+    return value.name
+
+
 BASE_FORMATTERS = {
     type(None): empty_formatter,
     bool: bool_formatter,
     list: list_formatter,
 }
+if Enum:
```
pawl: bool(Enum) always evaluates to false for me, so it needs to be `if Enum is not None:`

flask_admin/model/typefmt.py
singingwolfboy: Oops! Fixed.

flask_admin/model/typefmt.py
singingwolfboy: Good catch! I just assumed it would be truthy, but apparently not. Fixed.

singingwolfboy:
I've responded to review comments -- thanks, @pawl! Can you give it another look?

singingwolfboy:
Actually, doing a little bit more testing, I discovered another interesting, unexpected thing about Python enums:
```
% python3
Python 3.5.1 (default, Dec  7 2015, 21:59:10) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.1.76)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import enum
>>> class Test(enum.Enum):
...   a = 1
... 
>>> type(Test)
<class 'enum.EnumMeta'>
>>> isinstance(Test, enum.Enum)
False
>>> type(enum.Enum)
<class 'enum.EnumMeta'>
```

Based on this, should I be setting `BASE_FORMATTERS[type(Enum)]` instead of `BASE_FORMATTERS[Enum]`?

pawl:
Seems to work fine as-is.

Before:
Before

After:
After

It's also worth noting that only SQLAlchemy >=1.1 returns python Enum types so far: http://docs.sqlalchemy.org/en/latest/core/type_basics.html#sqlalchemy.types.Enum

PR #1251, opened 2016-04-10
add example of disabled field to docs

pawl:


cancan101:
I think if you set this on a relationship, then the value get's cleared out on save.

PR #1254, opened 2016-04-18
new:mongoengine Field display verbose_name

likaiguo:
- support mongoengine column name display verbose_name in FooField


pawl:
Closing this for the same reason as #1374.

PR #1258, opened 2016-04-19
new: support flask_admin display django's model

likaiguo:
- filter query
- update model
- del model
- other: add pydev related project files !minor
- get navigation name from model._meta.verbose_name
todo
- create model


flask_admin/contrib/django/view.py
```diff
@@ -0,0 +1,323 @@
...
+        except Exception as ex:
+            print 'Failed to delete record.', ex
+
+            return False
+        else:
+            self.after_model_delete(model)
+
+        return True
+
+#     def scaffold_sortable_columns(self):
```
petrus-jvrensburg: please remove the commented code

PR #1265, opened 2016-04-23
SQLA - fix column names error for export and details view

pawl:
The SQLA backend has some special logic for determining column names for relations. But, that logic wasn't getting applied to the list of columns in the export and details view. It's what's causing tests to fail on @xqliu in: https://github.com/flask-admin/flask-admin/issues/1263

This pull request fixes the issue, and I was also able to refactor duplicate code in `get_list_columns`, `get_details_columns`, and `get_export_columns` into the new `get_column_names`.

This changes some behavior in two edge cases:
* If `column_list`, `column_export_list`, or `column_details_list` evaluate to false (instead of just `None`), flask-admin will get columns from the model (scaffold_list_columns). I'm not sure why anyone would want to use an empty list for these.
* Columns in `column_exclude_list`, `column_details_exclude_list`, or `column_export_exclude_list` will never be shown on their respective views. Before, these were only excluded if `column_list`, `column_export_list`, or `column_details_list` were `None`. I'm not sure why anyone would have the same column in both the exclude_list and include list.

### DIFF
--- a/flask_admin/contrib/sqla/view.py
+++ b/flask_admin/contrib/sqla/view.py
@@ -494,43 +494,42 @@ class ModelView(BaseModelView):
 
             return result
 
-    def get_list_columns(self):
+    def get_column_names(self, only_columns, excluded_columns):
         """
             Returns a list of tuples with the model field name and formatted
-            field name. If `column_list` was set, returns it. Otherwise calls
-            `scaffold_list_columns` to generate the list from the model.
+            field name.
+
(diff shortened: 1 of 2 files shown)


pawl:
@mrjoes have time to review this one?

Users will see errors like `Neither 'InstrumentedAttribute' object nor 'Comparator' object associated with Model1.test1 has an attribute 'replace'` when they try to use export or details views with SQLA relations until this is fixed.

PR #1278, opened 2016-05-06
new: support search placeholder

TTWShell:


uSpike:
This wipes out translations for the search placeholder field.  Could you use `flask_admin.babel.gettext` to get the "search" string?

TTWShell:
@uSpike

flask_admin/contrib/sqla/view.py
```diff
@@ -555,6 +555,33 @@ def init_search(self):
...
+
+        for searchable in self.column_searchable_list:
+            if isinstance(searchable, InstrumentedAttribute):
+                placeholders.append(
+                    self.column_labels.get(searchable.key, searchable.key))
+            else:
+                placeholders.append(
+                    self.column_labels.get(searchable, searchable))
+
+        return 'Search: %s' % u', '.join(placeholders)
```
illume: Does gettext need to be used here?

TTWShell:
@pawl How about this pr?

flask_admin/contrib/sqla/view.py
xqliu: @illume Do you have any insight here on what will happen if we use lazy_gettext in column_labels, and then use self.column_labels.get, will it get the already translated label?

gettext might not necessary here.

petrus-jvrensburg:
This looks good to me. Any objections to merging?

One change that I would make would be to call `_gettext` explicitly on the substring `Search` from the template, to ensure it gets translated appropriately. I don't see the use in calling `gettext` on the list of column names.

PR #1322, opened 2016-07-27
support seletced multiple rows to export

521xueweihan:
My English is very poor, so I make a gif picture to show this feature:

http://oayueu6v6.bkt.clouddn.com/flask-admin-select-export-iloveimg-compressed.gif


petrus-jvrensburg:
Thanks for this!

I tried it out locally, but I don't think it's a great idea in general: At the moment, the behaviour we're looking for is that the user filter / search until they have only the records that they are interested in, and then they can export the list. I'm not sure there's a strong use-case for selecting only specific rows for export.

PR #1323, opened 2016-07-28
Updated hybrid property detection to new sqlalchemy API

vToMy:


mrjoes:
Do you what's the minimum SQLAlchemy version that exposes this functionality?

vToMy:
Trying to figure out the answer to your question made me realize the new detection would miss some hybrid properties
I changed the way the detection works to be more robust (using sqlalchemy_utils), and it seems to work on all the cases which once failed.
Note the current version works, but it would not on newer sqlalchemy.

Maybe changing `flask-admin\examples\sqla-hybrid_property\app.py` to test this would be beneficial.
Currently it creates the filter itself, instead of automatic scaffolding. Changing:
`column_filters = ('number_of_pixels', )`
Would at least ensure the example will reflect the current state of the system. It's not a test but it's something.
I could do that if you're okay with that.

vToMy:
It seems the tests are failing because it's missing sqlalchemy-utils dependency.
Are you okay with adding it as a dependency?

vToMy:
And... `sqlalchemy-utils` doesn't support python 2.6.
Honestly I don't think there's any need to support 2.6.

mrjoes:
I'm not sure it is a good idea to include sqlalchemy-utils - Flask-Admin is not dependent on SQLAlchemy, it is just one of the DB backends.

Do you know if it would be hard to reimplement the functionality so it is doing similar thing without the dependency?

vToMy:
Apparently it's a one-liner, so here you go.

pawl:
This is going to break backward compatibility for people using sqlalchemy <0.8.  We're going to need to add some notes to the change log about that.

mrjoes:
Yeah, I know. But SQLAlchemy 0.8 is ~3 years old though. 

I'll add a check to SQLa backend and changelog note.

PR #1333, opened 2016-08-25
Update actions.js

aq1:
Create function to toggle row checkbox if all checkboxes were toggled manually


aq1:
Wow, looks like someone needs to fix the build system. Although project looks more dead than alive.

PR #1334, opened 2016-08-28
support for WTForms >= 3.0

alanhamlett:
WTForms 3.0 [removes](https://github.com/wtforms/wtforms/commit/60267825490e3e6653ee0b1ae2ef1fb56a1b340f) `TextField`.

### DIFF
--- a/flask_admin/contrib/peewee/filters.py
+++ b/flask_admin/contrib/peewee/filters.py
@@ -1,6 +1,3 @@
-import time
-import datetime
-
 from flask_admin.babel import lazy_gettext
 
 from flask_admin.model import filters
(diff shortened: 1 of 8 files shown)


flask_admin/contrib/peewee/filters.py
```diff
@@ -314,7 +311,7 @@ def convert(self, type_name, column, name):
 
         return None
 
-    @filters.convert('CharField', 'TextField')
+    @filters.convert('CharField', 'StringField')
```
mrjoes: This is Peewee column type, not a WTForms field.

alanhamlett: Thanks, just updated the PR to remove this change.

PR #1346, opened 2016-09-14
Fix CSRF for production deployments

ei-grad:
Current SecureForm implementation generates CSRF secret using
`os.urandom()` every time when application start up. CSRF secret is used
to calculate csrf_token check value, so if someone would use a command
similar to

    gunicorn --workers=8 app

to run his flask-admin app on production then most form submissions
would silently fail (silently - as for now, it is probably another one
bug).

Instead of custom `os.urandom()` logic the `app.secret_key` value should
be used to produce CSRF token values.


antoine-lizee:
@mrjoes Would love to have that into a release when you have time :-)

PR #1350, opened 2016-09-17
Extract a method for calculating inline model keys

### LINKED ISSUE #1347: can not render two inline models with same model
I have a self M2M relationship to show in create/edit form. The code is below.

```
class A(Model):
  id = Column(Integer, primary_key=True)

class B(db.Model):
  id = Column(Integer, primary_key=True)
  up_id = Column(Integer, db.ForeignKey(A.id))
  down_id = Column(Integer, db.ForeignKey(A.id))
  upstream = db.relationship(A, backref='downstreams', foreign_keys=[up_id])
  downstream = db.relationship(A, backref='upstreams', foreign_keys=[down_id])


class AModelView(BaseModelView):
  column_list = ('upstreams', 'downstreams')
  inline_models = [
    (A, dict(form_columns=('id','upstream')),
    (A, dict(form_columns=('id','downstream')),
  ]
```

`downstreams` can be rendered correctly but `upstream` can not be rendered correctly.

lifei:
To provide a way to custom the inline model keys
in the case of two same inline model class.

Fix #1347

### DIFF
--- a/flask_admin/contrib/sqla/form.py
+++ b/flask_admin/contrib/sqla/form.py
@@ -533,33 +533,22 @@ class InlineModelConverter(InlineModelConverterBase):
 
         return result
 
-    def contribute(self, model, form_class, inline_model):
+    def _calculate_mapping_key_pair(self, model, info):
         """
-            Generate form fields for inline forms and contribute them to
-            the `form_class`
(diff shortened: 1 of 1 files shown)


lifei:
Thanks!

PR #1361, opened 2016-09-29
Use Google Places for map box search

jmagnusson:
Use Google Places search box for geolocation.

Also:

* Upgrades `leaflet` to 1.0.0 final.
* Removes `leaflet.search` which was not working reliably (not even in their own demos...)
* Map uses full width of form container by default
* Ability to use a new app configuration key `GOOGLE_MAPS_API_KEY`. [This allows for increased daily quotas and ability to track usage](https://developers.google.com/maps/documentation/javascript/get-api-key). Note that this key must have Google Places API enable. See link.


jmagnusson:
Any chance we can have this and #1362 merged soon?

I'm guessing you're quite busy these days considering the silence here on github for the last few months. If you'd like I could offer some help in maintaining the project. Though it would be more on a need-this-code-basis.

PR #1367, opened 2016-10-04
Convert some tests to pytest and other 2to3 changes. Some doc title fixes.

romeojulietthotel:
I will let the code speak for itself. Hopefully it's not rude.


jmagnusson:
Moving to pytest would be sweet. I'm guessing there's some work left to do though when you write "some tests"?

pawl:
At first glance, this would break a lot of stuff. Especially the changes to _compat.py: https://github.com/flask-admin/flask-admin/pull/1367/files#diff-fa43e63c44272277a4753ee75b993a71

Giant pull requests (+1,930 −1,937!) like this are really difficult to review. It would be better to submit a PR for one type of change at a time.

PR #1348, opened 2016-09-15
Added support for remote hybrid properties filters

vToMy:
Fixes #1345 

Changes:
1) Updated hybrid property detection: added support for period-delimited column name - what it does is figure out if the last attribute on the last model is a hybrid property.
2) Added the `key` attribute to hybrid properties (this field is not automatically populated by sqlalchemy for hybrid properties, for some reason), which is used by flask-admin to determine the column name.
3) Prettifying the filter name for nested properties.
4) Removed unused method `get_association_proxy_column_name` I added in a previous PR (sorry!)

PR #1379, opened 2016-10-23
Fix regression with late-evaluation relationships no longer working with filters

jmagnusson:
Regression was introduced in 47080ae (PR #1348) where `is_hybrid_property` was changed to support remote hybrid properties. When lambdas were encountered in the iteration they were passed as-is to the function `get_hybrid_properties`, which in turn threw the exception `sqlalchemy.exc.NoInspectionAvailable: No inspection system is available for object of type <class 'function'>`.

Test has been updated to ensure that flask-admin will continue to support lambdas as a relationship’s first argument.

### DIFF
--- a/flask_admin/contrib/sqla/tools.py
+++ b/flask_admin/contrib/sqla/tools.py
@@ -1,3 +1,5 @@
+import types
+
 from sqlalchemy import tuple_, or_, and_, inspect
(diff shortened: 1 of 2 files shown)


jmagnusson:
Thanks @mrjoes. Any plans on a new release anytime soon?

vToMy:
Thanks for that.

PR #1386, opened 2016-10-30
fix python3 compatibility in sqla examples without future

### LINKED ISSUE #1364: sqla/app2.py is broken with python3.5.2
This is under python3.5.2

1) python examples/sqla/app2.py
2) Go to localhost:5000
3) Create new car record
4) Save
5) Select Tyre tab
6) Create new tyre record
7) Note that the Car field is populated with <__main__.Car object at 0x7fe2bef98438>

The same 7 steps work fine on python2.7.12.

Using my rudimentary debugging skills I found that the value is stored in the _view.

pawl:
Fixes #1364

I tried running the examples for the first time in a while and got `ImportError: No module named 'future'`. After messing with it, I don't think this module is necessary to run the examples.

It's better to just use `__str__` in the models to be compatible with both python 2 and 3. I thought maybe `python_2_unicode_compatible` was being used to make `__unicode__` work in python 3, but it doesn't.

### DIFF
--- a/examples/sqla/app.py
+++ b/examples/sqla/app.py
@@ -2,7 +2,6 @@ import os
 import os.path as op
 from flask import Flask
 from flask_sqlalchemy import SQLAlchemy
-from future.utils import python_2_unicode_compatible
 
 from wtforms import validators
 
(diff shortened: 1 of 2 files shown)


vToMy:
I added these when I added support to python 3 for this example.
I thought `future` was a builtin module in python 2?
Didn't want to break anything, since `__str__` in python 2 should return bytes and not a string.
If it works without it I guess that's better.

PR #1393, opened 2016-11-13
Replace list_columns with columns_list in examples/sqla-hybrid_property

danidee10:
This pull request removes the deprecated `list_columns` attribute for `ModelViews` in the `sqla-hybrid_property` example  and replaces it with `column_list`


danidee10:
This will probably be in another pull request, but i noticed that this example doesn't work with the latest version of SQLAlchemy (at the time of writing this is `SQLAlchemy==1.1.3`) are there any plans to make this example compatible with that version, or we should just stick with the one in the `requirements.txt` which is `SQLAlchemy==1.0.13`

PR #1400, opened 2016-12-05
Added possibility to extend css/js assets from admin view

vgavro:
It's easy to extend extra js and css anyway, but I'm sure creating custom fields that required extra css/js is pretty common procedure, so it would be better to have it in flask-admin already.
Related to https://github.com/flask-admin/flask-admin/issues/1328

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -148,22 +148,11 @@ classes as follows::
         widget = CKTextAreaWidget()
 
     class MessageAdmin(ModelView):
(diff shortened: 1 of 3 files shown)


vgavro:
Other (and may be better?) way is django way to have assets defined in widgets, and collect them on form rendering. Then it would be easier to distribute custom fields/widgets for flask-admin. Maybe we can have both methods together.
RFC?

petrus-jvrensburg:
Thanks @vgavro, this is a nice and simple solution.

PR #1406, opened 2016-12-16
FIX: don't pre-populate forms for one-to-one that are not required in SQLA.

antoine-lizee:
There is a bug in the SQLA form rendering for relations. If the *remote* model of a one-to-many or one-to-one relation has the corresponding ForeignKey to `nullable=False`, the corresponding widget gets erroneously rendered with `allow_blank=True`. This happens because of https://github.com/flask-admin/flask-admin/blob/master/flask_admin/contrib/sqla/form.py#L115 happening after https://github.com/flask-admin/flask-admin/blob/master/flask_admin/contrib/sqla/form.py#L110. The property itself is not required, rather the foreign key of the related object is. 

This is a problem mainly for one-to-one relations because it *pre-populates* the corresponding field with an existing object that is not required - while it still doesn't pre-populate for one-to-many, it just doesn't propose you a blank choice.

I just used the existing logic from the validator to determine the allow_blank. In particular, any non-MANYTOONE relation will be allowed to have blank fields, which only make sense :-).

### DIFF
--- a/flask_admin/contrib/sqla/form.py
+++ b/flask_admin/contrib/sqla/form.py
@@ -111,9 +111,8 @@ class AdminModelConverter(ModelConverterBase):
             else:
                 kwargs['validators'].append(validators.InputRequired())
 
-        # Contribute model-related parameters
-        if 'allow_blank' not in kwargs:
-            kwargs['allow_blank'] = column.nullable
+        # Never prepopulate relation widgets:
+        kwargs['allow_blank'] = True
 
         # Override field type if necessary
         override = self._get_field_override(prop.key)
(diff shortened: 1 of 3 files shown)


antoine-lizee:
PS: I have no idea why Travis is failing on that one, espe since it's only on 2.6 and mongoengine related stuff. Any idea ? Would be happy to fix if it's related to my PR.

vgavro:
@antoine-lizee this is related that mongoengine dropped support for python2.6 since 0.11, so from now all travis tests would break https://github.com/MongoEngine/mongoengine/releases/tag/v0.11.0

vgavro:
So options would be:
* fix mongoengine==0.10.9 for tests
* skip python2.6 tests for mongoengine
* or drop python2.6 support since flask-admin 1.5..

personally I prefer second variant, but I guess this should be decision from maintainer

antoine-lizee:
@vgavro thanks for the heads-up. I'd go for (3), but this is indeed a big step, and potentially not to be discussed here. In the meantime I've gone for (2).

PR #1408, opened 2016-12-25
Upgrade leaflet

### LINKED ISSUE #1093: Leaflet JS Errors on examples.flask-admin.org
To reproduce, go here: http://examples.flask-admin.org/geo/admin/multipoint/edit/?url=%2Fgeo%2Fadmin%2Fmultipoint%2F&id=1

and click the "edit layers" icon (above the delete trashcan).

```
Uncaught TypeError: Cannot set property 'previousOptions' of undefined
```

Is this an issue with Leaflet, or something wrong with the flask-admin implementation?

petrus-jvrensburg:
This PR upgrades both Leaflet and Leaflet.draw to the latest release versions.

It also removes the call to http://ip-lookup.com for centering maps when creating new shapes. This call has been causing problems because it's over 'http' and not 'https'. In stead, two new config variables have been added for setting a sensible default.

closes #1093


dkerkow:
Hi,
I just tried this branch as a submodule in an existing project, because I wanted to see if it fixes a problem I am having with an older forked version of flask-admin. Problem is: Models where geometry_type is not GEOMETRY, are not editable in the Leaflet widget. But my data models are MULTIPOLYGON and I want to enforce that. The problem seems to persist with this branch, and I dont know wether that is a problem with flask-admin or my codebase. Is this a known issue or should I file a bug?

mrjoes:
Are we good with this one? Need some feedback, but I'm not into geo stuff at all.

dkerkow:
Sorry, didn't want to capture the PR, what I described is most probably only a problem in my app.

alanhamlett:
@petrus-jvrensburg LGTM, is it ready to merge?

petrus-jvrensburg:
@dkerkow MultiPolygons can definitely be handled, have a look at 'examples/geo_alchemy' if you are unsure.

@alanhamlett ;) Yes, this one should be ready for merging

PR #1415, opened 2017-01-20
remove speaklater dependency

pawl:
This PR added speaklater as a dependency: https://github.com/flask-admin/flask-admin/pull/1348

The current logic:
1. Take the replacement column label from column_labels
1. Replace any `.` in the column label with ` / `
1. Turn the replacement column label back into a lazy string (using speaklater)

This PR removes speaklater as a dependency by changing the logic to just use the lazy_gettext object from `column_labels` if it's available (and not replacing `.` with ` / `). If the user manually specified a column label, they probably don't want the `.`'s in whatever they wrote replaced with ` / ` anyway.

### DIFF
--- a/flask_admin/contrib/sqla/view.py
+++ b/flask_admin/contrib/sqla/view.py
@@ -2,7 +2,6 @@ import logging
 import warnings
 import inspect
 
-from speaklater import is_lazy_string, make_lazy_string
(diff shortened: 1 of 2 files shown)


mrjoes:
Looks good. Also, looks like we're no longer compatible with Py2.6, dropping it?

PR #1417, opened 2017-01-27
Added JSON and JSONB and support for peewee postgresql backend

jammie987:
This pull request adds the ability to edit JSON columns to the peewee backend.

I used JSONField from flask_admin/form/fields.py and wrote two converters, for JSON and BinaryJSON fields.

### DIFF
--- a/flask_admin/contrib/peewee/form.py
+++ b/flask_admin/contrib/peewee/form.py
@@ -3,8 +3,6 @@ from wtforms import fields
 from peewee import (CharField, DateTimeField, DateField, TimeField,
                     PrimaryKeyField, ForeignKeyField, BaseModel)
 
-from playhouse.postgres_ext import ArrayField, JSONField, BinaryJSONField
-
 from wtfpeewee.orm import ModelConverter, model_form
 
 from flask_admin import form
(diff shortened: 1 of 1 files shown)


flask_admin/contrib/peewee/form.py
```diff
@@ -3,6 +3,8 @@
 from peewee import (CharField, DateTimeField, DateField, TimeField,
                     PrimaryKeyField, ForeignKeyField, BaseModel)
 
+from playhouse.postgres_ext import ArrayField, JSONField, BinaryJSONField
```
pawl: Looks like ArrayField is unused, you can probably get rid of this import?

pawl: Looks like this also requires psycopg2. It's probably a good idea to handle the ImportError gracefully like:
```
try:
    from playhouse.postgres_ext import JSONField, BinaryJSONField
    has_postgres_ext = True
except:
    has_postgres_ext = False
```

```
        if has_postgres_ext:
            self.converters[JSONField] = self.handle_json
            self.converters[BinaryJSONField] = self.handle_json
```

PR #1423, opened 2017-01-30
use tox, fix mongoengine & python 2.6 tests

pawl:
The MongoEngine tests were broken because MongoEngine now requires a recent version of Flask-WTF which requires WTForms 2 for its CSRF validation implementation. I fixed this by skipping WTForms 1 tests for the MongoEngine backend. 

The Python 2.6 tests were broken because the `keys` method calls the normal (unordered) `dict.keys()`. I fixed this by adding `ordereddict` as a dependency for 2.6 and removing the "Bare-bones OrderedDict implementation for Python2.6 compatibility" in case there are other problems with it. If we want to keep it, we could probably also fix this by adding a `keys` method that returns `self.ordered_keys`.

I added a tox.ini to make these sorts of issues easier to test locally. And, I made the travis.yml use tox so we can do stuff like running flake8 in the future. I also made it test Python 3.6.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ examples/appengine/lib
 env
 *.egg
 .eggs
+.tox/
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
   - "3.3"
   - "3.4"
   - "3.5"
+  - "3.6"
 
 env:
   - WTFORMS_VERSION=1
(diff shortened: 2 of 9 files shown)


pawl:
Once this is merged, I'm also planning to make another PR for adding flake8 tests to the tox.ini: https://github.com/pawl/flask-admin/compare/fix_mongoengine_tests...pawl:add_flake8?expand=1

PR #1435, opened 2017-02-07
Updates README.rst with svg badges

sobolevn:


pawl:
Looks much less pixelated, thank you!

PR #1438, opened 2017-02-09
Fixing bug with using multiple URL attributes

GrayAn:
Fixed small bug with using multiple URL attributes:

- filters and page size were cleared if search was used;
- sorting, search and page size were cleared if filters were used.

Was fixed by adding hidden inputs for storing arguments.


mrjoes:
It's a little opinionated, but I like the change. @pawl any objections?

PR #1452, opened 2017-03-24
feat: ability to remove layer with popup

ptbrowne:
When you edit multi polygons, there was no way to remove a layer (or I did not see it ?).

Simple, easy :


ptbrowne:
My bad, did not see the bin ><

PR #1470, opened 2017-04-13
Update select2 to v4.0.3

xim:
I have some additions I want to make to my flask-admin, and they are much harder to do in select2 3.x than in 4.x.

Tested the changes locally with a couple of different enum / foreign key fields with no obvious errors. If you wish me to test the changes more, please advise on a sensible set of tests.


mrjoes:
This looks good. Are there any backward-incompatible changes with 3.x branch?

xim:
Good question. Reading the [announcement][0] and grepping for relevant function names, I think the Ajax stuff needs testing. Hmmm... I'll dig into that!

[0]: https://select2.github.io/announcements-4.0.html

pawl-rs:
I noticed a few bootstrap styling issues last time I tried updating select2: https://github.com/flask-admin/flask-admin/pull/921#issuecomment-137903042

Know if those were fixed?

xim:
No, @pawl-rs, apparently some sizing issues remain. X button on a multiple select works, but that may not be the same as this?
Also, sorry @mrjoes, the Ajax field breaks as well. Dang. Back to the drawing board. Don't know if I have the combination of time and expertise to fix this.

PR #1476, opened 2017-04-26
Validate integer field range

guillaumep:
Inputting an out-of-range number in an IntegerField results in a database error which causes an error page to be displayed. This pull request propose an active in-form validation to avoid this issue.

Limits according to: https://docs.microsoft.com/en-us/sql/t-sql/data-types/int-bigint-smallint-and-tinyint-transact-sql

### DIFF
--- a/flask_admin/contrib/sqla/form.py
+++ b/flask_admin/contrib/sqla/form.py
@@ -1,7 +1,7 @@
 import warnings
 
 from wtforms import fields, validators
-from sqlalchemy import Boolean, Column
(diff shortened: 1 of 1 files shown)


guillaumep:
Inputting an out-of-range number in an IntegerField results in a database error. This pull request propose an active in-form validation to avoid this issue.

guillaumep:
I just realized that this applies only to MSSQL, and not to any other DB supported by Flask-Admin. My PR thus is not valid. Closing.

PR #1485, opened 2017-05-30
Use table name rather than model name to construct filter query in QueryAjaxModelLoader

njatkinson:
* https://github.com/flask-admin/flask-admin/issues/1484


SBillion:
This pull request should be merged and released because filters on QueryAjaxModel are useless without this fix

SBillion:
@mrjoes Thank you :+1:

PR #1496, opened 2017-06-14
Fix sorting in subdirectories in fileadmin

hadleyrich:
Sorting was broken for subdirectories due to the path not being passed to the sort_url method.

Also, there was an unused default sort setting, I've made use of that.


hadleyrich:
This pull request that fixes sorting in subdirectories also covers pull request #1562 which just uses the default sort settings.

Issue #1456, opened 2017-03-27
JavaScript - deprecated usage of ".size()"


aosdf6dd:
I ran into the problem that the JavaScript in /static/admin/js/actions.js uses ".size()" (in line4) in order to query the amount of checked checkboxes. Unfortunately this function is deprecated since JQuery 1.8 and not working with Jquery3 anymore (https://api.jquery.com/size/). Hence, the "delete" button in a list view was not working.

Changing ".size()" to ".length" fixed the issue for me, so I'm suggesting to patch it this way. Im not sure where else the deprecated ".size()" is used, though.

When I switch to JQuery v2, the "delete" button is working again, but then the awesome multi-select boxes (those offering search functionality and giving better overview) are not working anymore.

with best regards

PR #1497, opened 2017-06-17
Fixed issue where 'size()' is still used.

AdamGS:
fixes issue #1456 .


AdamGS:
It seems the travisci build is failing to install some dependencies.

PR #1500, opened 2017-06-22
Added some functionality, and 'optimized' a TODO comment.

AdamGS:


flask_admin/contrib/pymongo/view.py
```diff
@@ -385,12 +388,11 @@ def is_action_allowed(self, name):
             lazy_gettext('Are you sure you want to delete selected records?'))
     def action_delete(self, ids):
         try:
-            count = 0
+            deleted_models = self.find_in(ids)
```
pawl: Wouldn't this make it slower? Now it's querying for the records then deleting them. Before, it was just deleting them.

I think the optimization the comment had in mind was some sort of bulk delete? That would probably involve making `delete_model` accept multiple models.

PR #1502, opened 2017-06-27
Add support for columns-mapped column_property

iavael:
Add support for column_property in virtual models like here
https://docs.sqlalchemy.org/en/latest/orm/nonstandard_mappings.html#mapping-a-class-against-multiple-tables


petrus-jvrensburg:
thanks @iavael

PR #1505, opened 2017-06-29
check for valid scheme in the redirect target url (prevent xss)

### LINKED ISSUE #1503: Reflected XSS
We're using this framework internally within a flask app and a colleague noted a reflected XSS issue:

Example:
https://10.0.0.1/admin/user/edit/?url=%20javascript%3aalert(document.domain)&id=8

User interaction is required, when clicking the "List" link the javascript is executed.

pawl:
Fixes #1503 

This PR adds logic to `is_safe_url` to detect whether the the redirect url has a valid scheme. This prevents urls starting with `javascript:` which can be used for XSS.

I took some inspiration from here: https://github.com/django/django/blob/550cb3a365dee4edfdd1563224d5304de2a57fda/django/utils/http.py#L370

### DIFF
--- a/flask_admin/helpers.py
+++ b/flask_admin/helpers.py
@@ -8,6 +8,9 @@ from flask_admin._compat import urljoin, urlparse, iteritems
 from ._compat import string_types
 
 
+VALID_SCHEMES = ['http', 'https']
+
+
 def set_current_view(view):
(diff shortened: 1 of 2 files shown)


lbhsot:
anyone review this pr?

PR #1507, opened 2017-07-03
search in joined model fields for peewee

nad2000:
if the searchable comlum name contains '.' , eg, **fk.col**, *init_search* will add joined model column to the search fields.


samuelhwilliams:
This seems like a useful feature, but I would want to see some suitable testing added around this use-case.

Please do feel free to open a PR with appropriate unit tests if this is still something you would find useful and are able to contribute to the project.

PR #1527, opened 2017-08-17
allow to use inherited form CharField or TextField fields for searching

nad2000:


flask_admin/contrib/peewee/view.py
```diff
@@ -213,10 +213,8 @@ def init_search(self):
                 if isinstance(p, string_types):
                     p = getattr(self.model, p)
 
-                field_type = type(p)
-
                 # Check type
-                if (field_type != CharField and field_type != TextField):
+                if not isinstance(p, (CharField, TextField, )):
```
pawl: Can you change this from `(CharField, TextField, )` to `(CharField, TextField)`?

nad2000: sure...

PR #1531, opened 2017-08-27
Keep name from sorted column tuple in sqla

isra17:
This else clause is not needed since the column_name has been set earlier. However leaving it there break the `tuple` behavior where one might want to define a specific column name. My actual use case is to sort on Postgresql JSON column by using something like:
```
    column_list = ['price']
    column_formatters = {
        'price': lambda v,c,m,a: m.data['price']
    }
    column_sortable_list = [('price', MyModel.data['price'])]
```

Right now the `price` column is not set since the sortable column name is `"('price', <sqlalchemy.sql.elements.BinaryExpression object at 0x7fac0ce2e320>)"`.

### DIFF
--- a/flask_admin/contrib/sqla/view.py
+++ b/flask_admin/contrib/sqla/view.py
@@ -499,8 +499,6 @@ class ModelView(BaseModelView):
                     # column is in same table, use only model attribute name
                     if getattr(column, 'key', None) is not None:
                         column_name = column.key
-                    else:
-                        column_name = text_type(c)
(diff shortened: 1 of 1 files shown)


nad2000:
Hi @isra17, 

the build is broken. Not a biggie. need to add workaround to install correct version of Postgist, flake8 got more stringent. There are too new requirements added: E722 and E741. You might need to add following changes:

That would fix the build.
Cheer!

PR #1538, opened 2017-09-02
Add support for postgres CITEXT type

### LINKED ISSUE #1196: PostgreSql CITEXT column not shown/editable.
I have a Postgres table with a column of type CITEXT (it's a extension type, but a core extension type, Support is via this plug-in [here](https://github.com/mahmoudimus/sqlalchemy-citext)).

This column is not reflected in the create/edit dialogs of flask-admin.

It does show up in the table view.

It's probably fairly simple to do, but how can I convince flask-admin to allow editing of this column? For all functional purposes, it's basically just a column of type TEXT.

beenje:
citext is a case-insensitive character string type.
See https://www.postgresql.org/docs/current/static/citext.html

Fix #1196

### DIFF
--- a/flask_admin/contrib/sqla/form.py
+++ b/flask_admin/contrib/sqla/form.py
@@ -264,7 +264,7 @@ class AdminModelConverter(ModelConverterBase):
 
     @classmethod
     def _string_common(cls, column, field_args, **extra):
-        if isinstance(column.type.length, int) and column.type.length:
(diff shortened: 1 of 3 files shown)


brasilikum:
I could really use this. Any updates?

brasilikum:
Wow, that was fast! Thanks @mrjoes !

PR #1546, opened 2017-10-13
allow to filter by primary key in scaffold_filters

xmonader:
I'm working on an application where my models used to have `id` and `uid` unique 5 letters column, and I was able to filter using `uid`. Then the specs changed to make uid is the ID of the table so the id would be a unique 5 letters, but I wasn't able to filter because filters are skipped if the column is a primary key, and I had to override the scaffold_filters function. I think it's reasonable to allow filtering by IDs in general, unless that's by design.

Thanks :)

### DIFF
--- a/flask_admin/contrib/sqla/view.py
+++ b/flask_admin/contrib/sqla/view.py
@@ -591,7 +591,7 @@ class ModelView(BaseModelView):
                     # TODO: Check for multiple columns
                     column = p.columns[0]
 
-                    if column.foreign_keys or column.primary_key:
+                    if column.foreign_keys:
                         continue
 
                     visible_name = '%s / %s' % (self.get_column_name(attr.prop.target.name),


nad2000:
Hi @xmonader   , 

the build is broken. You might need to add following changes:

That would fix the build.
Cheer!

xmonader:
@nad2000  Thank you! I'll try to do that today :)

petrus-jvrensburg:
I'm closing this because it doesn't fix an issue. 

You can already successfully filter on the ID field of a model with `column_filters = ['id', ]`, or on a related model with something like `column_filters = [ 'user.id', ]`. So the only change this PR makes is to automatically add a filter on the id field for related models when only the model is specified, in stead of the specific fields, e.g. column_filters = [ 'user', ].

The workaround is to explicitly specify the fields on the related models that needs to be filtered, including the id.

PR #1551, opened 2017-10-22
fix x-editable booleans on list views

briancappello:
This PR fixes an issue where boolean fields editable on the list view do not render correctly on initial page load. (They show up as clickable links with el.innerText as "True" or "False", and once clicked, the popup form doesn't understand which value the row's item currently has.) This patch adds a `data-value` attribute so that the popup form knows which value the row's item has, and modifies the corresponding JavaScript so that it can render the field immediately as an icon.


nad2000:
Hi @briancappello  , 

the build is broken. You might need to add following changes:

That would fix the build.
Cheer!

PR #1562, opened 2017-11-19
fileadmin: respect 'default_sort_column' and 'default_desc' view attributes

SpotlightKid:
Attributes `default_sort_column` and `default_desc` aren't actually used at the moment to set the default sort order of the list in fileadmin's index view.

This PR fixes this and adds tests.

Test `test_fileadmin_sort_default_sort_column` should fail with out this change.


hadleyrich:
Subset of #1496

PR #1563, opened 2017-11-19
fileadmin: catch error bogus 'sort' url param is passed 

SpotlightKid:
… and preserve default sort order

When a value for the `sort` url param is passed to fileadmin's index view that doesn't correspond to a valid column name, `ValueError` is  raised.

This PR catches that and sets the  `sort_column` to the value of the  `default_sort_column` vie wattribute.

Tests added, which fail with an error without this change, due to the `ValueError` raised.


petrus-jvrensburg:
Thanks @SpotlightKid

Issue #1583, opened 2017-12-28
identity_key is returning an array with 3 arguments and breaking the code


dmelo:
https://github.com/flask-admin/flask-admin/blob/f04b71b78f75c2c5afdb50ceb2acf2ab27811d2b/flask_admin/contrib/sqla/fields.py#L299

I think it should not expect `identity_key` to return precisely two elements.

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1997, in __call__
    return self.wsgi_app(environ, start_response)
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1985, in wsgi_app
    response = self.handle_exception(e)
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1540, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python3.6/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python3.6/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/local/lib/python3.6/site-packages/flask_admin/base.py", line 69, in inner
    return self._run_view(f, *args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/flask_admin/base.py", line 368, in _run_view
    return fn(self, *args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/flask_admin/model/base.py", line 2076, in edit_view
    return_url=return_url)
  File "/usr/local/lib/python3.6/site-packages/flask_admin/base.py", line 308, in render
    return render_template(template, **kwargs)
[...]

PR #1584, opened 2017-12-29
remove identity_token from fields.py

dmelo:
On the method `get_pk_from_identity`, replace the `identity_token` method by `__mapper__.primary_key_from_instance`, that returns only the primary_key, in an attempt to solve issue #1583

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,6 @@ sudo: false
 language: python
 matrix:
   include:
-    - python: 2.6
-      env: TOX_ENV=py26-WTForms1
-    - python: 2.6
-      env: TOX_ENV=py26-WTForms2
     - python: 2.7
       env: TOX_ENV=py27-WTForms1
     - python: 2.7
--- a/flask_admin/contrib/sqla/fields.py
+++ b/flask_admin/contrib/sqla/fields.py
@@ -295,6 +295,5 @@ class InlineModelFormList(InlineFieldList):
 
 
 def get_pk_from_identity(obj):
-    # TODO: Remove me
-    cls, key = identity_key(instance=obj)
+    key = obj.__mapper__.primary_key_from_instance(instance=obj)
     return u':'.join(text_type(x) for x in key)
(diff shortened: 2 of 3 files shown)


andreoliwa:
Looking at the [failed builds](https://travis-ci.org/flask-admin/flask-admin/builds/322881886?utm_source=github_status&utm_medium=notification), it seems that SQLAlchemy 1.2 has dropped support for Python 2.6.

```
Collecting SQLAlchemy>=0.8.0 (from Flask-SQLAlchemy>=0.15->-r /home/travis/build/flask-admin/flask-admin/requirements-dev.txt (line 2))
  Downloading SQLAlchemy-1.2.0.tar.gz (5.5MB)
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-mXZzQi/SQLAlchemy/setup.py", line 16, in <module>
        raise Exception("SQLAlchemy requires Python 2.7 or higher.")
    Exception: SQLAlchemy requires Python 2.7 or higher.
```

Maybe if you change `tox.ini` and remove Python 2.6 as well, the build could pass.

```
envlist =
py{27,33,34,35,36}-WTForms{1,2}
```

But then Flask-Admin won't support Python 2.6 by extension. 🤔

jaysonsantos:
As SQLAlchemy is a fundamental part of flask-admin maybe it is better to drop 2.6 here too.

andreoliwa:
Broken build due to timeout: https://travis-ci.org/flask-admin/flask-admin/jobs/325321900
I think restarting your build should work. 

Also, there is another pull request trying to fix this same problem: #1586

dmelo:
#1586 should solve the problem as well. I will try and test it this weekend.

andreoliwa:
#1586 is already merged (even though it fails with Python 2.6).
This pull request might be obsolete now.

dmelo:
@andreoliwa I agree. Closing the PR.

PR #1586, opened 2018-01-05
Change fields.py to be compatible with SQLAlchemy>=1.2

asakaplan:
SQLAlchemy 1.2 added a third return value to identity_key, which broke this line. It was added as an optional return value.

### DIFF
--- a/flask_admin/contrib/sqla/fields.py
+++ b/flask_admin/contrib/sqla/fields.py
@@ -296,5 +296,5 @@ class InlineModelFormList(InlineFieldList):
 
 def get_pk_from_identity(obj):
     # TODO: Remove me
-    cls, key = identity_key(instance=obj)
+    cls, key, *token = identity_key(instance=obj)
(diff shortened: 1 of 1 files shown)


mrjoes:
Unfortunately, I can't accept this pull request as Flask-Admin should be also compatible with SQLAlchemy < 1.2.

asakaplan:
I put it as an optional return value, so it should be compatible with SQLAlchemy<1.2

mrjoes:
Yep, missed an asterisk. Isn't it Python3 only feature?

asakaplan:
Yup, give me a second on a Python2 compatible fix.

asakaplan:
It should be good now, albeit a bit uglier to read

Fokko:
@mrjoes Any idea when the release will happen? Not much going on at `flask-admin`

dlin-me:
Please release for this.

PR #1589, opened 2018-01-16
Move embedded style to admin CSS

nad2000:


nad2000:
**SQLAlchemy** has dropped the support for Python 2.6. What can be done about that? Should we drop support for 2.6 too or used SQLAlchemy 1.1.x with the build for 2.6?

hugovk:
In general, I would go for dropping Python 2.6. It's been [EOL for over four years](https://en.wikipedia.org/wiki/CPython#Version_history), and hasn't received any security updates since then. 

It's also generally little used. Here's the pip installs for flask-admin from PyPI for last month:

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |   54.9% |         28,273 |
| 3.6            |   25.3% |         13,043 |
| 3.5            |   13.3% |          6,869 |
| 3.4            |    6.2% |          3,185 |
| 3.7            |    0.1% |             38 |
| 2.6            |    0.1% |             27 |
| 3.3            |    0.0% |             21 |
| None           |    0.0% |              1 |

Source: `pypinfo --start-date -47 --end-date -17 --percent --pip --markdown flask-admin pyversion`

And it'll only become more painful to pin to old dependencies as time goes on. Dropping it will allow old code to be removed and more modern features of Python to be used.

(Python 3.3 is also EOL and similarly little used, that could be dropped too.)

petrus-jvrensburg:
Hi @nad2000, I want to merge the CSS move, that is described in the title for the pull request. But there are other things being changed at the same time in this PR. Can you please split them up, and for each thing you want to change, please give a short description / link to the issue that it closes?

PR #1592, opened 2018-01-19
Drop support for python 2.6

### LINKED ISSUE #1591: Python 2.6 build is failing
SQLAlchemy 1.2 now requires Python 2.7 or higher: http://docs.sqlalchemy.org/en/latest/intro.html#supported-platforms, so the build is now failing.

For example: https://travis-ci.org/flask-admin/flask-admin/jobs/330586924#L733

What should we do? Use SQLAlchemy 1.1.x or drop Python 2.6 support too?

jbochi:
Fixes #1591

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,6 @@ sudo: false
 language: python
 matrix:
   include:
-    - python: 2.6
-      env: TOX_ENV=py26-WTForms1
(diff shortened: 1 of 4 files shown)


jbochi:
@hugovk Thanks. I have also removed the version check and updated the classifiers.

PR #1597, opened 2018-02-03
BaseModel changed to ModelBase in newer versions of peewee

ronreiter:
Starting from peewee version 3, BaseModel is called ModelBase.


mrjoes:
Closed in the favor of the backwards compatible fix.

PR #1599, opened 2018-02-05
Don't use typeerror to guess function argument count

ninjawithaneedle:
`TypeError` is being used to switch from an `on_model_change` function with 3 arguments to one with 2 arguments. If there is code that throws `TypeError` in `on_model_change` with 3 arguments, a warning 

`"on_model_change() now accepts third parameter is_created. Please update your code" `

is shown which is confusing. For example, for the below code.

```
def on_model_change(form, model, is_created):
    1 + "2"
```

### DIFF
--- a/flask_admin/model/base.py
+++ b/flask_admin/model/base.py
@@ -3,6 +3,7 @@ import re
 import csv
 import mimetypes
 import time
+import six
 from math import ceil
 
 from werkzeug import secure_filename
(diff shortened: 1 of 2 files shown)


mrjoes:
While I agree that `TypeError` is bad solution, but I'd like to avoid using reflection to figure out number of arguments.

Alternatively, it's OK to drop this compatibility warning - it was introduced few major versions back.

PR #1612, opened 2018-03-05
Updating views.py to support changes in Peewee version 3.0 (for Field…

jasgrider:
…s the model_class attribute has changed to model). Previous versions of Peewee should work because of the try: except: loop I put in to first try model_class then on AttributeError try model.

### DIFF
--- a/flask_admin/contrib/peewee/view.py
+++ b/flask_admin/contrib/peewee/view.py
@@ -234,14 +234,24 @@ class ModelView(BaseModelView):
             raise Exception('Failed to find field for filter: %s' % name)
 
         # Check if field is in different model
-        if attr.model_class != self.model:
-            visible_name = '%s / %s' % (self.get_column_name(attr.model_class.__name__),
+        try:
+            if attr.model_class != self.model:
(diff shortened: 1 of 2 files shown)


flask_admin/contrib/peewee/view.py
```diff
@@ -234,14 +234,24 @@ def scaffold_filters(self, name):
...
+                    visible_name = self.get_column_name(name)
+        except AttributeError:
+            if attr.model !=  self.model:
+                visible_name = '%s / %s' % (self.get_column_name(attr.model.__name__),
+                                            self.get_column_name(attr.name))
+            else:
+                if not isinstance(name, string_types):
+                    visible_name = self.get_column_name(attr.name)
+                else:
+                    visible_name = self.get_column_name(name)
```
Obihoernchen: This quite some duplicate code. What's about something like this:
```python
model_class = None
try:
    model_class = attr.model_class
except AttributeError:
    model_class = attr.model

if model_class != self.model:
    # ....
```

Gromina:
Will this PR be merged in near future? 
Or perhaps somebody could post good workaround.

Issue #1617, opened 2018-03-12
Filter on recursion relationship


DaniilAnichin:
In sqlalchemy ModelView, when having a model with self-join as relation(let's call it child), flask admin cannot filter over both child and child parent fields, since they're mapped to the single key_name in scaffold_filters method. However, all the needed joins are okay. When creating a query, both parent and child field refers to parent alias in raw sql(checked many times). Solution - add some part to the filter name, if there are joins already. Check PR pls #1618

ljluestc:
```

from flask_admin.contrib.sqla import ModelView
from flask_admin.contrib.sqla.filters import FilterEqual

class CustomModelView(ModelView):
    def scaffold_filters(self, name):
        filters = super().scaffold_filters(name)
        
        # Check if there are joins present in the query
        if hasattr(self, 'session') and self.session:
            query = self.session.query(self.model)
            if len(query._join_entities) > 1:  # Customize this condition based on your joins
                filters = {}
                # Customize filters as needed for your specific joins
                for attr, column in self.model.__mapper__.c.items():
                    if hasattr(column, 'foreign_keys'):
                        # Example: Add filters based on foreign keys
                        for fk in column.foreign_keys:
                            # Assuming 'parent' is the foreign key relationship name
                            if fk.column.table == self.model.__table__:
                                filters[attr] = FilterEqual(column, name=f'{name}.{attr}')
                    else:
                        filters[attr] = FilterEqual(column, name=f'{name}.{attr}')
        
        return filters
```

PR #1618, opened 2018-03-12
Check to avoid join replacement in recursive-joined models, #1617 

DaniilAnichin:
Check issue #1617 or feel free to contact me for details.


petrus-jvrensburg:
I tried this locally, and added a test for it (310fe4abee9a65f325a989322ea0d2ee622806f4). But your fix isn't solving the issue. The filter on the related model column isn't working as expected.

PR #1620, opened 2018-03-21
Make the Select2Field coerce enum values to their name as required.

### LINKED ISSUE #1315: Flask-Admin doesn't populates correct value from Enum column
I have a working example here: https://github.com/citizen-stig/flask_admin_enum

I investigated it a little bit and maybe problem is in SQLalchemy and how it provides information about Enum column.

TimotheeJeannin:
When an [SQLAlchemy Enum](http://docs.sqlalchemy.org/en/latest/core/type_basics.html#sqlalchemy.types.Enum) column is used: 

``` python
class Seasons(Enum):
    winter = 1
    spring = 2
    summer = 3
    autumn = 4

class MyModel(db.Model):
    season = db.Column(db.Enum(Seasons), nullable=False)
```
When a model is edited, the value selected by **Select2** is **not** the real value of the model but the first enum value instead. As a result, every time an edit is made to a model that has one or more enum columns, the user will reset the enum value without realizing.

The first enum value is selected instead of the real one because `Select2Field` is coercing enum values to `_compat.text_type` (`str` for python 2 and `unicode` for python 3) instead of coercing the name of the enum value (coercing to `Seasons.summer` instead of `summer` for example).

Fixes #1315.

### DIFF
--- a/flask_admin/contrib/sqla/form.py
+++ b/flask_admin/contrib/sqla/form.py
@@ -1,4 +1,5 @@
 import warnings
+from enum import Enum
 
 from wtforms import fields, validators
 from sqlalchemy import Boolean, Column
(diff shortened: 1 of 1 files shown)


TimotheeJeannin:
@mrjoes @pawl Let me know if there is anything I can do to improve this pull request.

bao-work:
for anyone looking for temp workaround, you need form_args.
them loop all column, add form_args according to column type.

```python
form_args[name] = dict(coerce=lambda v: v.name if isinstance(v, Enum) else text_type(v))
```

PR #1669, opened 2018-07-04
Notifications.

PaulWasTaken:
There should be a way to change message which a user will see when a filter fails.


PaulWasTaken:
Could anybody help me and tell what is wrong with my pull request?

ufo911:
error #1 is 
`flask_admin/model/base.py:2197:23: E128 continuation line under-indented for visual indent`

PR #1681, opened 2018-07-27
fileadmin - fix timestamp bug windows

joost823:
When creating a folder using flask fileadmin, and clicking on it to see its content, an OSError was raised:

`__init__.py` in `flask_admin/contrib/fileadmin`

`items.sort(key=lambda x: (x[0], x[1], x[2], x[3], datetime.fromtimestamp(x[4])), reverse=True)
OSError: [Errno 22] Invalid argument`

After some debugging, the cause is: `datetime.fromtimestamp(x[4]))`. This is a known python bug: https://bugs.python.org/issue29097.

Using datetime.utcfromtimestamp fixes this issue.

### DIFF
--- a/flask_admin/contrib/fileadmin/__init__.py
+++ b/flask_admin/contrib/fileadmin/__init__.py
@@ -827,7 +827,7 @@ class BaseFileAdmin(BaseView, ActionsMixin):
             # Sort by type
             items.sort(key=itemgetter(2), reverse=True)
             # Sort by modified date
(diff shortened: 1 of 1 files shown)


connorhess:
I found the same issue in the same folder on line 767

i changed
`return datetime.(timestamp).strftime(self.date_format)`
to
`return datetime.fromtimestamp(timestamp).strftime(self.date_format)`

and it worked after that
[__init__.zip](https://github.com/flask-admin/flask-admin/files/5189381/__init__.zip)

PR #1687, opened 2018-08-07
remove useless brackets

dzaytsev91:
Hey,
It seems like [here](https://github.com/flask-admin/flask-admin/blob/master/flask_admin/contrib/sqla/ajax.py#L58) the brackets are pretty useless, tuple will be returned any way


klinkin:
Good catch =))

PR #1699, opened 2018-09-04
refs #1503 fix reflected xss

lbhsot:
refs #1503 "%20javascrip%0at:alert(/xss/)" will also cause xss attack.

thanks Omar Eissa from Deloitte Germany who found this issue


lbhsot:
@pawl @mrjoes

flask_admin/helpers.py
```diff
@@ -132,7 +132,8 @@ def prettify_class_name(name):
 
 def is_safe_url(target):
     # prevent urls starting with "javascript:"
-    target = target.strip()
+    _substitute_whitespace = compile(r'[\s\x00-\x08\x0B\x0C\x0E-\x19]+').sub
```
fantix: `_substitute_whitespace` is better to be a const.

mrjoes: Agree here - please change it to const and I'll merge ASAP.

fantix:
Please also add tests, see https://github.com/flask-admin/flask-admin/pull/1505/files#diff-fbb5b1350ecfe9b74207458739d4e17a.

xqliu:
Dear @fantix  @mrjoes,

Is it possible that this could be merged and a new version released for this xss fix?

Thanks :)

flask_admin/helpers.py
```diff
@@ -131,8 +133,18 @@ def prettify_class_name(name):
 
 
 def is_safe_url(target):
+    # prevent urls like "\\www.google.com"
+    # some browser will change \\ to // (eg: Chrome)
+    # refs https://stackoverflow.com/questions/10438008
+    target = target.replace('\\', '/')
```
petrus-jvrensburg: is there a reason why you're only matching agains double slashes, and not doing `target.replace('\', '/')`?

flask_admin/helpers.py
```diff
@@ -9,6 +9,8 @@
 
 
 VALID_SCHEMES = ['http', 'https']
+_substitute_whitespace = compile(r'[\s\x00-\x08\x0B\x0C\x0E-\x19]+').sub
+_fix_multiple_slashes = compile(r'(^([^/]+:)?//)/*').sub
```
petrus-jvrensburg: please add a comment to make it easier to follow the logic of the regex here

flask_admin/helpers.py
```diff
@@ -9,6 +9,8 @@
 
 
 VALID_SCHEMES = ['http', 'https']
+_substitute_whitespace = compile(r'[\s\x00-\x08\x0B\x0C\x0E-\x19]+').sub
```
petrus-jvrensburg: are there tests for all these specific ASCII characters that we are matching against? The only whitespace that I see in the tests are in `assert not helpers.is_safe_url(' javascript:alert(document.domain)')` and for that use-case a simple `replace(' ', '') would do the job
Please either expand the tests, or simplify the logic

petrus-jvrensburg: P.S. when checking for control characters, wouldn't it be simpler to do it in the same way as it's done in Django by checking `unicodedata.category` (see https://github.com/django/django/blob/550cb3a365dee4edfdd1563224d5304de2a57fda/django/utils/http.py#L364)?

flask_admin/helpers.py
jhatch28: This matches one slash. Your comment is a syntax error because the first slash escapes the second one. You can't put a `\` before a quote at the end of a string literal in python. Try this:
`print '\\``

petrus-jvrensburg: Aah, yes, of course. Thanks @jhatch28

m-butterfield:
@lbhsot Will this be updated and merged anytime soon?

slint:
FYI, when having the latest version of `flask-admin` installed and you run `pipenv check` (which uses [pyup](https://pyup.io/)), you get something like:

```console
$ pipenv check
Checking PEP 508 requirements…
Passed!
Checking installed package safety…
36437: flask-admin <=1.5.2 resolved (1.5.2 installed)!
helpers.py in Flask-Admin 1.5.2 has Reflected XSS via a crafted URL.
see https://github.com/flask-admin/flask-admin/pull/1699
```

In which case if this is part of your tests/CI, you'll have to do `pipenv check --ignore 36437`

Lewiscowles1986:
Can this be tagged for release please?

amahlaka:
Well, github is now giving a big warning banner due to this vulnerability,  it would be really good to get this fix released

olorton:
@Lewiscowles1986 @amahlaka The tagged release for this is v1.5.3 - see [pypi.org/project/Flask-Admin/](https://pypi.org/project/Flask-Admin/)

mrjoes:
I'm a little lost here. The fix went live Dec 19, 2018. I don't see the GitHub banner either.

amahlaka:
nvm, it turns out that the github banner things was panicking because of a old requirements file that was just in a deprecated folder

PR #1712, opened 2018-09-26
Error in Ukrainian translate 

RomanGatchenko:
error in Ukrainian translate : "not equal" = "не дорівнює"


petrus-jvrensburg:
Thanks @RomanGatchenko

RomanGatchenko:
@petrus-jvrensburg Please update admin.mo also...

PR #1736, opened 2018-10-22
Clean up examples

petrus-jvrensburg:
The `examples` directory has become cluttered, with 32 different example apps, and no obvious place for a new user to start exploring.

This PR reduces that number by merging many of the examples, reducing the clutter, and making it more likely for a new user to discover something useful that they didn't know about.


petrus-jvrensburg:
Merging this one myself... it doesn't include any code changes, only examples & docs.

PR #1749, opened 2018-11-06
Improve documentation of contrib.sqla.ModelView.get_query.

### LINKED ISSUE #1739: Documentation for contrib.sqla.ModelView.get_query should mention overriding get_one
The documentation for [contrib.sqla.ModelView.get_query](https://flask-admin.readthedocs.io/en/latest/api/mod_contrib_sqla/#flask_admin.contrib.sqla.ModelView.get_query) reminds the user to override `get_count_query` as well. It should also mention to override `get_one` as well, for example like this:

```python
def get_one(self, id):
    """Return an element with limited to the elements returned by get_query."""
    query = self.get_query()
    return query.filter(self.model.id == id).first()
```

If it is not done this way, some elements might be filtered in the list view, but can still be accessed in the edit view by simply changing the id in the url. This potentially allows users to see/edit elements they should not be allowed to see/edit.

lapseofreason:
Mention overriding get_one to prohit access to elements by simply
changing the URL.

Resolves #1739.

### DIFF
--- a/flask_admin/contrib/sqla/view.py
+++ b/flask_admin/contrib/sqla/view.py
@@ -833,6 +833,17 @@ class ModelView(BaseModelView):
                 class MyView(ModelView):
                     def get_query(self):
                         return super(MyView, self).get_query().filter(User.username == current_user.username)
+
+            Individual elements that are filtered in the list view are still
+            accessible through the edit view by simply changing the URL, even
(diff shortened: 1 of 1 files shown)


petrus-jvrensburg:
thanks @lapseofreason

PR #1750, opened 2018-11-06
Fix "undefined app" error in introduction.rst

ei-grad:


petrus-jvrensburg:
Aah... thanks for that @ei-grad !

PR #1756, opened 2018-11-14
CheckboxListField: alternative field for many-to-many relationship

GrayAn:
Greetings!

This is a new field which can be used instead of default `QuerySelectMultipleField` for many-to-many relationships. It looks like this:

checkboxlist

Also made `flask_admin.contrib.sqla.fields` module visible in the documentation API index.

### DIFF
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -15,6 +15,7 @@ API
    mod_actions
 
    mod_contrib_sqla
+   mod_contrib_sqla_fields
    mod_contrib_mongoengine
    mod_contrib_mongoengine_fields
    mod_contrib_peewee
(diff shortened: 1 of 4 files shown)


petrus-jvrensburg:
Thanks @GrayAn !

pankus:
Can you provide an example, please

GrayAn:
@pankus I haven't worked with flask-admin for a while but I think the example from the docstring should work:
```
from flask_admin.contrib.sqla.fields import CheckboxListField
from my_models import Language  # SQLAlchemy model

class MyView(ModelView):
    form_columns = (
        'languages',
    )
    form_args = {
        'languages': {
            'query_factory': Language.query,
        },
    }
    form_overrides = {
        'languages': CheckboxListField,
    }
```

pankus:
Thanks @GrayAn. 
I already looked at this doc, but I notice that it only works when I pass a function to the 'query_factory'. For example,
`form_args = {
        'languages': { 'query_factory': lambda: Language.query }
 }`
Now it works, yet I cannot figure out the reason.

PR #1759, opened 2018-11-23
Case-sensitive search for SQLAlchemy views

GrayAn:
Greetings!

Case-sensitive search for SQLAlchemy views with the same syntax as for MongoEngine views: using `*` character as a prefix switches to `LIKE` SQL operator instead of `ILIKE`.


petrus-jvrensburg:
Thanks for this @GrayAn. Let me add a test for the case sensitivity.

petrus-jvrensburg:
I added a test on [case-sensitive-search](/flask-admin/flask-admin/compare/case-sensitive-search), but it doesn't seem to be working as expected. I'm closing this one now, but please re-open if I'm missing something.

PR #1763, opened 2018-12-04
Letting the edit html inherit from right source in form-files-field example

joost823:
Just a small fix in one of the examples, even though it does not make in functionality in the example, it's a bit cleaner like this.


petrus-jvrensburg:
Thanks @joost823 !

PR #1764, opened 2018-12-04
[mongoengine] added ajax filters

mjurenka:
Hey,

I needed to filter ajax responses by additional values, so I have added this filter for MongoEngine models, similarly as defined in tutorial for QueryAjaxModelLoader.
The list of filters is list of Q objects, because we work with Q objects to query the db.

Thanks
Martin


ufo911:
@mrjoes can this be merged ?

samuelhwilliams:
Closing as we have removed flask-mongoengine support for now, due to it being unmaintained.

PR #1765, opened 2018-12-05
Do not install examples in site-packages

cgohlke:


petrus-jvrensburg:
Thanks @cgohlke, but this has already been fixed. If you install the latest version from pypi you shouldn't get any examples in site-packages. But please let me know if I'm missing something.

Issue #1532, opened 2017-08-29
Notification about deleting records is not translated into ru.


PaulWasTaken:
![2017-08-29 18-48-24](https://user-images.githubusercontent.com/25581039/29824405-ecc73308-8cea-11e7-96da-0d375afea65a.png)

I have this after deleting a record. It works fine for 2 or more, also other notifications are translated(as far as I noticed). I tried setting path to .mo file manually but it didn't make the trick(in func messages_path() ). Any advice?

I use this to define language.
```
@babel.localeselector
def get_locale():
    return request.accept_languages.best_match(LANGUAGES.keys())
```
Found later
![2017-08-30 17-03-36](https://user-images.githubusercontent.com/25581039/29871664-0e0cdeee-8da6-11e7-8138-76057bc683f2.png)
![2017-08-30 17-09-24](https://user-images.githubusercontent.com/25581039/29871690-2753dede-8da6-11e7-8f50-ba964dfa18ff.png)

PR #1775, opened 2018-12-12
Fixing Notification and button tooltip about deleting records is not …

rocarras:
…translated  (#1532)

To solve this issue, i took the liberty of add the text that was missing on 'es' translation:
```
msgid "Delete record"
msgstr "Borrar registro"
```

Then compile 'po' file:
`pybabel compile -i flask_admin/translations/es/LC_MESSAGES/admin.po -o flask_admin/translations/es/LC_MESSAGES/admin.mo `

Finally, the new translations tests are all green :)
`nosetests flask_admin.tests.test_translation`

Please consider my request, today we are hacking this issue.


petrus-jvrensburg:
Thanks @rocarras, but this one has been fixed here: #1669

If you can submit the missing Spanish translation on its own, that would be great. But I don't want to merge the additional test: it's merely testing that the dependency is doing what it's supposed to be doing, which is already part of the dependency's own test suite.

PR #1786, opened 2019-01-08
Use consistent case for func.ST_AsGeoJSON

TomGoBravo:
Change case of func.ST_AsGeoJson to match name in func, needed for fix in https://github.com/geoalchemy/geoalchemy2/issues/202


MeTaNoV:
It would be useful. @TomGoBravo , can you rebase so that @mrjoes spend less time to merge this commit?

TomGoBravo:
Rebased on master. Please merge @mrjoes

PR #1798, opened 2019-01-21
Admin menu ordering

mush42:
Hello,

Currently Flask-Admin doesn't provide a way to order the links in the admin menu.
However, such a feature is  extremely important for admin interfaces targeting non-developers as well as libraries built on top of Flask Admin (such as [Oy Content Management System](https://github.com/mush42/oy-cms).

This PR addresses this issue by adding two new attributes as follows:
- *sort_order* to `flask_admin.menu.BaseMenu` and all its subclasses  (default to None).
- *menu_order* to `flask_admin.base.BaseView` and
`flask_admin.model.base.BaseModelView` and all their subclasses (defaults to 0)

Based on these attributes the `BaseView.()_menu` attribute is sorted every time a new view is added using `BaseView.add_view`.

For menu links added using `BaseView.add_link`, the `BaseView._menu_link* is also sorted whenever a new link is added.

For menu categories, the user can specify category orders on the `BaseView.category_menu_orders` dictionary the same way they specify `category_icon_classes`. In this case the `category._children` list is also sorted whenever a new child is added to the category.

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -118,6 +118,19 @@ instance running on the same machine as your app, you can::
     admin.add_view(rediscli.RedisCli(Redis()))
 
 
+Controling The Sort Order of The Admin Menu Links
+--------------------------------
+
+****
+
+The `menu_order` attribute of any admin view allows you to specify an integer that would be used to sort links in the admin menu regardless of the order of view registration.
+The typical usage of this attribute is to pass it as a keyword argument when initializing your admin views.
+Suppose you want to register an admin view called `MyView`, and you want it to be listed first,
+then you should give it a sufficiently large `menu_order`::
+
+    admin.add_view(MyView("myview", endpoint=".myview", menu_order=500))
+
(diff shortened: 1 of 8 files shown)


jma:
It would be nice to have it!

david672orford:
You might want to clarify some of the terminology. First of all, maybe you should change the first sentence of this pull request to:

> Currently Flask-Admin doesn't provide a way to order the links in the admin menu independently of insertion order.

Presumably you have a use case in which menu entries are added and removed as the application runs.

Second, I think the parameter name "menu_order" is subject to misunderstanding. It sounds like something that applies to the menu as a whole. (Like "ORDER BY" in SQL.) Until I looked at the code I expected it to take values such as "alphabetical" or "ascending" and wondered why you would want to do that.

It would be clearer if we referred to the "position" of a menu item rather than to its "order". The "menu_order" parameter could be "menu_pos" instead.

PR #1813, opened 2019-02-18
Remove extra call to `self._get_path`

citizen-stig:
I've noticed, that `self._get_path` is already evaluated at the beginning of the method, so it's not necessary to call it again.


samuelhwilliams:
I'm going to close this for now, as it's a very old PR. While it doesn't seem incorrect or wrong, it also doesn't make any tangible or useful impact so I think it's not required at the moment.

PR #1820, opened 2019-02-28
Fix case of ST_AsGeoJSON to agree with spatialite

TomGoBravo:
With this change and https://github.com/geoalchemy/geoalchemy2/issues/202 (in the coming 0.6.0 version of geoalchemy) spatialite works.


TomGoBravo:
@mrjoes Please merge this tiny fix. Thank you.

petrus-jvrensburg:
Thanks @TomGoBravo 👍

Issue #1671, opened 2018-07-05
WTForms InputRequired validator doesn't work for mongoengine ListField


nathan-steyer:
It seems that the `InputRequired` validator does not work for a mongoengine `ListField` with `required=True`. Even when the list is non-empty, the form validation fails because the validator is checking the `raw_data` attribute of the WTForms `FieldList` which is never populated rather than the `entries` attribute.

I'm not sure whether to consider this a bug in Flask-Admin or WTForms. It could be fixed in either place. In my opinion, the `InputRequired` validator should be modified to handle `FieldList`s correctly, in which case the bug would resolve itself in Flask-Admin.

nathan-steyer:
I've opened [#432](https://github.com/wtforms/wtforms/issues/432) in the WTForms repo to _hopefully_ get this fixed

nathan-steyer:
The folks at WTForms don't want to modify the `InputRequired` validator to support `FieldList`s, so this will need to be fixed on Flask-Admin's side. I'll work on a pull request for this issue.

PR #1825, opened 2019-03-13
Fix bug in mongoengine model converter

nathan-steyer:
Modify mongoengine `CustomModelConverter` to use new `FieldListInputRequired` validator for required list fields

[#1671](https://github.com/flask-admin/flask-admin/issues/1671)


nathan-steyer:
I see the problem. I think the best solution would be to attach the validators for required fields in the methods that will create the field objects, e.g. `conv_List`, since those methods will know which class of field is going to be created and therefore which validator should be used.

PR #1674, opened 2018-07-14
Added bootstrap 4 theme

ryanbeymer:
#1486 - First pass at adding bootstrap 4 support

PR #1847, opened 2019-04-15
New theme - Bootstrap4

Fomys:
I created a new theme for flask-admin.
This theme is compatible with bootstrap4 and based on the work of pull request #1674.
I added a navigation menu on the left, a global search bar (But it's not functional, if someone has an idea to make it work that would be great) and a disconnect button.

I am open to any suggestions to improve this theme and make it available to all.


Fomys:
I add a favicon parameter because it seems that admin_static.url only looks in the static flask-admin folder and I want to have the possibility to replace the favicon only for the administration interface.

momentxu:
We did the same thing. 😢 
Maybe we can complete it together. 😸

Fomys:
It's a good idea to complete our themes because I use #1674 to make my bootstrap4-left, and the theme is not complete (modals...) and does not provide bootswatch support.

Do you have a good ressource to use bootswatch (I don't know how to integrate it)?

PS: You have the right to commit on my repo if you want to merge your bootstrap4 theme (I think you can totaly replace bootstrap4, my theme bootstrap4-left is independant)

PS2: I will add example for bootstrap4-left

momentxu:
> It's a good idea to complete our themes because I use #1674 to make my bootstrap4-left, and the theme is not complete (modals...) and does not provide bootswatch support.
> 
> Do you have a good ressource to use bootswatch (I don't know how to integrate it)?
> 
> PS: You have the right to commit on my repo if you want to merge your bootstrap4 theme (I think you can totaly replace bootstrap4, my theme bootstrap4-left is independant)
> 
> PS2: I will add example for bootstrap4-left

My bootstrap4 theme and the support of bootswatch is similar to the bootstrap3 theme and I almost finish them. 😸 
PS: Maybe we should update the branch of bootstrap4. 😺

petrus-jvrensburg:
Thanks for this @Fomys.

I'm not merging, because I don't want to include a whole new set of static files in the repo, just for moving the navigation controls from the top to the side. What would be great is if you could help get the regular bootstrap4 implementation (on the bootstrap4 branch) onto the same level as you've got here, and then perhaps update the 'custom-layout' example to show how the controls can be shifted, and a search control can be added to the navbar.

PR #1851, opened 2019-04-25
AdminModelConverter should not filter out primary keys

Governa:
If the primary key is also part of a foreign key, it would be filtered out and the form would fail to render.


hasansezertasan:
This PR does not reference an issue and doesn't test the changes. What about closing it @samuelhwilliams, @ElLorans?

Issue #1889, opened 2019-07-22
refrerence to non-existing _data property in the case of multicolumn PK


nad2000:
Peewee 3.0 doesn't exposed model _data property:

https://github.com/flask-admin/flask-admin/blob/3643d78808596f4634c8779e5decde9ef8ebda6e/flask_admin/contrib/peewee/view.py#L189

It was renamed to __data__.

ElLorans:
Fixed in https://github.com/pallets-eco/flask-admin/pull/1890

PR #1890, opened 2019-07-22
Prevent using deprecated _data property in Peewee contrib

nad2000:
Fix of the issue #1889


igoncharov85:
Can this be merged? I'm having the same issue.

PR #1893, opened 2019-07-26
Update app.py

aliounen12:


david672orford:
What is the purpose of this change? It looks to me like to turns a blank line into a blank comment.

hasansezertasan:
@samuelhwilliams, should we close this since we have Ruff which handles this kind of thing automatically?

samuelhwilliams:
Yeah, I agree 👍 

I'm not sure when we want to add ruff and re-style everything, but let's definitely do it all in one fell swoop

PR #1898, opened 2019-08-13
fixes issue with sample db not getting created on first use 

dreed47:
fixes issue with sample db not getting created on first use.  This moves logic for checking for the db file from run_server.ap to app/__init__.py

https://github.com/flask-admin/flask-admin/issues/1897


hasansezertasan:
I am currently unable to reproduce the problem you are trying to solve here.

I'm closing this PR for the moment, please open it again if the problem persists.

PR #1899, opened 2019-08-16
use circle-o for rendering unchecked booleans

alanhamlett:
I think `fa-circle-o` is visually better and clearer than `fa-minus-circle` for unselected booleans.

## Before
Screen Shot 2019-08-15 at 10 21 00 PM

## After
Screen Shot 2019-08-15 at 10 21 14 PM

Any thoughts or objections?

### DIFF
--- a/flask_admin/model/typefmt.py
+++ b/flask_admin/model/typefmt.py
@@ -36,7 +36,7 @@ def bool_formatter(view, value):
             Value to check
     """
     glyph = 'ok-circle' if value else 'minus-sign'
-    fa = 'check-circle' if value else 'minus-circle'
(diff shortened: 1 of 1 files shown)


david672orford:
I agree this is an improvement. I was always puzzled by the use of a "do not enter" sign for "false". By "puzzled" I mean that the first time I saw it I took it for an indication of an error or a lack of access.

While your proposal is an improvement, I don't think it goes far enough. The 'reverse video' check mark doesn't look very nice either. Why not go with the icons which are intended for this purpose: "checked-square" and "square".

While we are at it, let's change the Bootstrap glyphicons too so they will look similar. We could use "checked" and "unchecked".

PR #1911, opened 2019-09-17
Create new entry in model(pop-up) with one-to-many relationship, like in django-admin(works only with bootstrap3)

AlexKorob:
Screenshot from 2019-09-17 16-02-47


fredericwyss:
More than nice to have !

samuelhwilliams:
We have removed support for bootstrap3 now as it's such an old release. Would be open to you adding this for the bootstrap4 theme - if you are willing. For now I will close this PR, though.

PR #1918, opened 2019-10-20
Fix the usage of the deprecated textual coercion in SqlAlchemy 1.3+

ArtemGordinsky:
Since version 1.3, SqlAlchemy [removed coercion of string SQL fragments to text()](https://docs.sqlalchemy.org/en/13/changelog/migration_13.html#change-4481), which breaks filters in `form_ajax_refs`.

This fixes the issue by wrapping filters in `text()` explicitly.

Screenshot of an error caused by this bug from Sentry:
image


ArtemGordinsky:
@mrjoes Thanks for merging! Are you planning on releasing a new version any time soon? Looking forward to switching back from using our fork!

PR #1922, opened 2019-11-07
Import secure_filename from utils.

marksantcroos:
For more information see: https://github.com/pallets/werkzeug/pull/1640/commits/08536c457c7125c05e0947e62487fbc4bcf51717

### DIFF
--- a/flask_admin/form/upload.py
+++ b/flask_admin/form/upload.py
@@ -1,7 +1,7 @@
 import os
 import os.path as op
 
-from werkzeug import secure_filename
+from werkzeug.utils import secure_filename
 from werkzeug.datastructures import FileStorage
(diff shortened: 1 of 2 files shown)


mrjoes:
We'll have to wrap it with try/catch and import from the old location to keep the backwards compatibility.

marksantcroos:
They didn't move it, only removed the top level import.
I've checked the minimal version of Flask that Flask-Admin supports (0.7) and that requires at least werkzeug 0.6.1. In that version the new import already works.

ashb:
@mrjoes Can you please publish a new release to PyPi? Werkzeug 1.0 was just released which breaks the published version of Flask-Admin

PR #1928, opened 2019-11-21
Add documentation and example for using flask-security-too with Flask-Admin

jwag956:
Flask-Security-Too is a maintained and improved fork of the original Flask-Security.

The example is more full-functioned - providing all major views adapted to the current
Flask-Admin templates.

The example also makes use of the new more batteries included patterns introduced in recent
versions of Flask-Security-Too.

It also shows how to localize - utilizing a new jinja method to allow the new
templates to access existing flask-security-too translations.


ElLorans:
Nice work! Flask-Security and Flask-Security-Too are now merged. I am temporarily closing this pull request but I would be happy to review it if you can update the current example.

Issue #1951, opened 2020-02-06
Flask Admin is broken by Werkzeug 1.0.0


satshabad-cr:
Apologizes if this is wrong place to post this, but using pipenv and depending on the latest version of flask-admin, we get this.

```
Traceback (most recent call last):
  File "run_www.py", line 4, in <module>
    from www.entry import app
  File "/app/www/entry.py", line 110, in <module>
    import www.views.admin  # noqa
  File "/app/www/views/admin.py", line 5, in <module>
    from flask_admin.contrib.sqla import ModelView
  File "/usr/local/lib/python3.8/site-packages/flask_admin/contrib/sqla/__init__.py", line 2, in <module>
    from .view import ModelView
  File "/usr/local/lib/python3.8/site-packages/flask_admin/contrib/sqla/view.py", line 19, in <module>
    from flask_admin.model import BaseModelView
  File "/usr/local/lib/python3.8/site-packages/flask_admin/model/__init__.py", line 2, in <module>
    from .base import BaseModelView
  File "/usr/local/lib/python3.8/site-packages/flask_admin/model/base.py", line 8, in <module>
    from werkzeug import secure_filename
ImportError: cannot import name 'secure_filename' from 'werkzeug' (/usr/local/lib/python3.8/site-packages/werkzeug/__init__.py)
```

davidism:
Looks like this was fixed by #1922 but may not have been released yet.

laundry-96:
Is there a workaround until this is released, or will this be released soon? Could I possibly force the installation of werkzeug 0.16.1 from flask-admin?

rrcc0044:
I tried downgrading werkzeug==0.16.1 and it worked

levsa:
> Is there a workaround until this is released, or will this be released soon? Could I possibly force the installation of werkzeug 0.16.1 from flask-admin?

If you add `werkzeug==0.16.1` to your requirements file, flask-admin will use that, since it only requires `werkzeug>=0.7`

ddelange:
ref https://github.com/pallets/flask/pull/3489

mrjoes:
I'm doing a release today.

mrjoes:
We're back in action.

rrcc0044:
Thank you for your work!

ashb:
:heart: Thanks @mrjoes

PR #1952, opened 2020-02-07
Pin werkzeug to <1

NDevox:
The latest release of werkzeug (v1) has removed deprecated features including `werkzeug.secure_filename` which this library relies upon.

This commit pins the dependency to anything lower than v1 so the latest release will stop breaking until a patch can be submitted to support v1 and above.

This references #1951

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,6 +39,7 @@ extras_require = {
 
 install_requires = [
     'Flask>=0.7',
+    'werkzeug<1',
     'wtforms'
 ]
 


NDevox:
Looks like this has already had a patch put in and is now just awaiting release. (https://github.com/flask-admin/flask-admin/pull/1922)

setup.py
```diff
@@ -39,6 +39,7 @@ def grep(attrname):
 
 install_requires = [
     'Flask>=0.7',
+    'werkzeug<1',
```
ddelange: can I suggest to pin to `werkzeug>=0.15.0,<1` to leave less room for interpretation by pips greedy dependency resolving @NDevox 
```suggestion
    'werkzeug~=0.15',
```

ddelange: scratch that just saw your comment

NDevox: 😄 it was a good suggestion though.

PR #1960, opened 2020-03-13
if field is unique，“Already exists.” not translation

EVA09:
1、/flask_admin/translations/zh_Hans_CN/LC_MESSAGES/admin.po
`#: ../flask_admin/contrib/sqla/validators.py:42
msgid "Already exists."
msgstr "已经存在。"
`
2、/flask_admin/contrib/sqla/validators.py:42
` 
self.message = field.gettext(u'Already exists.')
`
maybe this is use `self.message = gettext(u'Already exists.')`
it `from flask_admin.babel import gettext`
I think this right

### DIFF
--- a/flask_admin/contrib/sqla/validators.py
+++ b/flask_admin/contrib/sqla/validators.py
@@ -1,5 +1,5 @@
 from sqlalchemy.orm.exc import NoResultFound
-
+from flask_admin.babel import gettext, ngettext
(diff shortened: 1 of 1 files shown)


samuelhwilliams:
Closing as I believe I've fixed this now separately in https://github.com/pallets-eco/flask-admin/pull/2464

PR #1963, opened 2020-03-15
fileadmin: wtforms Required validator deprecated

cjmayo:
InputRequired available since wtforms 1.0.2 in 2012 [1].

[1] https://wtforms.readthedocs.io/en/stable/changes.html#version-1-0-2

---

Silences some deprecation warnings. They do recommend DataRequired although the wtforms documentation pefers InputRequired.


gunlinux (approved):
Otherwise, we need to lock to specific WTForms<=2.3.3

PR #1968, opened 2020-03-18
Enable testing of MongoEngine backend on Python 3

cjmayo:
Lets's see how this goes. flask-mongoengine are now testing newer versions of Python with few obvious changes. Maybe the pinned PyMongo 2.8 will be an issue.

### DIFF
--- a/flask_admin/tests/mongoengine/__init__.py
+++ b/flask_admin/tests/mongoengine/__init__.py
@@ -1,11 +1,6 @@
 from nose.plugins.skip import SkipTest
 from wtforms import __version__ as wtforms_version
 
-# Skip test on PY3
-from flask_admin._compat import PY2
-if not PY2:
-    raise SkipTest('MongoEngine is not Python 3 compatible')
-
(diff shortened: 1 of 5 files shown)


cjmayo:
Finally a green tick, but only by skipping testing the pymongo backend.

The pymongo backend, test at least, is stuck on PyMongo 2 but the mongoengine backend test on recent Pythons needs at least MongoEngine 0.16 and I could not find a compatible PyMongo 2.

The first Pymongo 3 was released in 2015 and the last PyMongo 2 in 2017.

cjmayo:
Turned out it was just the test that was incompatible with PyMongo 3. I've replaced the skipping of pymongo backend testing with an updated test (that won't work with PyMongo 2). There is one deprecation warning fix in there too and a new skip test class collection patch. There is another deprecation warning with Pymongo >= 3.7 that could be fixed.

The pymongo backend is unchanged, presumably it still works with PyMongo 2. There are deprecation warnings from Pymongo 3.0 and PyMongo 3.7.

PR #1971, opened 2020-03-27
fix cast failed VARCHAR2 in oracle

pedia:
A field in Oracle table declared as:
>  DESCRIPTION	VARCHAR2(2048 BYTE)

Access db code like:
```python
from sqlalchemy.dialects.oracle import VARCHAR2

description = Column(VARCHAR2(2048), nullable=True)

# add this in admin.py
column_searchable_list = ('description',)
```

exception caught in search:

> sqlalchemy.exc.DatabaseError: (cx_Oracle.DatabaseError) ORA-00906:  missing left parenthesis [SQL: SELECT count(:count_2) AS count_1 FROM t_event_log WHERE lower(CAST(t_event_log.description AS `VARCHAR2`)) LIKE lower(:param_1)] [parameters: {'count_2': '*', 'param_1': '%100003%'}] (Background on this error at: http://sqlalche.me/e/4xp6)

The SQL should be 
```sql
CAST(t_event_log.description AS VARCHAR2(2048))
```

'(2048)' is missing in SQL which cause error.

This PR fixed this.

### DIFF
--- a/flask_admin/contrib/sqla/ajax.py
+++ b/flask_admin/contrib/sqla/ajax.py
@@ -1,5 +1,4 @@
 from sqlalchemy import or_, and_, cast
-from sqlalchemy.types import String
 
 from flask_admin._compat import as_unicode, string_types
 from flask_admin.model.ajax import AjaxModelLoader, DEFAULT_PAGE_SIZE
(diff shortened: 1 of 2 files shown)


pedia:
@mrjoes Would you please review this PR?

mrjoes:
Unfortunately, I can't merge this as-is, as it will break compatibility with non-string types that require forced conversion into a string for LIKE to work. This is a main reason why the CAST is made.

I'm not sure what's the right solution here for Oracle if Oracle always wants a length of a string in a string type.

pedia:
@mrjoes 

Take a look at the closed [issue](https://github.com/sqlalchemy/sqlalchemy/issues/5213) from sqlalchemy?

I think
```python
cast(field, field.type)
```
is more reasonable than:
```python
cast(field, String)
```

This PR only fixed error for: 
```python
column_searchable_list = ('description',)
```
and there is no more harmful for 
```python
column_searchable_list = ('an int field',)
```

mrjoes:
Got it. I will have to run the test, because I doubt that LIKE would work when field is casted to an INT explicitly.

Issue #1927, opened 2019-11-18
SQLAlchemy.__init__ stopped working for create_model in 1.5.4


killthekitten:
The custom `__init__` does not get called when the model is created through the `create_view`. Noticed after upgrading 1.5.3->1.5.4.

#1811 seems to introduce the bug, and the [follow-on patch](https://github.com/flask-admin/flask-admin/commit/731d14c16f57debd31f7f7f394d1585e20c28d5d 
) doesn't help.

Tested with SQLAlchemy 1.3.8 and 1.3.11 by throwing a breakpoint in the `__init__`.

killthekitten:
Easy way to reproduce: https://gist.github.com/killthekitten/e9c2549a22bd8a372c2164781a9d524d

etcsayeghr:
I can confirm that I'm also experiencing this bug. Will pin down to 1.5.3 in hopes that resolves our issue.

mrjoes:
I'm not sure that this is a bug, more here: https://docs.sqlalchemy.org/en/13/orm/constructors.html

Long story short, SQLAlchemy does not call `__init__` when it creates new instances of a model because `__init__` might have a custom signature. Flask-Admin 1.5.4 follows the same convention now - whenever the model is instantiated, it is following the SQLAlchemy convention of not calling `__init__` on the model. Try adding the `@orm.reconstructor` decorator and see if it will help.

killthekitten:
I can see why the change was introduced in the first place, but I'd still consider this a bug.

My use case for `__init__` is to set defaults for a new record, most of the time. Whereas `@orm.reconstructor` serves a different purpose: 
> The SQLAlchemy ORM does not call `__init__` when recreating objects from database rows. ... If you need to do some setup on database-loaded instances before they’re ready to use, ... it is also available via a class-specific decorator called `orm.reconstructor()`. 

Explicit `__init__` [is ok in the declarative](https://docs.sqlalchemy.org/en/13/orm/extensions/declarative/basic_use.html#class-constructor) SQLAlchemy, and there are [examples](https://docs.sqlalchemy.org/en/13/orm/tutorial.html#building-a-many-to-many-relationship) scattered all over the docs. So the update introduces quite a bit of headache for everyone who uses `__init__`.

One way to resolve this on my side would be to change all models to use the `default=` argument in the column declarations, but there's nothing fundamentally wrong about the current code, and I'm sure I will end up writing an explicit init again in the future. Because, heh, this is python.

[...]

PR #1981, opened 2020-05-04
Add a hook for custom SQLAlchemy model initializers (#1927)

killthekitten:
A cleaner way to override `create_model` when you need to customize model builder. A good example is having custom `__init__` on your model. The story behind this is the discussion in #1927.

- Does this need more documentation?
- This is a public method. Should it be mirrorered in other types of views (i.e. mongo, appengine etc)?

### DIFF
--- a/flask_admin/contrib/sqla/view.py
+++ b/flask_admin/contrib/sqla/view.py
@@ -1111,6 +1111,20 @@ class ModelView(BaseModelView):
 
         return super(ModelView, self).handle_view_exception(exc)
 
+    def build_new_instance(self):
+        """
+            Build new instance of a model. Useful to override the Flask-Admin behavior
+            when the model has a custom __init__ method.
+        """
+        model = self._manager.new_instance()
+
+        # TODO: We need a better way to create model instances and stay compatible with
+        # SQLAlchemy __init__() behavior
+        state = instance_state(model)
+        self._manager.dispatch.init(state, [], {})
+
+        return model
+
     # Model handlers
     def create_model(self, form):
         """
@@ -1120,11 +1134,7 @@ class ModelView(BaseModelView):
                 Form instance
         """
         try:
-            model = self._manager.new_instance()
-            # TODO: We need a better way to create model instances and stay compatible with
-            # SQLAlchemy __init__() behavior
-            state = instance_state(model)
-            self._manager.dispatch.init(state, [], {})
+            model = self.build_new_instance()
 
             form.populate_obj(model)
             self.session.add(model)


JDSQ:
The documentation on readthedocs doesn't seem to have been updated beyond 1.5.3. After upgrading to 1.5.7, my custom __init__ method still isn't being called. What do we need to do in order for our custom __init__ method to be called?

jekolbe:
In #1927 @killthekitten mentioned that it is possible to "override directly, without copying the entire `create_model` into the base class.
My custom init method isn't called. Maybe @killthekitten can help us out here, as it seemed to work for him :)

killthekitten:
Hi @JDSQ and @jekolbe. 

I'm sorry for not providing enough info to fix the original issue. This PR doesn't "fix" the issue of custom constructors, instead it exposes the hook that you can override and customise in your code. 

For it to work as expected, you need to override it in your ModelView classes, say like this:

```python
    def build_new_instance(self):
        return self.model()
```

Of course, the example above is oversimplified, so check out the discussion in #1927 to understand what's missing and perhaps just copy it over.

Let me know if you have any troubles with it!

jekolbe:
Thanks for this clarification! This was an misunderstanding on my side, sorry!

I guess I am still missing the last step here. Assuming I have a model called `Task` with attribute `id` and `title` in my application and I am creating the admin page for this model in my `app.py` like this:
```python
admin = Admin(app, name='some_app', template_mode='bootstrap3')
admin.add_view(ModelView(Task, db.session))
```

Then I overwrite the `build_new_instance` method like you described. At this point I simply don't know where do I get the required values for `id` and `title` from? To do something like...
```python
def build_new_instance(self):
    model = Task(id, title)
    return model
```
And even if I would have the key value pairs for all attributes how do I do it for multiple models "dynamically"?

killthekitten:
@jekolbe the answer to this question depends on how you work with models. Usually, the `id` would be generated either in the database, or in the model's `__init__` method, then you would simply omit it. The rest of the attributes would be [filled automatically by Flask-Admin](https://github.com/flask-admin/flask-admin/pull/1981/files#diff-b56f55fc39ad3735ddcb017cd2533e3d46d19b689f7fdc569f23d5911508e18bL1129) using `form.populate_obj(...)`.

Which means that you are good to go if your code is set up like this: 

```python
class Base(db.Model):
    id = Column(UUID, primary_key=True)
    created_at = Column(DateTime)

    def __init__(self, *args, **kwargs) -> None:
        self.id = str(uuid.uuid4())
        self.created_at = datetime.utcnow()
        super().__init__(*args, **kwargs)

class Task(Base):
    pass

class TaskModelView(ModelView):
    def build_new_instance(self):
        return Task()

admin = Admin(app, name='some_app', template_mode='bootstrap3')
admin.add_view(TaskModelView(Task, db.session))
```

However, it would be harder - but not impossible - to set it up if you need to pass the form's data directly to initialiser. Let me know if you would want to go that route.

PR #1993, opened 2020-06-17
Update deprecated syntax of pymongo

xxxwarrior:
There's no longer 'Connection()' attribute in "pymongo". 
Although there's "pymongo==2.4.1" in requirements, it cannot be installed due to the changes in later versions of "setuptools".

In my local repository, I used the connection string for my own MongoDB Atlas. I'm pretty sure, the URI for your db should be in some other place, not in this file.
Also, to use mongodb+srv:// URIs "dnspython" module must be installed, might change the requirement file for that.

### DIFF
--- a/examples/pymongo/app.py
+++ b/examples/pymongo/app.py
@@ -2,7 +2,7 @@ import pymongo
 from bson.objectid import ObjectId
 
 from flask import Flask
-import flask_admin as admin
+from flask_admin import Admin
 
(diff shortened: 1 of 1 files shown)


ElLorans:
Thanks for the nice work! Unfortunately this is being superseded by https://github.com/pallets-eco/flask-admin/pull/2611 but maybe you can help with your expertise there?

PR #1997, opened 2020-07-02
make MenuLink accept a callable as url, to allow dynamic urls

guillp:
e.g. the link url could be derived from some session or user attributes


guillp:
The same effect can be achieved by editing the master template, that's what I finally did. Closing.

PR #2003, opened 2020-07-16
Fix #1724 saving ARRAY(Integer) and other types than string

### LINKED ISSUE #1724: Integer array field in postgress is not handled by flask admin properly
Integer array field in my table in postgres exists and work perfectly. I can create/apdate/insert/delete it  with numbers.
Flask-admin sees and interprets that field by SQLAlchemy model. Field shows up with its content in edit form in flask-admin.
But whenever I try to save the form (regardless of editing or leaving contents of array field as is) I got error 500. Postgres throws an error:
`sqlalchemy.exc.ProgrammingError: (psycopg2.ProgrammingError) column "stepik_courses" is of type integer[] but expression is of type text[]`
So it seems that flask-admin interprets my int array field as text array.
Is there any solution on how to instruct Flask-Admin that the array field is of type int ?

catalintrif:


kbenchaaboun (approved):

ElLorans:
Superseded by https://github.com/pallets-eco/flask-admin/pull/2902

PR #2015, opened 2020-08-10
fix: Correct styling issues

michaelbukachi:
Before:
screenshot_20200810213813
After:
screenshot_20200810213727


flask_admin/static/vendor/multi-level-dropdowns-bootstrap/README.md
```diff
@@ -0,0 +1,61 @@
+# Bootstrap 4 Multiple Level Dropdown
```
TitaniumHocker, line 1: Do you think there is need to add this README file to static folder?

michaelbukachi: I went through some of the other vendor packages and there have README's so I just left it there.

TitaniumHocker: Ok

PR #2033, opened 2020-09-20
Update import location for SQLAlchemy project reorganization.

jtbaker:
For the pending 1.4 SQLAlchemy release, the `_class_resolver` class is moving to the `orm` module [here](https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/orm/clsregistry.py#L317-L354). This PR updates the module import path for one of the dependencies.

### DIFF
--- a/flask_admin/contrib/sqla/tools.py
+++ b/flask_admin/contrib/sqla/tools.py
@@ -1,7 +1,12 @@
 import types
 
 from sqlalchemy import tuple_, or_, and_, inspect
-from sqlalchemy.orm.clsregistry import _class_resolver
+try:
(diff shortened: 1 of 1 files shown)


mrjoes:
Can you add a try/except around the import and use the old import to keep backwards compatibility?

jtbaker:
> Can you add a try/except around the import and use the old import to keep backwards compatibility?

Will do, good idea.  👍

jtbaker:
@mrjoes I updated the PR with your suggestion and added a few comments about the handling. If you'd rather not have those in the source code, I'm happy to remove.

mrjoes:
This is good, thank you!

PR #2041, opened 2020-10-12
Use distinct on get_list queries

jojofeng:
I think how Flask Admin currently works is that the `get_list` queries don't use `distinct` on primary keys, so we include duplicate records in `query` and `count_query`. There is some code that (when rendering the list view) dedupes records on a page-by-page basis (but not across the entire search result), and we see weird behavior.

As an example, if page size is 5, and the query returns:

AAAAA,AAAAA,ABCDE, the count query returns 15 results (when in fact there are only 5 unique results), and we will see page 1 with just one record (A), page 2 with just (A), and page 3 with ABCDE. Naturally, we would expect there to really just be 1 page with all 5 unique records ABCDE.

By using `distinct` in our `get_list` `query` and `count_query`, we get the expected pagination behavior. I feel like by default, these `get_list` queries should return only unique records so the list renders correctly and consumers aren't responsible for deduping.

Closes https://github.com/flask-admin/flask-admin/issues/2040

### DIFF
--- a/flask_admin/contrib/sqla/view.py
+++ b/flask_admin/contrib/sqla/view.py
@@ -9,7 +9,7 @@ from sqlalchemy.sql.expression import desc
 from sqlalchemy import Boolean, Table, func, or_
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.sql.expression import cast
-from sqlalchemy import Unicode
+from sqlalchemy import Unicode, distinct
 
 from flask import current_app, flash
 
@@ -337,6 +337,11 @@ class ModelView(BaseModelView):
         if self._primary_key is None:
             raise Exception('Model %s does not have primary key.' % self.model.__name__)
 
+        if isinstance(self._primary_key, tuple):
+            self._primary_key_value = tuple(getattr(model, attr) for attr in self._primary_key)
+        else:
+            self._primary_key_value = getattr(model, self._primary_key)
+
         # Configuration
         if not self.column_select_related_list:
             self._auto_joins = self.scaffold_auto_joins()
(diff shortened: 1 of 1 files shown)


jojofeng:
@mrjoes what do you think?? would love to get this in if you think it's a reasonable change

jojofeng:
@mrjoes bump on this! would like to get your thoughts :)

jojofeng:
hey @mrjoes, any chance you could take a look at this sometime soon? would be great to get in so we don't have to manually override `get_list`! let me know what you think

mrjoes:
Sorry for late response, but I think I know what's the problem.

When there are joins to external tables (usually because of eager loading of record), joined table might have 5 related records. As a result, when Flask-Admin asks to get one page worth of data (20 record), 5 of them will be occupied by a single source record. SQLAlchemy deduplicates the records and only returns 15 - that's why it does not work as expected.

I saw this in a project where quick search was adding joins to related tables and it broke paging.

A quick fix was, basically, applying joins to subqueries instead of to the main query.

```python
    def _apply_search(self, query, count_query, joins, count_joins, search):
        terms = search.split(' ')

        filter_stmt = []

        for term in terms:
            if not term:
                continue

            stmt = tools.parse_like_term(term)

            for field, path in self._search_fields:
                q = self.session.query(self.model.id)

                q, _, alias = self._apply_path_joins(q, dict(joins), path, inner_join=True)
                column = field if alias is None else getattr(alias, field.key)

                q = q.filter(column.ilike(stmt))

                filter_stmt.append(self.model.id.in_(q.subquery()))

        query = query.filter(or_(*filter_stmt))

        if count_query is not None:
            count_query = count_query.filter(or_(*filter_stmt))

        return query, count_query, joins, count_joins
```

I'll be honest - I did not really test how distinct helps with the above problem.

jojofeng:
@mrjoes Thank you for the thorough response, and sorry for the late follow-up! I did end up integrating your suggestion above into my project and it worked wonderfully! Why isn't this the default `_apply_search` function in the `ModelView` class?

caffeinatedMike:
@jojofeng If you'd like to make the suggested changes previously mentioned and add test(s) showing it does not break backwards compatibility (or any other aspects of the codebase that rely on the same aforementioned function) I'd be happy to approve the PR :)

ElLorans:
Closing the PR since the real solution follows a different approach.

PR #2045, opened 2020-10-26
Use col-form-label instead of control-label in bootstrap4 templates.

TimotheeJeannin:
`.control-label` has been renamed to `.col-form-label.` in bootstrap 4.

See the migration guide here : https://getbootstrap.com/docs/4.5/migration/#forms-1


TimotheeJeannin:
I will make a PR with more fixes for the bootstap4 support.

PR #2046, opened 2020-10-26
Add the "row" class to form groups in bootstrap 4 templates.

TimotheeJeannin:
In bootstrap 4, `.form-group` no longer applies styles from the `.row` via mixin, so `.row` is now required for horizontal grid layouts (e.g., `<div class="form-group row">`).

See https://getbootstrap.com/docs/4.2/migration/#forms-1 for more details.


TimotheeJeannin:
I will make a PR with more fixes for the bootstap4 support.

PR #2067, opened 2020-12-18
Use the new python google datastore api

carlosgalvez-tiendeo:
Use the new python google datastore api
Add Flask-WTF as a dependency


pb-jeff-oneill:
Is this being considered?  I use Google App Engine with cloud ndb and would be great to have this.

Happy to volunteer to do what needs to be done.

samuelhwilliams:
Hi 👋  I've just removed the old appengine module, as it was outdated and causing maintenance issues.

I'd be happy to see this contributed back in if it's a more suitable variation. Would you like to update the PR and try again?

We'd also need to add tests for the core functionality you're bringing in.

PR #2096, opened 2021-03-17
Revert "Adding missing dependency on email_validator package"

alanhamlett:
Reverts flask-admin/flask-admin#2094

`email_validator` is only used in a docs example, not in this package.


alanhamlett:
Oops, reverted this revert.

PR #2116, opened 2021-06-03
Implement "frozen" fields

falko17:
This adds support for **frozen fields**: By setting a model view's `freeze_fields` member to `True` (default is `False`), clicking on "Save and add another" will cause the fields' contents to stay the same instead of being emptied.

This can be useful to quickly enter similar data successively.
For example, assume that there is a model for users where each user is represented by a name, and has certain roles and an expiry date. Now let's say the administrator needs to create entries for 10 new users with the same set of roles and the same expiry date. With this feature, instead of having to enter the roles and expiry date again for each new user (10 times in total), the administrator only has to do it once — the fields will stay "frozen", and after clicking on "Save and add another", only the name of each new user has to be modified.


falko17:
Alright, it took me some time to get around to it, but I've implemented the suggested change now. The only difference is that instead of a checkbox (which would only have affected a single button anyway) I've simply added another button if the user wishes to enable the "freezing" behavior.

falko17:
I've updated the pull request to work with the latest `master` now, but tests on Python 3.8, 3.9, and 3.12 still fail (other Python versions seem fine). To be honest, I'm not sure what causes this. I don't really see the relation between the error message and my changes yet. 

I also haven't really looked at this in 2 years, so I hope the PR itself is still fine—please tell me if I should change anything (apart from fixing the tests) 😅

ElLorans:
Sorry for jumping in after so long, but I would rather have a method to upload data on multiple rows or to create multiple objects rather than this. Even better, allowing to upload a csv.

falko17:
That's alright, but I don't have the time to make those larger-scale changes right now. A CSV upload feature seems like it should belong in a separate PR anyway. If I'm understanding correctly that you'd rather not merge the frozen fields feature, then feel free to close the PR.

ElLorans:
I am not particularly against it, I am only trying to imagine how to make it as user friendly as possible.
EDIT: and to be honest, I would like to know your opinion on it given your use case and your work on this PR. Would you prefer such a solution for flask-admin or not?

PR #2139, opened 2021-08-25
Fix JSONField when the value is empty json

ojayyezzir:
Previously it returned empty string instead of empty json

### DIFF
--- a/flask_admin/form/fields.py
+++ b/flask_admin/form/fields.py
@@ -187,11 +187,11 @@ class JSONField(fields.TextAreaField):
     def _value(self):
         if self.raw_data:
             return self.raw_data[0]
-        elif self.data is not None:
+        elif self.data:
             # prevent utf8 characters from being converted to ascii
             return as_unicode(json.dumps(self.data, ensure_ascii=False))
         else:
-            return ''
+            return '{}'
 
     def process_formdata(self, valuelist):
         if valuelist:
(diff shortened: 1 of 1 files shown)


caffeinatedMike:
What is the expected behavior here? From the looks of it, this won't behave as expected. If you pass an empty string to `json.dumps` it still ends up being an empty string, not a `dict` or `list`.

```python
>>> import json
>>> json.dumps("") == '""'
True
```

Can you add a test to show that this performs as expected?

ojayyezzir:
@caffeinatedMike
I would expect empty json to be stringified like below. 
```python
>>> import json
>>> json.dumps({})
'{}'
```
However because the condition(1) doesn't match when `self.data` is `{}`, it returns empty string by `else` statement(2)
```python
if self.raw_data:
  return self.raw_data[0]
elif self.data: # 1
  # prevent utf8 characters from being converted to ascii
   return as_unicode(json.dumps(self.data, ensure_ascii=False))
else: # 2
  return ''
```

caffeinatedMike:
Ahh, I see. However, the `elif` statement should remain what it was and the `else` statement is what should be updated to `return {}`

caffeinatedMike:
@ojayyezzir No problem :) And yes, you're right, good catch! We're returning stringified values, so `'{}'` is the correct return value.

caffeinatedMike:
@mrjoes PR approved and ready to merge

PR #2140, opened 2021-08-27
fix for flask-admin not using flask json_encoder

oleksandr-l5:


caffeinatedMike:
Can you explain the reasoning for this change? An empty PR description will in most cases lead to the PR being vetoed.

oleksandr-l5:
for example, I added a ObjectId or Decimal or LazyString type field to the flask json_encoder and I want it to be correctly serialized in the whole project

oleksandr-l5:
With this code I can use custom jsonencoder in whole app by setting jsonencoder in flask app. Do you need some example?

samuelhwilliams:
Closing this due to the age and the lack of clarity around purpose. If this is still an issue you are welcome to open another PR with more details, test cases, etc.

Issue #2144, opened 2021-09-08
Bootstrap 4 X-editable date causes error


mccarthysean:
I think I've found an error/bug in Flask-Admin's X-editable dates, but only in Bootstrap 4.

[Here's a StackOverflow question](https://stackoverflow.com/questions/69109930/flask-admin-bootstrap-4-x-editable-date-causes-error) that's identical to below. 

Notice in the below code, `template_mode='bootstrap3'` works fine, but `template_mode='bootstrap4'` causes the error.

Code to reproduce the error:
```python
import os
import datetime

from flask import Flask
from flask_sqlalchemy import SQLAlchemy
import flask_admin as admin
from flask_admin.contrib.sqla import ModelView


# Create application
app = Flask(__name__)

# Create dummy secrey key so we can use sessions
app.config['SECRET_KEY'] = '123456789'

# Create in-memory database
app.config['DATABASE_FILE'] = 'sample_db.sqlite'
app.config['SQLALCHEMY_DATABASE_URI'] = 'sqlite:///' + app.config['DATABASE_FILE']
app.config['SQLALCHEMY_ECHO'] = True
db = SQLAlchemy(app)

# Instantiate the admin class with Bootstrap mode
# admin = admin.Admin(app, 'Example', template_mode='bootstrap3')
admin = admin.Admin(app, 'Example', template_mode='bootstrap4')


# Models
class User(db.Model):
    id = db.Column(db.Integer, primary_key=True)
    name = db.Column(db.Unicode(64))
    email = db.Column(db.Unicode(64))
    date_added = db.Column(db.Date)

    def __unicode__(self):
        return self.name


class UserAdmin(ModelView):
    column_searchable_list = ('name', 'email')
    column_editable_list = ('name', 'email', 'date_added')
    column_display_pk = True

    can_set_page_size = True
    # the number of entries to display in the list view
    page_size = 5


# Add Flask-Admin views
admin.add_view(UserAdmin(User, db.session))

# Flask normal views
@app.route('/')
def index():
    return '<a href="/admin/">Click me to get to Admin!</a>'


def build_sample_db():
    """Populate a small db with some example entries"""

    db.drop_all()
    db.create_all()

    first_names = [
        'Harry', 'Amelia', 'Oliver', 'Jack', 'Isabella', 'Charlie','Sophie', 'Mia',
    ]
    last_names = [
[...]

PR #2146, opened 2021-09-09
fixes issue #2144 where X-editable combodate doesn't show up in Bootstrap 4

mccarthysean:
See [here](https://github.com/flask-admin/flask-admin/issues/2144) for background on the issue this fixes.


caffeinatedMike:
Barring styling woes (these can be addressed at a later point in time), I'm signing off on this PR @mrjoes

PR #2159, opened 2021-10-18
Fixed Upload fields compatibility with WTForms 3.x

### LINKED ISSUE #2142: Incompatibility problem with wtforms 3.0.0a1
FileUploadField seems to be incompatible with wtforms >= 3.0.0a1
The problem is:
`TypeError: process() got an unexpected keyword argument 'extra_filters'`
The process function overridden in the FileUploadFile class does not specify 'extra_filters' as a keyword argument.

wtforms==2.3.3 works, at least for this issue.

caffeinatedMike:
Closes #2142

### DIFF
--- a/flask_admin/form/__init__.py
+++ b/flask_admin/form/__init__.py
@@ -40,7 +40,7 @@ def recreate_field(unbound):
     return unbound.field_class(*unbound.args, **unbound.kwargs)
 
 
-if int(float(wtforms_version[0])) > 1:
+if int(wtforms_version[0]) > 1:
     # only WTForms 2+ has built-in CSRF functionality
     from os import urandom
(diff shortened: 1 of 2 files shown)


flask_admin/form/__init__.py
```diff
@@ -40,7 +40,7 @@ def recreate_field(unbound):
     return unbound.field_class(*unbound.args, **unbound.kwargs)
 
 
-if int(wtforms_version[0]) > 1:
+if int(float(wtforms_version[0])) > 1:
```
mrjoes, line 43: This is strange. What's the type of wtforms_version that it needs `float` now? Also, don't need `int` conversion - `float` would work as expected.

PR #2164, opened 2021-10-23
Add ability to use form_ajax_refs fields as editable columns

### LINKED ISSUE #1627: how to let AjaxSelectField editable?
I use relationship model ,so that can use 
```python
    form_ajax_refs = {
        'user': {
            'fields': (User.username,User.email)
        },
    }
```
to select user in **EDIT PAGE**  

and now , I want edit it in list page , I try to add `user` to ` column_editable_list`, but An Exception say:

```
Exception: Unsupported field type: <class 'flask_admin.model.fields.AjaxSelectField'>
```

Any idea? Thanks!

### LINKED ISSUE #2063: Editable field with "form_ajax_refs" causes Exception: Unsupported field type: <class 'flask_admin.model.fields.AjaxSelectField'>
Here's my [StackOverflow question](https://stackoverflow.com/questions/65193474/how-to-get-flask-admin-editable-field-dropdown-menu-to-lazy-load-with-ajax-inste) that made me realize this was an issue.

Apparently `form_ajax_refs` does not work for editable fields in the list view. I was having performance issues eager-loading relationship values for editable fields in the list view, so I tried `form_ajax_refs` to lazy-load it on-demand, but I kept getting the following error:
```
Exception: Unsupported field type: <class 'flask_admin.model.fields.AjaxSelectField'>
```

It happens when the `form_ajax_refs` field is also in the `column_editable_list`. 

**Here's my setup:**

In Flask-Admin, I have a view of my `Structure` model called `StructureView` which contains an *editable foreign key field* called `power_unit`. The `PowerUnit` model and database table contains many, many records, which are all apparently eager-loaded into the HTML, slowing down the loading time for the view.

I'd like the dropdown menu for the `power_unit` field to lazy-load when the user clicks on the field to select something from the dropdown list, and not on page-load.

Here are my models and my Flask-Admin view:
```
class Structure(db.Model):
    __tablename__ = 'structures'
    __table_args__ = {"schema": "public"}

    id = db.Column(INTEGER, primary_key=True)
    structure = db.Column(TEXT, nullable=False)

    power_unit_id = db.Column(INTEGER, db.ForeignKey('public.power_units.id'))
    power_unit = relationship('PowerUnit', back_populates='structures')


class PowerUnit(db.Model):
    __tablename__ = 'power_units'
    __table_args__ = {"schema": "public"}

    id = db.Column(INTEGER, primary_key=True)
    power_unit = db.Column(TEXT, nullable=False)

    structures = relationship('Structure', back_populates='power_unit')


class StructureView(MyModelView):
    """Flask-Admin view for Structure model (public.structures table)"""

    column_list = ('structure', 'power_unit')
    form_columns = column_list
    column_editable_list = form_columns

    # I can't get these "form_ajax_refs" to work due to Exception:
    # Unsupported field type: <class 'flask_admin.model.fields.AjaxSelectField'>...
    form_ajax_refs = {
        'power_unit': {
            'fields': [PowerUnit.power_unit], # searchable fields
            'minimum_input_length': 0,
            'placeholder': 'Please select',
            'page_size': 10,            
        },

        # The following doesn't work either...
        # 'power_unit': QueryAjaxModelLoader(
        #     'power_unit', db.session, PowerUnit, fields=['power_unit']
        # )
    }
```

Here's a picture of the long dropdown menu when editing the `power_unit` field:
[![enter image description here][1]][1]

When I inspect the HTML, I see a long array of name-value pairs for the dropdown menu, and this array is repeated for every `power_unit` cell in the `structures` table view, so it's a lot of HTML to render, which I think slows down the page loading considerably.
[![enter image description here][2]][2]

  [1]: https://i.stack.imgur.com/o1uEw.png
  [2]: https://i.stack.imgur.com/lFlXS.png

### LINKED ISSUE #2157: Support use of form_ajax_refs as editable columns in list views
Currently ModelView form fields defined in `form_ajax_refs` cannot be used as editable columns in list.html views.

To address this, we can implement ([this solution](https://stackoverflow.com/questions/65193474/can-i-use-form-ajax-refs-and-column-editable-list-for-the-same-column-in-fla))

caffeinatedMike:
Adds support for inline-editing of form_ajax_refs fields from `list.html` views:
- One-to-one relationship
  - [x] Ajax search
  - [x] Select2 field
  - [x] Selections/adjustments persist/are saved to db
  - [x] Editable field reflects the updated/selected value from the modal
- One-to-many relationship
  - [x] Ajax search
  - [x] Select2 multiple field
  - [x] Selections/adjustments persist/are saved to db
  - [x] Editable field does not always reflect the updated (multiple) values once clicking save in the modal. Changes do reflect on refresh though.
 
🐛 Both single and multi-select inline editable fields are unable to have **all** their values deleted from the inline select2 field in the `list.html` views. This relates to #357 and should be addressed when that issue is addressed.

Closes #1627 
Closes #2063 
Closes #2157

### DIFF
--- a/flask_admin/model/widgets.py
+++ b/flask_admin/model/widgets.py
@@ -79,12 +79,48 @@ class XEditableWidget(object):
         display_value = kwargs.pop('display_value', '')
         kwargs.setdefault('data-value', display_value)
 
+        lazy_load = field.type in ('AjaxSelectField', 'AjaxSelectMultipleField')
+        if lazy_load:
+            kwargs.setdefault('data-url-lookup', get_url('.ajax_lookup', name=field.loader.name))
+
         kwargs.setdefault('data-role', 'x-editable')
         kwargs.setdefault('data-url', './ajax/update/')
 
         kwargs.setdefault('id', field.id)
         kwargs.setdefault('name', field.name)
         kwargs.setdefault('href', '#')
+
+        if lazy_load:
+            kwargs.setdefault('type', 'hidden')
+            kwargs.setdefault(
+                'data-placeholder',
+                field.loader.options.get('placeholder', gettext('Search'))
+            )
+            kwargs.setdefault(
+                'data-minimum-input-length',
+                int(field.loader.options.get('minimum_input_length', 0))
+            )
+
+            if field.type == 'AjaxSelectMultipleField':
+                result = []
+                ids = []
+
+                for value in field.data:
+                    data = field.loader.format(value)
+                    result.append(data)
+                    ids.append(as_unicode(data[0]))
+
+                separator = getattr(field, 'separator', ',')
+
+                kwargs['value'] = separator.join(ids)
+                kwargs['data-json'] = json.dumps(result)
+                kwargs['data-multiple'] = u'1'
+            else:
+                data = field.loader.format(field.data)
+
+                if data:
+                    kwargs['value'] = data[0]
+                    kwargs['data-json'] = json.dumps(data)
 
         if not kwargs.get('pk'):
             raise Exception('pk required')
@@ -148,6 +184,8 @@ class XEditableWidget(object):
         elif field.type in ['FloatField', 'DecimalField']:
             kwargs['data-type'] = 'number'
             kwargs['data-step'] = 'any'
+        elif field.type in ['AjaxSelectField', 'AjaxSelectMultipleField']:
+            kwargs['data-type'] = 'select2'
         elif field.type in ['QuerySelectField', 'ModelSelectField',
                             'QuerySelectMultipleField', 'KeyPropertyField']:
             # QuerySelectField and ModelSelectField are for relations
--- a/flask_admin/model/widgets.py
+++ b/flask_admin/model/widgets.py
@@ -78,11 +78,6 @@ class XEditableWidget(object):
     def __call__(self, field, **kwargs):
         display_value = kwargs.pop('display_value', '')
         kwargs.setdefault('data-value', display_value)
-
-        lazy_load = field.type in ('AjaxSelectField', 'AjaxSelectMultipleField')
-        if lazy_load:
-            kwargs.setdefault('data-url-lookup', get_url('.ajax_lookup', name=field.loader.name))
-
         kwargs.setdefault('data-role', 'x-editable')
         kwargs.setdefault('data-url', './ajax/update/')
 
@@ -90,7 +85,8 @@ class XEditableWidget(object):
         kwargs.setdefault('name', field.name)
         kwargs.setdefault('href', '#')
 
-        if lazy_load:
+        if field.type in ('AjaxSelectField', 'AjaxSelectMultipleField'):
+            kwargs.setdefault('data-url-lookup', get_url('.ajax_lookup', name=field.loader.name))
             kwargs.setdefault('type', 'hidden')
             kwargs.setdefault(
                 'data-placeholder',
--- a/flask_admin/static/admin/js/form.js
+++ b/flask_admin/static/admin/js/form.js
@@ -456,7 +456,6 @@
                 return true;
             case 'x-editable':
                 var choices = {};
-                var text_to_pk = {};
                 $el.editable({
                     params: overrideXeditableParams,
                     combodate: {
@@ -494,7 +493,6 @@
                                 for (var k in data) {
                                     var v = data[k];
                                     choices[v[0]] = v[1];
-                                    text_to_pk[v[1]] = parseInt(v[0]);
                                     results.push({ id: v[0], text: v[1] });
                                 }
 
@@ -515,7 +513,6 @@
                                     for (var k in value) {
                                         var v = value[k];
                                         choices[v[0]] = v[1];
-                                        text_to_pk[v[1]] = parseInt(v[0]);
                                         result.push({id: v[0], text: v[1]});
                                     }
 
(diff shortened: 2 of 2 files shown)


michaelbukachi:
Yes. I'm referring to state on the frontend.

caffeinatedMike:
@michaelbukachi Sorry for falling off the map. It's been a long and tough week since my last comment. I finally have some time tomorrow to have a look at this.

michaelbukachi:
> @michaelbukachi Sorry for falling off the map. It's been a long and tough week since my last comment. I finally have some time tomorrow to have a look at this.

No pressure :slightly_smiling_face:

caffeinatedMike:
@michaelbukachi I'm just getting back to this. After looking at this, it seems it's a matter of determining the type of relationship (one-to-one or one-to-many). In the case of the Structures column at the `/powerunit` endpoint, it's a one-to-one (a structure can only be related to one powerunit). I'm struggling to figure out a way to figure out the relationship type ahead of time before the list view is rendered, otherwise it'll be as easy as adding a data-attribute to the column indicating which relationship type it is. From there I can use js to reflect the changes appropriately without refreshing the page.

I just can't see in the code where I can inject the additional data-attribute. Thoughts?

caffeinatedMike:
mrjoes You might be the better party to ask here since you built the internals. How could I go about passing on the type of relationship an Ajax field is dealing with? Ideally I'd like to add it as a data-attribute to the editable field, so we would ultimately need to pass this value to the xeditable widget, right? The question is how?

michaelbukachi:
@caffeinatedMike sorry for the late response. This slipped my mind so I forgot to take a look. I'll take a look at it during the day.

michaelbukachi:
> @michaelbukachi I'm just getting back to this. After looking at this, it seems it's a matter of determining the type of relationship (one-to-one or one-to-many). In the case of the Structures column at the `/powerunit` endpoint, it's a one-to-one (a structure can only be related to one powerunit). I'm struggling to figure out a way to figure out the relationship type ahead of time before the list view is rendered, otherwise it'll be as easy as adding a data-attribute to the column indicating which relationship type it is. From there I can use js to reflect the changes appropriately without refreshing the page.
> 
> I just can't see in the code where I can inject the additional data-attribute. Thoughts?

Are fields aware of the data they hold? Or the models they are associated with? If we are able to determine the type as `AjaxSelectField` or `AjaxSelectMultipleField` maybe we can pull more information about the model. If that information does not exist, perhaps we can add new fields which extend the two mentioned which hold that information? :thinking:. Let me try and prototype something.

caffeinatedMike:
@michaelbukachi Only while they're first being instantiated (example: within `__call__(...)` functions for input widgets). So, like all other data that is passed to those widgets via the `kwargs`, there definitely should be a way to pass on the relationship type if a field is one of the Ajax fields that contain the model.

After reviewing the codebase a little more, I believe the spot that is the most ideal for determining this is the `create_editable_list_form` function in `flask_admin.model.form`

michaelbukachi:
I see. Maybe we can introduce enums for the relationship types and pass them around?

caffeinatedMike:
My thoughts exactly, but haven't figured out where I can inject the data-attribute value other than on the widget level.

caffeinatedMike:
@michaelbukachi `flask_admin.model.base` > `get_list_form` function docstring shows a perfect example of how to add extra data-attributes to the widgets utilized in editable list fields, now we just need to figure out how to get the relationship type there. Could you have a look?

michaelbukachi:
@caffeinatedMike I've been playing with the widgets. I'm actually really close to figuring it out.

caffeinatedMike:
Awesome! Please let me know your findings when you do. This has been a thorn in my side for a bit. It's really the only thing stopping this PR from being merge-ready.

michaelbukachi:
@caffeinatedMike We might not have to go through overriding `get_list_form`. Is the choice of using `AjaxSelectField` over `AjaxSelectMultipleField` or vice versa determined based on the relationship type? Or is it something the user has to specify :thinking: ?

caffeinatedMike:
@michaelbukachi It is. However, what we need is the opposite relationship. For example, in the code we're using to test:
```python3
form_ajax_refs = {
    "structures": QueryAjaxModelLoader(
        "structures", db.session, Structure, fields=["name"]
    )
}
```
What we need is to determine the `direction.name` of Structure.power_unit, which is `MANYTOONE` (many structures to one power_unit). Since the `ModelView` is for `PowerUnit` it isn't clear how to get there. `PowerUnit.structures` is accessible easy enough (which is `ONETOMANY`). We can't simply assume the relationships are absolute opposites since there are legitimate reasons for one end to enforce a `ONETOONE` as well.

caffeinatedMike:
@michaelbukachi That's what I've been trying to figure out: where to do this and how to ultimately pass it onto the widget kwargs.

michaelbukachi:
> @michaelbukachi That's what I've been trying to figure out: where to do this and how to ultimately pass it onto the widget kwargs.
The first step is determining how is `AjaxSelectField` or `AjaxSelectMultipleField` selected.

michaelbukachi:
> `flask_admin.contrib.sqla.form.AdminModelConverter` functions `_model_select_field` and `_convert_relation`

This is where we should determine the relationship and return the appropriate widget which has the relationship information.

caffeinatedMike:
Yes, that's why I was pointing it out... I'm currently tied up with a few other tasks at the moment, but putting it here as a note.

michaelbukachi:
@caffeinatedMike so checking `_convert_relation`, as long a relationship is `ONETOMANY`, `MANYTOMANY` and `uselist` is set then `AjaxSelectMultipleField` is used. Anything else `MANYTONE` or if `uselist` is false (meaning `ONETOONE`) then `AjaxSelectField` is used. So just checking the type of field being used is more than enough.

caffeinatedMike:
@michaelbukachi I'm just getting back to this now. The logic you provided is contradictory to the example code we've been using to test. In your gif of the bug each PowerUnit can have many Structures, but a Structure can only belong to a single PowerUnit at a time. So, just because the inline field is a SelectMultiple that doesn't mean we can assume that values should remain. The values should only remain in MANYTOMANY. Net net, we need a different way to determine this because using just the field type of SelectMultiple as an indicator would mean having a 50% chance of getting it wrong.

michaelbukachi:
@caffeinatedMike what needs to be displayed for `MANYTOMANY` as opposed to `ONETOMANY`. Maybe I'm interpretating this in the wrong way 🤔

caffeinatedMike:
@michaelbukachi nothing needs to be done for MANYTOMANY. For ONETOMANY we need some sort of data-attribute to let us know to iterate over all other entries present on the current page in order to remove the value from those. In your gif you point out how the element is not removed from the other field until refresh. Thats the situation where we need a signifier to tell us to do this via JS to avoid needing a page refresh.

mccarthysean:
FYI, I didn't realize you had done this already. I added the same functionality in the following pull request:

https://github.com/flask-admin/flask-admin/pull/2256

PR #2172, opened 2021-11-04
Flake8 fixes

gunlinux:
Should fix atleast flake8 ci fall


gunlinux:
I start fixing CI, that one fix 1st break :|

PR #2184, opened 2021-11-19
Change edit_view `Save and continue` url

### LINKED ISSUE #2183: `Save and continue` issue in edit view
I have a flask_admin application behind a reverse proxy which rewrites the prefix of the url.
The issue is that when working in the edit view, the `Save and continue` button uses `request.url` which is missing the prefix.
So we are redirected to a wrong url.

It works ok in the create_view but fails in the edit_view.

I think we should change [this](https://github.com/flask-admin/flask-admin/blob/master/flask_admin/model/base.py#L2147):
```python
return redirect(request.url)
```

To same as create_view:

```python
return redirect(self.get_url('.edit_view', id=self.get_pk_value(model)))
```

I'll create a simple PR for it.

aminalaee:
Fixes #2183.

When using flask_admin behind reverse proxy, working in the edit view, the Save and continue button uses `request.url` which is missing the prefix.

This works ok in the create_view, We need to use the `self.get_url('.edit_view')` so the correct url is returned.

### DIFF
--- a/flask_admin/model/base.py
+++ b/flask_admin/model/base.py
@@ -2144,7 +2144,7 @@ class BaseModelView(BaseView, ActionsMixin):
                 if '_add_another' in request.form:
                     return redirect(self.get_url('.create_view', url=return_url))
                 elif '_continue_editing' in request.form:
-                    return redirect(request.url)
+                    return redirect(self.get_url('.edit_view', id=self.get_pk_value(model)))
                 else:
                     # save button
                     return redirect(self.get_save_return_url(model, is_created=False))


aminalaee:
@mrjoes @briancappello Can you review this?

aminalaee:
@ulgens Can you review this?

ulgens (approved):

ulgens:
Looks good and makes sense to me.

aminalaee:
@ulgens Thanks.
Any plans for new release of flask_admin? I can see there's been some fixes, besides this one.

ulgens:
> Any plans for new release of flask_admin? I can see there's been some fixes, besides this one.

Not sure if the question was for me, but I will try my best to push for a new release after couple other fixes.

PR #2195, opened 2021-12-03
wtf form rollback

gunlinux:


ulgens:
I think it will be better in a form of revert commit and with a reference to original PR.

mrjoes:
I assume previous commit broke compatibility with some wtforms version(s)?

gunlinux:
Yeah, i think we need to rebuild it for passing both wtf versions. But right now passing build more important

PR #2217, opened 2022-01-23
Pass datetime format instead of replacing it.

### LINKED ISSUE #2216: DateTimeField with default time incompatible with WTForms 3.0.1
I have a model including a date

```python
class MyClass(db.Model):
    id = db.Column(db.Integer, unique=True, primary_key=True, index=True)
    date = db.Column(db.DateTime, default=datetime.utcnow)
    ...
```

After upgrading WTForms to version 3.0.1 the corresponding field in the form generated by FLaskAdmin (v1.5.3) now renders with "%"  as default, which also causes the timpepicker widget to become useless as well.

I suppose the new behaviour is a result of one [commit in the wtforms repo](https://github.com/wtforms/wtforms/commit/522c18ab672ab8b923f3f04cbeaa81b3282baf3e#diff-2a73ffb3cfb9f7ba01ff1df8d42594abe44d120d5f672233425ddf0e06221aa8) but wasn't able to find a workaround so far.

shoeffner:
Resolves #2216.

Without this fix, DateTimePickers look like this:
image

With this fix, they become:
image

I am not sure if this breaks older WTF versions. If so, it might be a good idea to add 
```
if not self.format:
    self.format = '%Y-%m-%d %H:%M:%S'
```
after the call to `super()`.

### DIFF
--- a/flask_admin/form/fields.py
+++ b/flask_admin/form/fields.py
@@ -36,9 +36,9 @@ class DateTimeField(fields.DateTimeField):
             :param kwargs:
                 Any additional parameters
         """
-        super(DateTimeField, self).__init__(label, validators, **kwargs)
+        format = format or '%Y-%m-%d %H:%M:%S'
+        super(DateTimeField, self).__init__(label, validators, format, **kwargs)
 
-        self.format = format or '%Y-%m-%d %H:%M:%S'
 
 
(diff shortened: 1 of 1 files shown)


shoeffner:
Ah there's already a fix: #2190, so this is a duplicate. Sorry!

PR #2220, opened 2022-02-02
In contrib/geoa factor out tileLayer(...).addTo(map) and update mapbox attributes

TomGoBravo:
This change fixes [issue 2352](https://github.com/flask-admin/flask-admin/issues/2352).

Details: 
* updates the mapbox attributes at match https://docs.mapbox.com/help/troubleshooting/migrate-legacy-static-tiles-api/
* Factors out the two calls to `L.tileLayer` into one
* Fixes `geom_formatter` to no longer return `<textarea data-tile-layer-url="None" ...` when `view.tile_layer_url` is None.

## Tested

Setup postgis and ran the demo app following `examples/geo_alchemy/README.rst` then created an object in the polygon table. Without this change http://127.0.0.1:5000/admin/polygon/ doesn't show tiles and with it tiles appear.

### DIFF
--- a/flask_admin/contrib/geoa/typefmt.py
+++ b/flask_admin/contrib/geoa/typefmt.py
@@ -7,16 +7,23 @@ from sqlalchemy import func
 
 
 def geom_formatter(view, value):
-    params = html_params(**{
+    kwargs = {
         "data-role": "leaflet",
         "disabled": "disabled",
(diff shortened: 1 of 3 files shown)


TomGoBravo:
@mrjoes Please take a look. Thank you.

PR #2221, opened 2022-02-02
Change map search to leaflet-control-geocoder with Nominatim

TomGoBravo:
This PR removes code using Google Maps API and adds leaflet-control-geocoder. By default it uses the OSM Nominatim service which seems to work without an API key. Enable search control by default in 

## Tested

Setup postgis and ran the demo app following examples/geo_alchemy/README.rst then created an object in the polygon table. With this change http://127.0.0.1:5000/admin/polygon/edit/?id=1&url=%2Fadmin%2Fpolygon%2F there is now a little search icon in the top right that can find place names.


samuelhwilliams:
Hey 👋  this feels like a good change to default behaviour - would you be willing to update this PR? 🙏

PR #2233, opened 2022-03-02
Fix Navbar

Mo8it:
`nav` is missing which does cause the problem of the view links to expand the container (see screenshot):
image

After adding `nav`:
image

`nav` is only missing in bootstrap4. In bootstrap3, `nav` is there:
https://github.com/flask-admin/flask-admin/blob/3f9bda85b3a6040ee0327474fbf5e3515467b58a/flask_admin/templates/bootstrap3/admin/base.html#L56

Might be forgotten during the migration.


Mo8it:
@mrjoes @caffeinatedMike One liner is easy to merge. Thanks a lot for the awesome package!

caffeinatedMike (approved):

PR #2044, opened 2020-10-25
Support Import data

Insutanto:
#1353  I have seen this issue is still opening and realize I can finish it.
Support import feature through tablib.
import feature:
* Add rows
* Support CSV and XLS
* Support create data
* Run everything in one DB transaction as default - all or nothing
* Support Bootstrap3 and Bootstrap4
* Support modal

PR #2247, opened 2022-04-12
Support import data via csv or xls

caffeinatedMike:
This closes #2044 as it picks up where that PR left off.

cc: @mrjoes @michaelbukachi @Insutanto @DaniilAnichin please feel free to offer feedback

### DIFF
--- a/examples/bootstrap4/app.py
+++ b/examples/bootstrap4/app.py
@@ -1,6 +1,6 @@
-import datetime
 import os
 import os.path as op
+from datetime import datetime
 from flask import Flask
 from flask_sqlalchemy import SQLAlchemy
 
@@ -29,7 +29,7 @@ class User(db.Model):
     name = db.Column(db.Unicode(64))
     email = db.Column(db.Unicode(64))
     active = db.Column(db.Boolean, default=True)
-    created_at = db.Column(db.DateTime, default=datetime.datetime.now)
+    created_at = db.Column(db.DateTime, default=datetime.now)
 
     def __unicode__(self):
         return self.name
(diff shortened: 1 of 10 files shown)


caffeinatedMike:
@michaelbukachi Can you check my work here, please? Everything seems to be in order now.

@mrjoes For some reason the following 2 tests are failing, any idea why when I haven't touched the related components?
```
flask_admin/tests/test_model.py::test_safe_redirect
flask_admin/tests/test_model.py::test_mockview
```

ibushong:
#2269 should (hopefully) address those test failures

michaelbukachi:
@ibushong so should we prioritize that first?

ibushong:
@michaelbukachi I think that's prob a good idea.

Dhruv-Sachdev1313:
Hey @caffeinatedMike : I tried using your forked repo to have the import functionality. Even though it say `Records added successfully` but doesn't really add any rows. 

On further debugging, we find in this code by printing imported data, it gives an empty list!, so the code never enters the loop and directly returned True
```
 model_columns = {
            header: reversed_labels[header]
            if header in reversed_labels
            else dbify_name(header)
            for header in imported_data.headers
        }
        print("imported data is")
        print(type(imported_data))
        print(imported_data.export('json'))
        for row in imported_data:
            print(row)
            try:
                model = self.model(**{
                    model_columns[key]: row[i]
                    if model_columns[key] not in self.column_formatters_import
                    else self.column_formatters_import[model_columns[key]](self, row[i])
                    for i, key in enumerate(imported_data.headers)
                })
                self.session.add(model)
                self.session.commit()
```

PR #2252, opened 2022-05-18
Contrib: add example for managing hierarchical data

jayaddison:
Provides an example `flask-admin` site that provides management of a basic tree-oriented model implemented using ~~[`sqlalchemy-mptt`](https://sqlalchemy-mptt.readthedocs.io/) on the backend and~~ [`jquery-treetable`](https://plugins.jquery.com/treetable/) on the frontend.

This is something I've wanted to build for a data management task, so it seemed to make sense to begin by implementing it in a way that can hopefully be re-used and improved.

The code here was originally based on the [`sqla-custom-inline-forms`](https://github.com/flask-admin/flask-admin/tree/3f9bda85b3a6040ee0327474fbf5e3515467b58a/examples/sqla-custom-inline-forms) example.

The current implementation is a bit 'lazy' in that it performs a page reload after tree nodes are moved.  That's largely to avoid the complexity of trying to keep client-side tree ordering and consistency in-sync with the server-side representation.

## Demonstration
https://user-images.githubusercontent.com/55152140/169123091-dc7a2c12-3f4f-4f70-8829-0856fd89d22a.mov


jayaddison:
With pagination disabled as of 15d7bc3, this functionality works correctly for datasets larger than the default of 20 items per page, but I'm seeing some performance issues when rendering a set of ~2.5k items.

Both retrieval of a response from the server and also in-browser rendering of the treetable after the response has been loaded appear to be sluggish.

I'll try to investigate where those performance costs are coming from and whether there are straightforward ways to reduce them.

jayaddison:
Ok, one half of the performance issue (data transfer / page size) has been solved.  The rendered page was 500MB+ (!) for ~2.5k records, and the cause was the [`data-source` HTML attribute](https://github.com/flask-admin/flask-admin/blob/3f9bda85b3a6040ee0327474fbf5e3515467b58a/flask_admin/model/widgets.py#L169) that is rendered in the list view to support inline editing.

In particular that's because the [`parent` self-reference relation is editable](https://github.com/jayaddison/flask-admin/blob/8b5ca1c5d84c1956fc8c25502697255c23b1c8ac/examples/sqla-mptt/app.py#L32), meaning there were ~2.5k JSON-rendered entries for each of the ~2.5k rows of data (square product).

Server-side page rendering and data transfer to the client is now almost instantaneous, but client-side construction of the treetable remains sluggish.  Figuring out what's going on there will likely require some profiling and tracing within the browser.  My best guess/hope at the moment is that it's something to do with [this ID-retrieval-and-assignment loop](https://github.com/jayaddison/flask-admin/blob/8b5ca1c5d84c1956fc8c25502697255c23b1c8ac/examples/sqla-mptt/templates/admin/treelist.html#L23-L30).

jayaddison:
From using Firefox's developer tools "Performance" tab, it looks like much of the cost may be due to the initial [expansion of nodes in the tree](https://github.com/ludo/jquery-treetable/blob/73690a9df26dbc1a49981f9335df2a9cd9808813/jquery.treetable.js#L73-L91) that is performed by [`jquery-treetable`](https://github.com/ludo/jquery-treetable/).

image

This sounds like the same issue as reported upstream at ludo/jquery-treetable#144 -- and that has been addressed with a pull request.  Perhaps upgrading to a more recent version of `jquery-treetable` will improve performance here.

jayaddison:
Yes, updating to https://github.com/ludo/jquery-treetable/commit/aa784e2f5a7b057215225b5649ed91c6a67deddc results in a significant improvement in page-load time - from a ~60s to ~5s page render on local hardware here.

jayaddison:
~40% of the remaining rendering delay is now within the branch sort code introduced in 613e2c0.

That code is only necessary when data is received out-of-tree-order from the server.  If we can pre-sort the table data into tree order -- which [`sqlalchemy-mptt`](https://pypi.org/project/sqlalchemy_mptt/) may be able to help with -- then we can remove that sorting code for another significant performance improvement.

jayaddison:
Moving back into draft status while considering whether to remove the `sqlalchemy-mptt` dependency here.

Reason: in [`openculinary/backend`](https://github.com/openculinary/backend/), use of a parent-child `sqlalchemy`' [`relationship`](https://github.com/openculinary/backend/blob/196c3b31c16a0915f9de4ab39e6c0f1d7d16a371/reciperadar/models/recipes/product.py#L31-L36) plus a self-referential [`joinedload`](https://github.com/openculinary/backend/blob/196c3b31c16a0915f9de4ab39e6c0f1d7d16a371/reciperadar/admin/products.py#L48) is proving practical for a ~2.5k item dataset, and avoids the need for an additional dependency.

jayaddison:
Yep, I think the code is a little bit simpler without the `sqlalchemy-mptt` dependency.  It's likely useful in situations where the ability to navigate the tree nodes in-place is valuable; that's not really the case here.  Opening this for potential review.

jayaddison:
I'm not sure this is great quality example code; the tree-view functionality is a bit cobbled-together.  Closing.

PR #2255, opened 2022-05-29
Changing web srid to db srid

odilxon:
I had ready geofiled with SRID: 3857
and i have data on it
`geometry = db.Column(Geometry(geometry_type="MULTIPOLYGON", srid=3857), nullable = True)`

When i used flask admin with Leaflet, geometry transformed incorrectly, 
i need: [69.224854,41.268517], but i got [0.00069224854,0.00004268517]
Then i saw query in debug toolbar
`SELECT ST_AsGeoJSON(ST_Transform(ST_GeomFromEWKB(%(ST_GeomFromEWKB_1)s), %(ST_Transform_1)s)) AS "ST_AsGeoJSON_1"`
He is transforming coordinates systems
in code i saw that he is always transforming from 4326 to 3857 in my case
But i don't need transforming, because of my srid in db is 3857, not 4326
Then i changed `self.web_srid = self.srid`, to use my srid wheli converting

IF there were other method to fix that, comment

### DIFF
--- a/flask_admin/contrib/geoa/fields.py
+++ b/flask_admin/contrib/geoa/fields.py
@@ -22,6 +22,7 @@ class GeoJSONField(JSONField):
         if self.srid == -1:
             self.transform_srid = self.web_srid
         else:
+            self.web_srid = self.srid
             self.transform_srid = self.srid
         self.geometry_type = geometry_type.upper()
         self.session = session


hasansezertasan:
Thanks for the report, and apologies for how long this has sat.

I don't think we can merge this as-is, but I want to explain why rather than just close it, because there's a real underlying issue worth discussing.

## Why the current `web_srid = 4326` is intentional

The code path on read is:

```python
func.ST_AsGeoJSON(func.ST_Transform(self.data, self.web_srid))
```

i.e. it emits **GeoJSON** to the browser. Per [RFC 7946 §4](https://datatracker.ietf.org/doc/html/rfc7946#section-4), GeoJSON coordinates MUST be WGS84 longitude/latitude (EPSG:4326); the older 2008 spec's `crs` member was removed. So transforming to 4326 before `ST_AsGeoJSON` is the conforming behavior regardless of what SRID the column uses.

Leaflet's [`L.geoJSON`](https://leafletjs.com/reference.html#geojson) layer follows the spec — it reads `[lng, lat]` and then projects through the map's [CRS](https://leafletjs.com/reference.html#crs) (default `L.CRS.EPSG3857` for tile display). The GeoJSON payload itself stays lon/lat; the map CRS only controls screen projection.

If we apply this PR, a column genuinely storing EPSG:3857 metres would emit GeoJSON like `[7706000, 5160000]`, which Leaflet would treat as lon/lat degrees and render somewhere far off the visible world.

## What I think is actually happening in your case

The coordinates you described — `[0.00069224854, 0.00004268517]` instead of `[69.224854, 41.268517]` — are the signature of **lon/lat degree values stored inside a column tagged `srid=3857`**, then transformed *as if* they were 3857 metres into 4326 degrees. The ratio is roughly 1e-5, which is what you'd get dividing degrees by the Mercator scale factor near the equator.

Quick way to confirm:

```sql
SELECT ST_SRID(geometry), ST_AsText(geometry) FROM your_table LIMIT 1;
```

If `ST_AsText` shows values in the ~`POINT(69.22 41.26)` range, the data is lon/lat and the column should be `srid=4326` (or the data needs a one-time `ST_Transform` into real 3857 metres).

## The legitimate feature hiding here

There *is* a real gap: flask-admin's `LeafletWidget` doesn't expose a way to use a non-default Leaflet CRS. If someone genuinely wanted to render a 3857-native map without the round-trip transform, the proper fix would be:

1. Let the widget accept a map CRS (via [proj4leaflet](https://github.com/kartena/Proj4Leaflet) for arbitrary CRSes, or `L.CRS.EPSG3857` directly).
2. Skip `ST_Transform` only when the map CRS matches the column SRID **and** emit non-conforming GeoJSON (or switch to a different transport like EWKT) understood by the configured CRS.

Happy to leave this open as a marker for that work if you'd like to take a run at it — a failing test in `flask_admin/tests/geoa/` that demonstrates the 3857-native map scenario would be the right starting point.

---

*Disclosure: I drafted this comment with help from Claude Code while triaging stale PRs; the references above were verified manually.*

PR #2256, opened 2022-06-01
AJAX and x-editable now allowed, plus uploading images to SQL database

mccarthysean:
### Problem: can't use AJAX and x-editable at the same time.

Together they're awesome. AJAX-filled dropdowns make the list view page load MUCH faster, and x-editable fields greatly improve the user's editing experience in the list view.

Now we can have both!

### Problem: can't save images to the database, only file system

It's more efficient to store images in the database, especially when storing image files in Docker images, which bloat the image size. This way, with the new "ImageUploadFieldDB" class, our images can be stored and served from the SQL database as large binary files 👍


ElLorans:
Thanks for the PR! Unfortunately, after 3 years with no activity, I must close the PR since too big and not passing tests

PR #2257, opened 2022-06-15
Details view like edit and create views

mccarthysean:
I've never had any use for the "details" view. In my opinion, it would be better if the details view looked just like the "create" and "edit" views. At least, that would be more convenient for my IJACK Technologies Inc stakeholders.


mccarthysean:
not a very good idea after all 😂

Sorry about that

PR #2260, opened 2022-06-21
Add missing imports in documentation example

brodokk:
It seem that there is a missing import in two of your documentation example


hasansezertasan (changes requested):
There are multiple missing imports on this page, can you also take a look at those?

brodokk:
Hello, yes no problem, I will look at this as soon as possible!

ElLorans:
@samuelhwilliams makes sense to me, I'd be happy to merge

PR #2263, opened 2022-06-23
Add accessible title attribute to boolean items in list view

### LINKED ISSUE #2262: Feature Request: provide accessible field names for boolean rendered icons in list view
In the list view, the "Edit record" and "Delete record" icons have corresponding accessible text attached to them [using the `title` attribute](https://github.com/flask-admin/flask-admin/blob/3f9bda85b3a6040ee0327474fbf5e3515467b58a/flask_admin/templates/bootstrap4/admin/model/row_actions.html#L34).

image

It could be useful to have similar text attached to boolean values that are rendered as icons/glyphs - particularly when viewing large lists where the column header list has been scrolled out-of-view.

I'll include a possible implementation as a pull request alongside the feature request.

jayaddison:
Resolves #2262.

### DIFF
--- a/flask_admin/model/base.py
+++ b/flask_admin/model/base.py
@@ -1846,7 +1846,7 @@ class BaseModelView(BaseView, ActionsMixin):
                 type_fmt = formatter
                 break
         if type_fmt is not None:
-            value = type_fmt(self, value)
+            value = type_fmt(self, value, name)
 
         return value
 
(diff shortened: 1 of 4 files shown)


mrjoes:
This formatter API change is no longer backwards compatible and will break custom formatters. If we can keep backwards compatibility - looks good.

jayaddison:
The unit test issue _may_ be related to pallets/flask#4624; it's not yet clear to me exactly what the cause is, but it seems to relate to the change in method signature for view function dispatch (removal of positional arguments).  Possibly the same cause as the issue reported here in flask-admin/flask-admin#2289.

PR #2264, opened 2022-06-23
Importing markupsafe instead of MarkupSafe

asdrubalivan:
There's an import error due to `MarkupSafe` module being imported even if it doesn't exist. The import has been renamed as `markupsafe`


samuelhwilliams:
This has been fixed in 45e8b657d now. Thanks :)

PR #2265, opened 2022-06-26
Fix value of `desc` param in filter form (`1` instead of `"True"`)

### LINKED ISSUE #1761: `desc` query string arg incorrectly preserved when setting filters
If you do descending order on a column, the URL will have `?desc=1`. If you then apply a filter, this part of the url becomes `desc=True`, which breaks the `ViewArgs` parsing, resulting in the sorting going back to ascending.

One possible fix (I think..) is to remove the `bool` casting of the `sort_desc` argument to `ViewArgs`

ibushong:
Fixes #1761

"hardcoding" `1` might seem bad, but since it's within the `if sort_desc` conditional, it should be fine. Though it could be replaced by something like `{{ 1 if sort_desc else 0 }}` for extra robustness.

### DIFF
--- a/flask_admin/templates/bootstrap2/admin/model/layout.html
+++ b/flask_admin/templates/bootstrap2/admin/model/layout.html
@@ -41,7 +41,7 @@
         <input type="hidden" name="sort" value="{{ sort_column }}">
         {% endif %}
         {% if sort_desc %}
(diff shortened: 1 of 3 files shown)


samuelhwilliams (approved):
This works - thanks for the fix! 👍

PR #2270, opened 2022-06-27
Added flask-babel as fallback for flask-babelex

TitaniumHocker:
I didn't tested all cases, but it seems to work.


TitaniumHocker:
@mrjoes If needed, I can add note to the documentation.

PR #2273, opened 2022-06-30
Fix unexpected keyword argument for MongoFileField

mbednarik:


mbednarik:
@mrjoes this fix has been tested on the latest mongoengine v0.24.2, but it shouldn't break earlier versions.

alanhamlett (approved):

PR #2280, opened 2022-07-21
ItemsRequired validator: do not include items marked for deletion in field item count

### LINKED ISSUE #1930: sqla.validators.ItemsRequired doesn't sheck for "_should_delete"
I'm using this for an inline model where items can be added and deleted but at least one item is required.

```python
    form_args = {
        'names': {
            'validators': [sqla.validators.ItemsRequired()]
        }
    }

    inline_models = (
        (
            ChannelName,
            dict(
                form_label='Name',
                form_columns=('id', 'name', 'default')
            )
        ),
    )
```

This works on creating items but it's still possible to delete all items.

The validator checks for `if len(field.data) < self.min` only but it should only count items not selected for deletion.

jayaddison:
The [`ItemsRequired` validator](https://github.com/flask-admin/flask-admin/blob/73febb13a18b48ebb21d5927de944cd5edbae068/flask_admin/contrib/sqla/validators.py#L48-L68) offers the ability to require a minimum number of related items for a given field.

One factor that it doesn't currently consider is that during deletion of items, the submitted form field will contain entries for both the existing and deleted items -- so it may over-count the number of remaining items.

This changeset updates the logic so that only non-deleted entries are included in the item count.

Resolves #1930.

### DIFF
--- a/flask_admin/contrib/sqla/validators.py
+++ b/flask_admin/contrib/sqla/validators.py
@@ -6,6 +6,8 @@ try:
 except ImportError:
     from wtforms.validators import Required as InputRequired
 
+from flask_admin._compat import filter_list
+
 
 class Unique(object):
     """Checks field value unicity against specified table field.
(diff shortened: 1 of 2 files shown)


jayaddison:
cc @michaelbukachi @mrjoes for review

michaelbukachi (approved):
LG! @mrjoes I think this can be merged.

jayaddison:
Thanks @michaelbukachi @mrjoes!

Issue #2143, opened 2021-09-03
No Apply button when removing all the filters


giucataldi16:
In the Model views, when you add filters, the Apply button correctly shows up.
However, when removing all the filters, the "Apply" button disappears, and the state of the page doesn't change, so I'm stuck with the filtered data.

In my use case, using the "Reset filters" button is not a viable option, because by default, the page I'm experiencing this issue, has a filter applied. The problem I am trying to solve is preventing me from removing this default filter (when I remove it, I cannot press the Apply button, and if I press the "Reset filters", the page goes back to the default applied filter state).

For now, the workaround is to manually remove the query parameter from the URL of the page, but it would be better for our users to have the possibility to use the Apply button.

<img width="1174" alt="Schermata 2021-09-03 alle 09 31 55" src="https://user-images.githubusercontent.com/54641802/131968095-67992dff-cb56-4d09-8860-8338770efa1f.png">
<img width="1198" alt="Schermata 2021-09-03 alle 09 32 03" src="https://user-images.githubusercontent.com/54641802/131968105-cc0f4239-7097-4ff6-b47e-d37b396c5ab3.png">

samuelhwilliams:
From my quick testing, this appears to have already been fixed, so I will close this. If you can still reproduce this in some circumstances, please do let us know.

![2024-07-13 17 56 54](https://github.com/user-attachments/assets/014a7059-248b-4a60-96fa-ab775a6d5a77)

PR #2285, opened 2022-08-11
Display Apply Button when removing all the filters

lfansu35:
Refs : #2143


samuelhwilliams:
From my quick testing, this appears to have already been fixed, so I will close this. If you can still reproduce this in some circumstances, please do let us know.

PR #2290, opened 2022-09-14
Conditionally provide positional object-instance argument when running wrapped view methods

### LINKED ISSUE #2289: Exception TypeError in MethodView example
Can someone reproduce this error while running the example and clicking on any api-link?
```
Traceback (most recent call last):
  File "/home/fapsi/Projects/flask-admin/examples/methodview/venv/lib/python3.10/site-packages/flask/app.py", line 2548, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/fapsi/Projects/flask-admin/examples/methodview/venv/lib/python3.10/site-packages/flask/app.py", line 2528, in wsgi_app
    response = self.handle_exception(e)
  File "/home/fapsi/Projects/flask-admin/examples/methodview/venv/lib/python3.10/site-packages/flask/app.py", line 2525, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/fapsi/Projects/flask-admin/examples/methodview/venv/lib/python3.10/site-packages/flask/app.py", line 1822, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/fapsi/Projects/flask-admin/examples/methodview/venv/lib/python3.10/site-packages/flask/app.py", line 1820, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/fapsi/Projects/flask-admin/examples/methodview/venv/lib/python3.10/site-packages/flask/app.py", line 1796, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/home/fapsi/Projects/flask-admin/examples/methodview/venv/lib/python3.10/site-packages/flask_admin/base.py", line 69, in inner
    return self._run_view(f, *args, **kwargs)
  File "/home/fapsi/Projects/flask-admin/examples/methodview/venv/lib/python3.10/site-packages/flask_admin/base.py", line 368, in _run_view
    return fn(self, *args, **kwargs)
TypeError: View.as_view.<locals>.view() takes 0 positional arguments but 1 was given
```

jayaddison:
This is a work-in-progress, ~~and requires a tweak in upstream `flask` (PR pending),~~ so I'm opening it for discussion in a draft state.

Could resolve #2289.

### DIFF
--- a/flask_admin/base.py
+++ b/flask_admin/base.py
@@ -67,7 +67,18 @@ def _wrap_view(f):
         if abort is not None:
             return abort
 
-        return self._run_view(f, **kwargs)
+        # Workaround for https://github.com/flask-admin/flask-admin/issues/2289
+        #
+        # From flask version 2.2.0 onwards, view methods accept keyword arguments only, and
+        # should not accept positional arguments.
+        #
+        # flask-admin wraps a mixture of instance and non-instance view functions; it would
+        # be preferable to distinguish those and use separate wrapping techniques for each,
+        # but here we check the function signature for positional arguments and provide the
+        # view object (self) if any are expected.
+        spec = inspect.getfullargspec(f)
+        args = [self] if spec.args else []
+        return self._run_view(f, *args, **kwargs)
 
     inner._wrapped = True
 
@@ -366,19 +377,7 @@ class BaseView(with_metaclass(AdminViewMeta, BaseViewClass)):
             :param kwargs:
                 Arguments
         """
-        # Workaround for https://github.com/flask-admin/flask-admin/issues/2289
-        #
-        # From flask version 2.2.0 onwards, view methods accept keyword arguments only, and
-        # should not accept positional arguments.
-        #
-        # flask-admin wraps a mixture of instance and non-instance view functions; it would
-        # be preferable to distinguish those and use separate wrapping techniques for each,
-        # but here we check the function signature for positional arguments and provide the
-        # view object (self) if any are expected.
-        spec = inspect.getfullargspec(fn)
-        if spec.args:
-            return fn(self, **kwargs)
-        return fn(**kwargs)
+        return fn(*args, **kwargs)
 
     def inaccessible_callback(self, name, **kwargs):
         """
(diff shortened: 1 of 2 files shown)


jayaddison:
@mrjoes @davidism There's a cross-compatibility issue here between `flask-admin` and `Flask >= 2.2.0` here.  I thought it might be possible to address it by permitting the [class-based-view `view` method signatures](https://github.com/pallets/flask/blob/9c3deeee963e62dfb60c9343fad4ec2ec752fdba/src/flask/views.py#L101-L113) to accept a single (anonymous) positional argument (`self`, as `_`), but in practice when I attempt that, it breaks purely-functional (non-instance) based views (that _should not_ accept any positional arguments).

I'm digging around to see whether there's a clean way to handle this (preferably in a backwards-compatible way), but I feel like it's near the limits of the complexity I can understand - so I'm mentioning you in this thread in case you have any guidance/thoughts based on your experience.

davidism:
I think you're referring to this https://github.com/pallets/flask/pull/4776, which explains why positional args won't be supported directly in Flask. I don't know enough about Flask-Admin to say what's going on here or what the appropriate solution is.

jayaddison:
I've pushed further commits with a workaround, but it doesn't feel like it's the correct solution.  It'd be preferable to distinguish the situations during which view functions are wrapped into class-based and function-based categories - but I haven't yet figured out how to do that.

It's possible that some of the original reason that wrappers were added was to avoid repeated work during re-instantiation of view classes -- something that pallets/flask#4624 could help address natively -- although that would require an interface upgrade.

The latest commits solve a few unit test failures, but other failures remain due to an issue with repeat blueprint registration (a separate issue).

mrjoes:
I'm going to take a look this weekend.

Long story short: flask-admin is using custom class-based view, where some method can be decorated and automatically exposed as flask routes. There's a helper method that allows injecting/adjusting arguments passed to view methods for each and every request. And it's passing `self` as a first positional argument to a wrapped method.

jayaddison:
Experimented a little more with this today -- including some attempts to rewrite the function wrapping to be called from a static context (`__new__` instead of `__init__`) in order to find a cleaner solution, but neat solutions evaded me.  I'm going to close this pull request as stale / unlikely to be continued.

PR #2309, opened 2022-11-24
Fix test

BenjaminLECHA:
Fix all flask_admin test (except one on @base.expose_plugview)


alanhamlett:
Tests are already passing on master branch so I think this PR is no longer necessary. Let me know if I should re-open it.

PR #2315, opened 2022-12-29
fix: example code for sqla-custom-inline-forms

Titotix:
Update example code for sqla-custom-inline-forms.

- Fix import error due to WTForms version outdated
- Fix out of flask app context error for `db.create_all()`


samuelhwilliams:
Thanks for this. As it's gone stale, I'll close this PR. Have pulled your commit into the related PR above.

Issue #2319, opened 2023-01-12
InlineHstoreList is missing extra_filters parameter which can crash wtforms


Zasgard:
InlineHstoreList currently looks this with the process inheriting process and calling the super. However, the super version has extra_filters which can be called if using the newest wtforms. This needs to be included in InlineHstoreList to prevent crashes
```
class InlineHstoreList(InlineFieldList):
    """ Version of InlineFieldList for use with Postgres HSTORE columns """

    def process(self, formdata, data=unset_value):
        """ SQLAlchemy returns a dict for HSTORE columns, but WTForms cannot
            process a dict. This overrides `process` to convert the dict
            returned by SQLAlchemy to a list of classes before processing. """
        if isinstance(data, dict):
            data = [KeyValue(k, v) for k, v in iteritems(data)]
        super(InlineHstoreList, self).process(formdata, data)
```
^ Original

```
def process(self, formdata, data=unset_value, extra_filters=None):
        res = super(InlineFieldList, self).process(
            formdata, data)

        # Postprocess - contribute flag
        if formdata:
            for f in self.entries:
                key = 'del-%s' % f.id
                f._should_delete = key in formdata

        return res

```
^Inherited process method

```
def process(self, formdata, data=unset_value, extra_filters=None):
        """ SQLAlchemy returns a dict for HSTORE columns, but WTForms cannot
            process a dict. This overrides `process` to convert the dict
            returned by SQLAlchemy to a list of classes before processing. """
        if isinstance(data, dict):
            data = [KeyValue(k, v) for k, v in iteritems(data)]
        super(InlineHstoreList, self).process(formdata, data, extra_filters)
```

^My fix that resolves the issue. Just needed to add the extra_filters into the super to allow the extra_filters to be properly passed on as utilized in the WTForms code (wtforms>form.py>line127)

Zasgard:
I don't have rights to submit a pr so I just forked it temporarily with the fix in.

alanhamlett:
Closed with #2320.

PR #2320, opened 2023-01-12
Fixes missing parameters for InlineHstoreList

Zasgard:
Fixes parameters for method that causes issues. See associated issue #2319


michaelbukachi:
@Zasgard Can you link the issue properly using '#'?

Issue #490, opened 2014-03-27
creating/editing nullable unique properties


abathur:
Apologies if I'm duplicating a known issue--I've looked around for a while now and haven't found a report that seems to line up, but it seems like something others would've run into already.

We've got a few nullable unique string fields in our models; when we create a new instance of one of these models through the administrator, leaving one of these fields empty results in flask-admin trying to create them with zero-length strings instead of null values, which causes an integrity error on the uniqueness constraint the second time you try to create a model in this way.

We're running Flask-Admin==1.0.7, for reference. Hopefully there's a workaround or configuration option I'm missing.

PR #2321, opened 2023-01-19
String null fix

### LINKED ISSUE #1840: Unexpected differing behavior for db.String and db.Text on Postgres (latter saves empty field as "")
This seems related to a bunch of closed issues around handling nullable and non-nullable fields, but I recently was stumped by the following (IMO) weird behavior:

```python
from app import admin, db  # vanilla Flask setup

# Declarative model using Flask-SQLAlchemy, Postgres as the backend
class Document(db.Model):
    description = db.Column(db.String)

class Document2(db.Model):
    description = db.Column(db.Text)

### Add to Flask-Admin instance
admin.add_view(ModelView(Document, db.session))
```

Leads to `Document` instances having a null description field when they're created in Flask-Admin (what I'd expect by default) and `Document2` instances having an empty string. In Postgres, I _believe_ these are actually equivalent, so I can just use `db.String` as a workaround.

Is this a bug? Edge case of something about these column types I'm not thinking of (or that doesn;'t apply to Postgres)? Certainly a bit of a gotcha that I didn't jump out at me in any of the documentation or examples.

dgilman:
This closes #1840. It is probably worth release noting as it is a behavior change.

Digging into the issue, [this comment](https://github.com/flask-admin/flask-admin/issues/1840#issuecomment-502850787  points out that String and Text don't handle falsey strings the same way. Digging into the history of the code, the difference originated with [this commit](https://github.com/flask-admin/flask-admin/commit/902b9c375f988010a8174b26f338807109b0cc94) which was a fix to #490. The author of that issue only brought up String fields, and the change only fixed String fields. I think it was simply an oversight to not handle Text fields the same way at that time, and it was likely because the opener of the ticket didn't happen to bring up both field types. The two classes were otherwise identical in behavior up until that PR.

The unit tests narrowly dodged the issue. Although there are nullable String columns none of them asserted on the casting of empty string to None, which probably would have made the inconsistency more obvious earlier on.

### DIFF
--- a/flask_admin/contrib/sqla/form.py
+++ b/flask_admin/contrib/sqla/form.py
@@ -276,11 +276,11 @@ class AdminModelConverter(ModelConverterBase):
     def _string_common(cls, column, field_args, **extra):
         if hasattr(column.type, 'length') and isinstance(column.type.length, int) and column.type.length:
             field_args['validators'].append(validators.Length(max=column.type.length))
+        cls._nullable_common(column, field_args)
 
     @converts('String')  # includes VARCHAR, CHAR, and Unicode
     def conv_String(self, column, field_args, **extra):
         self._string_common(column=column, field_args=field_args, **extra)
-        self._nullable_common(column, field_args)
         return fields.StringField(**field_args)
 
     @converts('sqlalchemy.sql.sqltypes.Enum')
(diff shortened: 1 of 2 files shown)


samuelhwilliams:
This looks good. I've just fixed up the test cases which needed to switch to some pytest fixtures and tweak indenting. Thanks for contributing 🙏

PR #2322, opened 2023-01-19
Fix many depreciation warnings

dgilman:
Note that this likely bumps to a minimum requirement of SQLAlchemy 1.4+

### DIFF
--- a/flask_admin/contrib/sqla/ajax.py
+++ b/flask_admin/contrib/sqla/ajax.py
@@ -1,4 +1,4 @@
-from sqlalchemy import or_, and_, cast, text
+from sqlalchemy import or_, and_, cast, text, select
 from sqlalchemy.types import String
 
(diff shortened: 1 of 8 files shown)


flask_admin/contrib/sqla/ajax.py
```diff
@@ -58,16 +58,13 @@ def format(self, model):
 
         return getattr(model, self.pk), as_unicode(model)
 
-    def get_query(self):
```
mrjoes, line 61: This is incompatible change. `get_query` allowed subclassing and mutating the generic query object in child admin classes.

mrjoes:
Unfortunately can't accept this one in current form - it changes Flask-Admin SQLa-related APIs.

ankit-gautam23 (approved):
LGTM

alanhamlett:
Closing for now. Will re-open if `get_query` api made backwards compatible.

PR #2326, opened 2023-01-30
Use SQLAlchemy.init_app() in tests and document

cjmayo:
As described in:

https://flask-sqlalchemy.palletsprojects.com/en/3.0.x/quickstart/#configure-the-extension

Was only working by accident prior to Flask-SQLAlchemy 3:

https://github.com/pallets-eco/flask-sqlalchemy/pull/1151

Revealed that geoa tests were creating tables twice outside of an
application context.

---

This came to light when testing with WTForms 3, that is now showing a warning.

Relies on "Add app.app_context() inside SQLAlchemy setup for testing geoa" from #2309 which is included here. As #2309 already needs rebasing as a result of 5c89c59e (and possibly fixes its TODO), I don't think merging this would cause more work.

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -199,7 +199,8 @@ from the GeoAlchemy backend, rather than the usual SQLAlchemy backend::
     from flask_admin.contrib.geoa import ModelView
 
     # .. flask initialization
-    db = SQLAlchemy(app)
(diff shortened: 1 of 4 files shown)


mrjoes:
As long as this backwards (2.x) compatible - I'm fine merging it.

PR #2327, opened 2023-01-30
Stop using Docker for flake8 and sphinx, and test minimum requirements

cjmayo:
A few different things here. Happy to split up. Notably:

- WTForms 1 was not being tested anyway, requirements-dev.txt has wtforms==2.3.3
- Save installing the database containers for flake8 and doc-html
- Test the minimum version of requirements: excercise the current compatibilty code and support maintaining existing versions while enabling newer versions. e.g. this would allow testing WTForms 2 and 3.
Where there wasn't version information in requirements-dev.txt I picked versions that are, or could have been, in Ubuntu 20.04 and then tweaked a few to get them working together (see commit message). Clearly versions could be further adjusted.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -8,17 +8,12 @@ on:
       - master
 
 jobs:
-  test-job:
+  tests:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
         tox-env: [py]
-        include:
-          - python-version: '3.x'
-            tox-env: flake8
-          - python-version: '3.x'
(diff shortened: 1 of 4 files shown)


mrjoes:
I tried current master with latest flask-sqlalchemy (3.0.1) and sqlalchemy (3.0 release), which required Flask>=2.0 and tests passed. However, there are some reports that flask-admin is not compatible with Flask 2.x - it's something I did not have a chance to look at.

I believe requirements-dev.txt should be latest versions of flask-admin dependencies, while requirements-min.txt should represent lowest supported versions.

samuelhwilliams:
I think this is redundant now after integrating the pallets-eco GitHub action workflow files, so I will close this off.

PR #2330, opened 2023-02-01
Fix exceptions when flask-babel is installed but not used

cjmayo:
flask-babel's gettext() expects a Babel object to have been created.

Otherwise an exception is raised. For an example see tests/test_form_upload.py

---

I think this illustrates the problem. Not clear that it is the correct fix.


samuelhwilliams:
Closing to move new discussion here: https://github.com/pallets-eco/flask-admin/issues/2447

PR #2328, opened 2023-02-01
Support latest flask, sqlalchemy, flask-sqlalchemy and wtforms

cjmayo:
For information on the test failures with the latest versions of these key dependencies and a tracker for some of the fixes from other PRs.

PR #2332, opened 2023-02-04
Fix tests with latest dependencies

cjmayo:
The commits tested with latest dependencies in #2328.


samuelhwilliams:
I am going to close this as outdated as we've just updated the testing framework for flask-admin after it was incorporated into the pallets-eco project.

I do think we need to address the testing situation to have better variant testing on some key dependencies, however that may be best to look at from a fresh slate now. Please don't take this as discouragement from helping tighten this up, but as this PR is fairly stale now I think a reboot would be best.

PR #2336, opened 2023-02-09
Test with flask-babel

cjmayo:
To demonstrate what happens when using flask-babel instead of flask-babelex.


samuelhwilliams:
Closing to move new discussion here: https://github.com/pallets-eco/flask-admin/issues/2447

Issue #2034, opened 2020-09-24
Default Leaflet Basemap Tiles


jtbaker:
Would it be possible to move toward a non-vendor [lock in](https://github.com/flask-admin/flask-admin/blob/master/flask_admin/static/admin/js/form.js#L77) from Mapbox for the basemap tile provider? Per #1910 There seem to be some issues with the implementation and the MAPBOX_ID URI reference may be deprecated in their APIs.

There are also several open map tile [providers](https://wiki.openstreetmap.org/wiki/Tile_servers) that do not require an API key or other authentication mechanism. I think one of these (probably the default OpenStreetMap tiles) would make more sense for most users as a default configuration since this is a simple admin tool.

I'll write up a PR for such a change, including the documentation if no one objects.

LvanWissen:
This is possible if you make use of the data attributes you can pass to the form, in particular this setting: https://github.com/flask-admin/flask-admin/blob/c8877df933ed4842e4bb8f2c8e9301c5507a42b4/flask_admin/static/admin/js/form.js#L162

Example:
```python
from flask_admin.contrib.geoa import ModelView

class CityView(ModelView):

    form_widget_args = {
        # field 'coordinates' is of type Geometry in my model
        'coordinates': {
            'data-tile-layer-url': "{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
        }
    }
```

Check out all the `$el.data('my-setting')` configurables that you can set this way in https://github.com/flask-admin/flask-admin/blob/master/flask_admin/static/admin/js/form.js

jtbaker:
Thank you @LvanWissen! I didn't realize it was configurable via subclassing the model.

TheoLechemia:
Thanks for the tip @LvanWissen 
However, we still need put the `MAPBOX_MAP_ID` in the config in we want to have a map.
this parameter control the load of the leaflet lib : https://github.com/flask-admin/flask-admin/blob/master/flask_admin/templates/bootstrap4/admin/lib.html#L260
and other stuff if form.js : https://github.com/flask-admin/flask-admin/blob/master/flask_admin/static/admin/js/form.js#L76
Would it not be better to use a parameter like `MAP_TILE_LAYER` in the config ?
I can make a PR if you approve.

LvanWissen:
Good point, but in terms of features it's only a naming flaw. 
[...]

PR #2338, opened 2023-02-15
Document Environment Variables

TheoLechemia:
Small contribution on documentation.
See https://github.com/flask-admin/flask-admin/issues/1694
No sure if it's the way you want to do it.. tell me

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -545,3 +545,20 @@ While the wrapped function should accept only one parameter - `ids`::
                     raise
 
                 flash(gettext('Failed to approve users. %(error)s', error=str(ex)), 'error')
+
+
+Settings reference
+------------------
+
+This section list settings supported in the current release
+
+  FLASK_ADMIN_SWATCH
+    Theme for bootswatch in order to customize you UI. See https://bootswatch.com/
+    Example : FLASK_ADMIN_SWATCH = "simplex"
+
+  FLASK_ADMIN_FLUID_LAYOUT
+    Boolean. Use this if you want to use the bootstrap "container-fluid" (use the full width of the screen)
+    
+  ADMIN_RAISE_ON_INTEGRITY_ERROR
+    Boolean. Control if any exeption is raised on integrity error or if flash messages are returned
+    Default = None (flash message)


samuelhwilliams:
Thanks for the suggested improvement to document env var configuration on Flask-Admin; I agree this is something we're missing at the moment.

We've just removed env var configuration for theming (swatch, fluid layout) and move it to a Theme config object passed to the admin instance. It looks like we still have two possible env vars: `ADMIN_RAISE_ON_INTEGRITY_ERROR` and `ADMIN_RAISE_ON_VIEW_EXCEPTION`. It would be good to add documentation for those. I think it makes sense to flag these at the bottom of 'introduction.rst' rather than hiding them in 'advanced.rst'.

If you wanted to update the PR I'd be very grateful - if not I'll try to get around to it soon.

doc/advanced.rst
```diff
@@ -545,3 +545,20 @@ While the wrapped function should accept only one parameter - `ids`::
...
 
                 flash(gettext('Failed to approve users. %(error)s', error=str(ex)), 'error')
+
+
+Settings reference
+------------------
+
+This section list settings supported in the current release
+
+  FLASK_ADMIN_SWATCH
```
hasansezertasan, line 555: Version 2 doesn't have this anymore.

doc/advanced.rst
```diff
@@ -545,3 +545,20 @@ While the wrapped function should accept only one parameter - `ids`::
...
+    Theme for bootswatch in order to customize you UI. See https://bootswatch.com/
+    Example : FLASK_ADMIN_SWATCH = "simplex"
+
+  FLASK_ADMIN_FLUID_LAYOUT
```
hasansezertasan, line 559: Version 2 doesn't have this anymore.

hasansezertasan (changes requested):

hasansezertasan:
> Thanks for the suggested improvement to document env var configuration on Flask-Admin; I agree this is something we're missing at the moment.
> 
> We've just removed env var configuration for theming (swatch, fluid layout) and move it to a Theme config object passed to the admin instance. It looks like we still have three possible env vars: `ADMIN_RAISE_ON_INTEGRITY_ERROR`, `ADMIN_RAISE_ON_VIEW_EXCEPTION` and `FLASK_ADMIN_SEARCH_SIZE_MAX`. It would be good to add documentation for those. I think it makes sense to flag these at the bottom of 'introduction.rst' rather than hiding them in 'advanced.rst'.
> 
> If you wanted to update the PR I'd be very grateful - if not I'll try to get around to it soon.

Looks like we have a lot more...

https://github.com/pallets-eco/flask-admin/blob/68bc6de4056c493a44b616e9bc4ac7850c0441fe/flask_admin/templates/bootstrap4/admin/lib.html#L259-L285

samuelhwilliams:
🥲  We should have a consistent namespace for all of these, too.

TheoLechemia:
Maybe I would take a look when the 2.0 will be almost stabilised ? 
I see there is still a `MAPBOX_MAP_ID` which is discuss here : https://github.com/pallets-eco/flask-admin/issues/2034

samuelhwilliams:
Thanks for flagging #2034 - agree that we should fix that interface. I've assigned myself to it and will think about what we want to do there.

For now I'm going to close this PR as I'll put one up shortly that rationalises all of the config variables under a FLASK_ADMIN_ namespace and adds some documentation.

PR #2340, opened 2023-02-17
Remove use of deprecated query.get()

mklassen:
SQLAlchemy has deprecated the use of `query.get()`

https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.get


alanhamlett (approved):
Thanks!

PR #2341, opened 2023-02-17
Tuple field_flags are deprecated

mklassen:
WTForms has deprecated the use of `tuples` for field flags.

https://github.com/wtforms/wtforms/blob/master/src/wtforms/fields/core.py#L131


samuelhwilliams:
Thank you 🙏

samuelhwilliams (approved):

PR #2342, opened 2023-02-17
Fixed AJAX data separator for select2

mklassen:
`select2` needs the `separator` option to properly parse the AJAX results when using custom data separators, so the separator needs to be propagated to the field. 

jQuery 3.5.1 and select2 3.5.2 use `separator` option, but if select2 is updated the `tokenSeparator` array option will be needed instead.


alanhamlett:
Thanks!

PR #2349, opened 2023-04-11
Add in Content Security Policy "nonce" support.

mchineboy:
Relies on Flask-Talisman for setting up the csp_nonce() jinja function.

Otherwise, just places a lambda that returns an empty string.


flask_admin/contrib/mongoengine/view.py
```diff
@@ -343,7 +343,7 @@ def scaffold_sortable_columns(self):
 
         for n, f in self._get_model_fields():
             if type(f) in SORTABLE_FIELDS:
-                if self.column_display_pk or type(f) != mongoengine.ObjectIdField:
+                if self.column_display_pk or not isinstance(f, mongoengine.ObjectIdField):
```
alanhamlett, line 346: These are good improvements but they belong in a separate PR. Since they're not required by your feature, they just make it harder to review.

alanhamlett:
The inline script problem can be solved without adding new dependencies on Tailwind and Flask-Talisman. I don't think we're ready for such a large change to this project, but without those I'll be willing to merge a PR that only fixes the inline script bug.

PR #2354, opened 2023-04-27
exit processLeafletWidget early when called a second time

TomGoBravo:
Fix (or workaround?) for https://github.com/flask-admin/flask-admin/issues/2351


samuelhwilliams:
Thanks for proposing a workaround. I discovered an issue similar to this while fixing a regression I've just spotted after a recent merge. I've taken a different approach, so I'm going to close this down - but thank you for flagging it

The other PR is here: https://github.com/pallets-eco/flask-admin/pull/2468

PR #2359, opened 2023-05-24
Correct translation IT

ElLorans:
Minor correction


ElLorans:
Already opened in https://github.com/flask-admin/flask-admin/pull/2300

PR #2368, opened 2023-07-08
Update lib.html to allow for horizontal forms in Bootstrap 4

mdanse:
Use the FLASK_ADMIN_FORM_HORIZONTAL setting to control using vertical (default) or horizontal forms.

### DIFF
--- a/flask_admin/templates/bootstrap4/admin/lib.html
+++ b/flask_admin/templates/bootstrap4/admin/lib.html
@@ -119,56 +119,61 @@
   {% set direct_error = h.is_field_error(field.errors) %}
   {% set prepend = kwargs.pop('prepend', None) %}
   {% set append = kwargs.pop('append', None) %}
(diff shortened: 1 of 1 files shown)


samuelhwilliams:
Happy to accept this kind of PR in principle.

Flask-Admin now has a `Theme` config object that replaces the need for environment variables like FLASK_ADMIN_FORM_HORIZONTAL. If this is possible for all Bootstrap versions, please can you add it to the base class and add support everywhere?

If it's only available for BS4, please create a new Bootstrap4Theme class and add it only to that.

It might be that should we should just deprecate Bootstrap2Theme and Bootstrap3Theme soon, but haven't made that call yet...

princerb (commented):
[this message was deleted]

samuelhwilliams:
??? Will it?

princerb:
> ??? Will it?

Oh, sorry, that was a mistake. I saw it like it was Bs3.

PR #2397, opened 2023-10-23
Replace inspect.getargspec() that was removed in Python 3.11

### LINKED ISSUE #2343: Flask-Admin wrongfully reported as python 3.11 compatible
Hello,

After migration to Python 3.11, we encountered a bug because it still uses `inspect.getargspec` for a fallback behaviour on filters in  https://github.com/flask-admin/flask-admin/blob/master/flask_admin/contrib/sqla/view.py (and maybe elsewhere), while is has been [removed](https://docs.python.org/3/whatsnew/3.11.html#removed).

As a workaround we've upgraded our filters to support the `alias` parameter, but still this is something that needs to be fixed from this repository, given that pypi documents it as compatible with Python 3.11.

Would it be possible to update the code?
I may contribute if you'd need help.

cjmayo:
Closes  #2343.

### DIFF
--- a/flask_admin/contrib/sqla/view.py
+++ b/flask_admin/contrib/sqla/view.py
@@ -999,7 +999,7 @@ class ModelView(BaseModelView):
             try:
                 query = flt.apply(query, clean_value, alias)
             except TypeError:
-                spec = inspect.getargspec(flt.apply)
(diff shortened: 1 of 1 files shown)


samuelhwilliams (approved):
Thank you 👍

PR #2426, opened 2024-04-19
add mongoengine date field

wangsha:
fixes: https://github.com/flask-admin/flask-admin/issues/1921

### DIFF
--- a/flask_admin/contrib/mongoengine/filters.py
+++ b/flask_admin/contrib/mongoengine/filters.py
@@ -288,7 +288,7 @@ class FilterConverter(filters.BaseFilterConverter):
     def conv_float(self, column, name):
         return [f(column, name) for f in self.float_filters]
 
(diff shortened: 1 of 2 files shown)


samuelhwilliams (changes requested):
This feels like a useful change in principle, but right now it doesn't work.

The filter value popup shows a datetime widget, and should probably only show a date widget.

Ideally there should also be some good test coverage for the new field to provide some basic confidence that it works correctly.

I am willing to support this addition in principle, but it is not ready in this current state.

---

image

image

samuelhwilliams:
I'm going to close this for now, as it doesn't work and is a little old now. There's also some questions around how supportable/maintainable the flask-mongoengine integration is, as it's not really supported any more and that will start to cause issue soon.

Feel free to open again if you want to raise this again.

Issue #2435, opened 2024-07-03
Mirror Pallets-eco setup for projects


Abdur-rahmaanJ:
@samuelhwilliams working on it.

samuelhwilliams:
I'm starting work on this in a branch/PR on my fork: https://github.com/samuelhwilliams/flask-admin/pull/1.

Going to just dump some notes here as I work through it - please bear with as a bunch of this will be stuff I haven't touched before.

I'm using `Flask-SQLAlchemy` as the reference project in terms of files/structure to copy, etc.

* I've migrated `requirements-dev.txt` to `requirements/tests.in`, and not pip-compiled that to a `tests.txt` because I think each Python version tox runs the tests in needs to pick appropriate versions internally - there's not a single set of dependencies that I think we can freeze that would work for all python versions we're matrix testing.
* `flask-sqlalchemy` has a `docs` directory, `flask-admin` has a `doc` directory. I'm not migrating these files because it's a huge changeset. We can do it in a separate PR if we thing it's worth doing.
  * The docs also need updating to the Pallets theme; possibly some fairly chunky changes needed to do this? I think it's best to do this separately.
* `flask-sqlalchemy` has a `main` branch; `flask-admin` is still using a `master` branch.
* I've copied across the `typing` target from `flask-sqlalchemy` but I don't imagine it will pass because it doesn't appear there's been typing tests on the current repo.
* `flask-sqlalchemy` has a `py-min` tox target and some `requirements-skip` files; do we need similar here? Not a thing I've seen elsewhere/am familiar with.

Abdur-rahmaanJ:
@samuelhwilliams Thanks for taking up the PR.

1. Please branch in here to get feedback
2. Please use this as the reference https://github.com/pallets-eco/flask-debugtoolbar

samuelhwilliams:
I don't have write access to this repo - I can't make a branch on here?

Abdur-rahmaanJ:
@samuelhwilliams Oh just a normal WIP PR, thanks!

PR #2436, opened 2024-07-05
Standardise project structure for the Pallets Ecosystem

samuelhwilliams:
Copies across, and then customises, the Pallets Ecosystem standard project structure from [Flask-DebugToolbar](https://github.com/pallets-eco/flask-debugtoolbar)

The tests throw a number of warnings, which are now being flagged as errors because of the standard project config. I've suppressed a number of these where it feels appropriate, but probably worth discussing if it's right to suppress them here or if we should immediately fix them. I suspect a number of them should be addressed in separate issues, eg things around DeprecationWarnings.

The test requirements file is a lot more complex for this project than for FDT, so I'm not sure we can freeze it in the same way. I have kept a `tests.in` file, with the existing `test_requires` from the original `setup.py`; not sure what we would want to do about this.

I suspect we may also need to disable the `typing` check and address that in a separate PR.

fixes: https://github.com/pallets-eco/flask-admin/issues/2435


Abdur-rahmaanJ (approved):
LGTM. Now, once merged, we need to fix CI/CD issues!

@aenglander

aenglander:
> LGTM. Now, once merged, we need to fix CI/CD issues!
> 
> @aenglander

The only thing that seems to need fixing is typing. If the library isn't ready for typing checks, the test should be disabled and a ticket created to add type support.

Issue #2440, opened 2024-07-08
Add typing support to project


Abdur-rahmaanJ:
1. Run mypy without strict flag, fix issues
2. Run mypy with strict flag, fix issues
3. Enable typing in tests once again

samuelhwilliams:
To do list:

 - [x] [Re-enable the `typing` github action](https://github.com/pallets-eco/flask-admin/pull/2445)
 - [x] Re-enable `warn_unused_configs` in pyproject.toml
 - [x] Re-enable `warn_redundant_casts` in pyproject.toml
 - [x] Re-enable `warn_unused_ignores` in pyproject.toml
 - [x] Re-enable `strict_equality` in pyproject.toml
 - [x] Re-enable `strict_concatenate` in pyproject.toml
 - [x] Re-enable `check_untyped_defs` in pyproject.toml
 - [x] Re-enable `disallow_subclassing_any` in pyproject.toml
 - [x] Re-enable `disallow_untyped_decorators` in pyproject.toml
 - [x] Re-enable `disallow_any_generics` in pyproject.toml
 - [x] Re-enable `disallow_untyped_calls` in pyproject.toml
 - [x] Re-enable `disallow_incomplete_defs` in pyproject.toml
 - [x] Re-enable `disallow_untyped_defs` in pyproject.toml
 - [ ] Re-enable `no_implicit_reexport` in pyproject.toml
 - [ ] Re-enable `warn_return_any` in pyproject.toml
 - [ ] Resolve/remove any `type: ignore` comments
 - [ ] Resolve/remove any `[[tool.mypy.override]]` blocks in pyproject.toml

ElLorans:
Does https://github.com/pallets-eco/flask-admin/pull/2608 unblock something?

PR #2445, opened 2024-07-13
Re-enable the `typing` github action

samuelhwilliams:
Related discussion/stream of work: https://github.com/pallets-eco/flask-admin/issues/2440

This turns back on the `typing` check, with all of the '--strict' options disabled, and then makes the minimum set of code changes required to make the typing tests pass.

Future patches will slowly work to re-enable each of the disabled mypy checks so that the burden of typing the codebase is a bit less overwhelming.

Also fixes running the test suite on macos with libgeos by pulling in an updated version of geoalchemy2.


samuelhwilliams:
Going to move forward with this just to unblock other work on flask-admin, but happy to hear thoughts retrospectively - especially if anyone catches anything I've missed here.

PR #2450, opened 2024-07-15
Update README.md: Little fixes, better grammar, syntax and look.

princerb:
Fixed some markdown mistakes. Added more formal language for the project originality, and other improvements.


README.md
```diff
@@ -50,41 +50,41 @@ add your own, or improve on the existing examples, and submit a
...
-    1. Clone the repository::
+        git clone https://github.com/pallets-eco/flask-admin.git
+        cd flask-admin
 
-          git clone https://github.com/pallets-eco/flask-admin.git
-          cd flask-admin
+2. Create and activate a virtual environment:
+        
+        virtualenv env -p python3
+        source env/Scripts/activate
```
samuelhwilliams, line 61: I think we should probably recommend `python3 -m venv venv`?

```suggestion
        python3 -m venv venv
        source venv/Scripts/activate  # Windows
        source venv/bin/activate  # unix
```

princerb: I agree, so does PSF ).

samuelhwilliams: thanks! 🙏

README.md
```diff
@@ -50,41 +50,54 @@ add your own, or improve on the existing examples, and submit a
...
+    pip install -r examples/sqla/requirements.txt
+    ```
+4. Run the application:
+
+    ```bash
+    # Windows:
+    python examples/sqla/run_server.py
+
+    # Linux:
+    python3 examples/sqla/run_server.py
```
samuelhwilliams, line 82:
```suggestion
    python examples/sqla/run_server.py
```

We don't really need to distinguish here.

README.md
```diff
@@ -50,41 +50,54 @@ add your own, or improve on the existing examples, and submit a
...
 ## Documentation
 
 Flask-Admin is extensively documented, you can find all of the
 documentation at <https://flask-admin.readthedocs.io/en/latest/>.
 
 The docs are auto-generated from the *.rst* files in the */doc* folder.
-So if you come across any errors, or if you think of anything else that
-should be included, then please make the changes and submit them as a
-*pull-request*.
+if you come across any errors or if you think of anything else that
```
samuelhwilliams, line 92:
```suggestion
If you come across any errors or if you think of anything else that
```

README.md
```diff
@@ -116,29 +129,31 @@ You should see output similar to:
...
 If you\'re using Homebrew on MacOS, you might need this:
 
-    # install postgis and geos
-    > brew install postgis
-    > brew install geos
+```bash
+# Install postgis and geos
+brew install postgis
+brew install geos
+export DYLD_LIBRARY_PATH=/opt/homebrew/opt/geos/lib/
```
samuelhwilliams, line 151: We shouldn't need this any more - we were using an older version of geoalchemy, which didn't check this directory. Now it does. Should all of the DYLD_LIBRARY_PATH stuff should be able to go (it's also specific to just darwin/macos).

Issue #2423, opened 2024-04-03
Can't use multiple `Admin` instances with different Bootstrap themes (`FLASK_ADMIN_SWATCH`)


princerb:
❓  How can we use different Bootstrap themes through multiple `Admin` instances?

👨‍💻  This is how I tried to do it:
```python
...
app.config['FLASK_ADMIN_SWATCH'] = 'cerulean'
admin1 = Admin(app, endpoint='admin1')
app.config['FLASK_ADMIN_SWATCH'] = 'cosmo'
admin2 = Admin(app, endpoint='admin2')
...
```
❌  But this does NOT work, I wrote this to clarify what I want to do with my issue.
_Thanks!_

princerb:
If I solve, will PRs be viewed/merged soon?

samuelhwilliams:
This feels like a useful feature. I'm not sure yet how easy or hard it would be to support this, but if you've got ideas I'd be happy to hear them and would be willing to help review any PR you do put up. Would be good to understand your proposed approach first.

samuelhwilliams:
I had a look at this and it might be something we could get into the v2 release. Here's a possible approach: https://github.com/pallets-eco/flask-admin/pull/2462

As I've coded it, it's a breaking change. We could make this a bit softer and wrap the breaking change if we really wanted to - although since we're planning a v2.0.0 release, now's a fair time to get it in.

princerb:
I actually already made the reslover changes for this, just needed to write the end of the example ) [the button that switches even the current theme]

princerb:
@samuelhwilliams as I see you created a complete new configuration class for the Bootstrap options, while I used a new param to `Admin` which is `bootswatch_theme`. I think your approach also is awesome 👍

samuelhwilliams:
I'd love to see what you've done so far or what you're planning - if you've covered some use cases that I haven't, we can take the best of both :)

princerb:
#2463

princerb:
But mine is quite simple one. Your approach gives the project a seperate and full control over the theming. But there is a catch, you added also `folder` param, are we also going to plan to use different Bootstrap versions along along the different `Admin` instances?

samuelhwilliams:
> But there is a catch, you added also folder param, are we also going to plan to use different Bootstrap versions along along the different Admin instances?

[...]

PR #2444, opened 2024-07-10
Initial attempt for Tabler Template Mode

hasansezertasan:
Tries to implement #2443.

Fixes #2443 

It's still a draft...

PR #2462, opened 2024-07-19
Pull together all Flask-Admin theme configuration

samuelhwilliams:
fix: https://github.com/pallets-eco/flask-admin/issues/2423

### DIFF
--- a/flask_admin/theme.py
+++ b/flask_admin/theme.py
@@ -1,0 +1,20 @@
+import typing
+from dataclasses import dataclass
+from functools import partial
+
+
+@dataclass
+class Theme:
+    folder: str  # The templates folder name to use
+
+
+@dataclass
+class BootstrapTheme(Theme):
+    folder: typing.Literal['bootstrap2', 'bootstrap3', 'bootstrap4']
+    swatch: str = 'default'
+    fluid: bool = False
+
+
+Bootstrap2Theme = partial(BootstrapTheme, folder='bootstrap2')
+Bootstrap3Theme = partial(BootstrapTheme, folder='bootstrap3')
+Bootstrap4Theme = partial(BootstrapTheme, folder='bootstrap4')
--- a/flask_admin/theme.py
+++ b/flask_admin/theme.py
@@ -1,16 +1,18 @@
-import typing
 from dataclasses import dataclass
+import typing
 from functools import partial
 
 
 @dataclass
 class Theme:
     folder: str  # The templates folder name to use
+    base_template: str
 
 
 @dataclass
 class BootstrapTheme(Theme):
     folder: typing.Literal['bootstrap2', 'bootstrap3', 'bootstrap4']
+    base_template: str = 'admin/base.html'
     swatch: str = 'default'
     fluid: bool = False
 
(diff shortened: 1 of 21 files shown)


princerb:
Will you plan to expect the new theming frameworks? Not just Bootstrap. Like adding Tailwind, Materialize etc?

samuelhwilliams:
I don't necessarily expect them - but this would support them fairly easily and allow end users to provide whatever configuration they want for their template mode by subclassing TemplateMode with additional attributes.

princerb:
Would it be better if we create a seperate module for their understanding then? The usage would look like this:
```python
from flask_admin.theming import Bootstrap5Theme, TailwindTheme

admin = Admin(..., theme=TailwindTheme(**parameters))
```

samuelhwilliams:
Yes, that feels like a good shout 👌

hasansezertasan:
I like how the things are going 🤓. I might have few ideas about this:

First of all, I think the base class should not have a swatch attribute because future implementations (etc. Tabler #2444) might not have that option.

I will get back with more if/when I get chance to review 🧐.

Great ideas 🚀.

samuelhwilliams:
The base class doesn't have a swatch attribute - just a mode :)

flask_admin/theme.py
```diff
@@ -0,0 +1,20 @@
+import typing
+from dataclasses import dataclass
+from functools import partial
+
+
+@dataclass
+class Theme:
```
princerb, line 7: BaseTheme ?

samuelhwilliams: I think just `Theme` is OK - it could actually be used directly rather than just as a base class. If we had a Tabler theme that needs no config, I think it would read better as `Admin(theme=Theme('tabler'))` rather than `Admin(theme=BaseTheme('tabler'))`?

hasansezertasan:
> The base class doesn't have a swatch attribute - just a folder 🙂

Oh sorry, I was mistaken. I had my phone while looking at the change you made and ı probably didn't notice it 😅.

> Tablet, Tailwind etc should inherit from the base 'Theme'

You are right 🧐.

samuelhwilliams:
Do we think `base_template` should also be part of the theme? I feel like ... probably yes 🤔

flask_admin/theme.py
princerb: I thought even that Tabler theme should inherit Theme, and do it like this:
```python
Admin(theme=TablerTheme())
```

princerb: We announce the new themes through the changelogs, so, when importing from the module `theme`, they are able to get all themes by knowing with the class names like (not Theme or BaseTheme at all):
```python
from flask_admin.theme import Bootstrap5Theme, TablerTheme, TailwindTheme, TailwindTheme, TablerWithBootstrap4Theme
```
This was the way I thought this theme feature would look like.
But the current implementation is also awesome.

princerb: The disadvantage of this, of course but it requires more coding.

princerb: Maybe considering both methods like `Theme('ThemeName', **params)` and `ThemeName(**params)`.
It gives us some complexity, but I haven't really analyzed how convenient this can be for the users.

PR #2467, opened 2024-07-21
Compatibility with Flask 3, WTForms 3, SQLAlchemy 2.0 and others

cjmayo:
Updated version of #2328.

A problem with monogengine that is skipped here- use flask-mongoengine2?

Obviously there is quite a lot here. What is the best way forward?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,9 @@ geoalchemy = [
 mongoengine = [  # TODO: seems out-of-date/unmaintained; replace or deprecate?
     "Flask-Admin[sqlalchemy]",
     "flask-mongoengine<1",
-    "Flask<2.3.0",  # flask-mongoengine tries to access `flask.json`
+    "Flask<2.3.0",  # flask-mongoengine tries to access `flask.json`,
+    "sqlalchemy>=1.4,<2",
+    "wtforms>=2,<3",
 ]
 pymongo = ["pymongo>=3.7.0"]
 peewee = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,8 +22,8 @@ classifiers = [
 ]
 requires-python = ">=3.8"
 dependencies = [
-    "flask>=2.0,<3",
-    "wtforms>=2,<3",
+    "flask>=2.0",
+    "wtforms>=2",
 ]
 
 [project.optional-dependencies]
--- a/flask_admin/contrib/sqla/widgets.py
+++ b/flask_admin/contrib/sqla/widgets.py
@@ -20,9 +20,13 @@ class CheckboxListInput:
 
     def __call__(self, field, **kwargs):
         items = []
-        for val, label, selected in field.iter_choices():
+        for field_choices in field.iter_choices():
+            if len(field_choices) == 3:  # wtforms <3.1, >=3.1.1, <3.2
+                value, label, selected = field_choices
+            else:
+                value, label, selected, _ = field_choices
             args = {
-                'id': val,
+                'id': value,
                 'name': field.name,
                 'label': escape(label),
                 'selected': ' checked' if selected else '',
--- a/flask_admin/model/widgets.py
+++ b/flask_admin/model/widgets.py
@@ -158,7 +158,11 @@ class XEditableWidget(object):
 
             choices = []
             selected_ids = []
-            for value, label, selected in field.iter_choices():
+            for field_choices in field.iter_choices():
+                if len(field_choices) == 3:  # wtforms <3.1, >=3.1.1, <3.2
+                    value, label, selected = field_choices
+                else:
+                    value, label, selected, _ = field_choices
                 try:
                     label = text_type(label)
                 except TypeError:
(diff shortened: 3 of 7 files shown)


samuelhwilliams:
Yeah I saw the flask-mongoengine issue recently too. Unsure whether we'd want flask-mongoengine2 or flask-mongoengine3. The latter looks a 'bit' more popular, and keeps wtforms support (unsure right now if that's relevant to us - maybe?)

Both do seem to have very little in the way of support/activity, though, so it makes me nervous to switch over to them.

flask_admin/tests/mongoengine/test_basic.py
```diff
@@ -132,8 +132,8 @@ def test_model(app, db, admin):
     model = Model1.objects.first()
     assert model.test1 == 'test1small'
     assert model.test2 == 'test2large'
-    assert model.test3 == ''
-    assert model.test4 == ''
+    assert model.test3 in ('', None)  # WTForms 2, WTForms 3
+    assert model.test4 in ('', None)  # WTForms 2, WTForms 3
```
samuelhwilliams, line 136: This should just be fine as `is None` now, I think

samuelhwilliams:
I do agree that getting flask-admin into a place where we can officially support the latest version of these key dependencies is a priority. I think Flask v3+ and SQLAlchemy v2+ are big ones.

I _think_ wtforms3 already works - unsure if there are specific issues you've seen that I've maybe missed so far, though?

I think we probably to adjust how we are pinning dependencies for tests, to get a bit more confidence that combinations are working. I have been very vaguely thinking about this in a wip/provisional commit on my fork here: https://github.com/pallets-eco/flask-admin/commit/ff228639ee55e5cc7f0f37e1ddfc655400468caf. But I still have a lot of musings/unanswered questions.

I feel like we maybe next a few sets of `tests.in` requirements file that pin different combinations of flask/wtforms/sqlalchemy etc.

Maybe on `tests-flask2.in` file that pins to flask v2, wtforms v2, sqlalchemy v1 and another `tests-flask3.in` that pins to flask v3, wtforms v3, sqlalchemy v2, etc?

I'm unsure the current [`tox.ini` config](https://github.com/pallets-eco/flask-admin/blob/master/tox.ini#L21-L24) is really 'good enough'; maybe separate test requirements files is a bit clearer?

samuelhwilliams:
I am probably ok with us dropping wtforms2 - and maybe for v2.0.0 makes sense since it's a breaking change anyway? If so we should aim to get that in sooner rather than later?

samuelhwilliams:
Breaking it down a little bit would probably be helpful, yes 👍 

1 PR for each dependency would probably be OK/a good level? Flask3, WTForms3, SQLAlchemy2 - unless you think some other split would be good.

samuelhwilliams:
I'm undecided about `wtf-peewee`.

Part of me thinks that we should require `wtf-peewee>=3.0.5`, which would then implicitly require `wtforms>=3.1.0` as you suggest. That would mean people using Flask-Admin for peewee would have to use wtforms 3.1.0+. But I think that's better than limiting users to `wtf-peewee==3.0.4` and then implicitly `wtforms<3.1.0` - ie forcing them onto older dependencies.

I think neither situation is perfect, but I would tend towards wanting people to be on newer versions of things than forcing them to be on older ones. What do you think?

(flask-)mongoengine is a whole other beast. I am reluctant to remove it altogether, but I don't know what approach we _do_ want to take with supporting a dependency that isn't compatible with the latest versions of flask and probably isn't going to become compatible due to lack of maintenance. I feel like we probably need to cut that loose ... but I'd be very happy to hear an alternative proposal.

samuelhwilliams:
> Should the filters go in like this or do they now belong in pyproject.toml?

I think I'm OK with those filters going in where you've put them. I've just noticed that we no longer need the "Please update your type formatter" filterwarnings - I fixed that issue a while back but clearly missed those.

If you want to strip those filters out in this PR, that'd be great. Otherwise I'll remove them separately after this is in.

samuelhwilliams:
I think we need to change our use of `field.iter_choices()` to support both 3-tuples and 4-tuples, then we should be able to unpin wtf-peewee?

In both `flask_admin.models.widgets` and `flask_admin.contrib.sqla.widgets`:

```
for value, label, selected in field.iter_choices()
```

to

```
for field_choices in field.iter_choices():
    if len(field_choices) == 3:  # wtforms <3.1, >=3.1.1, <3.2
        value, label, selected = field_choices
    else:
        value, label, selected, _ = field_choices
```

pyproject.toml
```diff
@@ -47,7 +47,9 @@ geoalchemy = [
 mongoengine = [  # TODO: seems out-of-date/unmaintained; replace or deprecate?
     "Flask-Admin[sqlalchemy]",
     "flask-mongoengine<1",
-    "Flask<2.3.0",  # flask-mongoengine tries to access `flask.json`
+    "Flask<2.3.0",  # flask-mongoengine tries to access `flask.json`,
+    "sqlalchemy>=1.4,<2",
```
samuelhwilliams, line 51: I don't think we need to repeat the `>=1.4` here, as we get this from the `sqlalchemy` optional group already.

However it's all a bit moot, as I have a PR up to remove support for `mongoengine` altogether.

samuelhwilliams (approved):

Issue #2351, opened 2023-04-23
List view shows each map twice


TomGoBravo:
Each time `applyGlobalStyles` is called a map div is appended to the DOM after the textarea. I added logging in processLeafletWidget and see that there are two calls:

In form.js:
```
    // Apply global styles for current page after page loaded
    $(function() {
        faForm.applyGlobalStyles(document);
    });
```

and bs4_modal.js
```
$(function() {
  // Apply flask-admin form styles after the modal is loaded
  window.faForm.applyGlobalStyles(document);
});
```

I don't know if both calls to applyGlobalStyles are needed. A hack fix is exiting processLeafletWidget if the textarea is already hidden.

TomGoBravo:
@mrjoes : please take a look at my proposed fix (or workaround?) https://github.com/flask-admin/flask-admin/pull/2354

PR #2407, opened 2023-12-19
Fixes #2406: Multiple requests are sent when opening create, edit, and detail modals due to repetitive load of `bs4_modal.js`

irtimir:
Script tags that load `bs4_modal.js` in `templates/bootstrap4/admin/model/modals/create.html`, `templates/bootstrap4/admin/model/modals/edit.html`, and `templates/bootstrap4/admin/model/modals/details.html` template files are removed.

Fix: https://github.com/flask-admin/flask-admin/issues/2406

PR #2468, opened 2024-07-21
Fix modal regression

samuelhwilliams:
In https://github.com/pallets-eco/flask-admin/pull/2407 we removed some seemingly-duplicate network requests for
`bs4_modal.js` every time a modal window is loaded.

This would have been fine if that file did not include any
side-effects, but it did: it re-applied 'global styling' every time the
JS file is loaded.

The goal here was to process 'styling' for any dynamic/interactive
elements, such as map widgets, and fields that use 'select2' or
'x-editable'.

When we removed these 'extra' JS calls, the modal windows stopped
getting these interactive elements enabled/setup.

This patch restores that functionality, while still retaining the goal
of PR https://github.com/pallets-eco/flask-admin/pull/2407: to prevent duplicate network requests for a file.

We load `bs<x>_modal.js` on all pages that can display modals; this file
now only sets up the event listener that is responsible for making sure
a modal is cleared out/refreshed between multiple modal displays.

The logic for applying styling to the window is now added inline to all
of the relevant modal components.

A further improvement on what we had before is that now, the call to add
styling is scoped specifically to the modal. This prevents the issue
described in https://github.com/pallets-eco/flask-admin/issues/2351, where map elements may show up multiple times on
the page because `applyGlobalStyles` is called multiple times.

fixes: #2351 

---

## After this PR

---

## To test manually:

* Run `python examples/geo_alchemy/app.py`
* Add some points using the create menu. Maps may be blank unless you've setup mapbox, but you can still save points.
* Go to the list view and load/leave the edit modal a few times.
* The edit modal should show the map widget, and the list view should only show one map instance throughout.

Repeat this on the `master` branch and you'll see the modal view is broken.

Do `git revert 2fbdabce4c19db0eb144b0aa5049eb0ae7c4db37` and `git revert 7df333629faa6baa56d409c507a37882a22c40e9` to manually unwind PR #2407. You'll see the modal view works, but the list view shows more and more elements because global styling is reapplied at the global scope multiple times.

### DIFF
--- a/examples/geo_alchemy/app.py
+++ b/examples/geo_alchemy/app.py
@@ -60,13 +60,17 @@ def index():
 # Create admin
 admin = admin.Admin(app, name='Example: GeoAlchemy', theme=Bootstrap4Theme())
 
+
+class ModalModelView(ModelView):
+    edit_modal = True
+
 # Add views
-admin.add_view(ModelView(Point, db.session, category='Points'))
-admin.add_view(ModelView(MultiPoint, db.session, category='Points'))
-admin.add_view(ModelView(Polygon, db.session, category='Polygons'))
-admin.add_view(ModelView(MultiPolygon, db.session, category='Polygons'))
-admin.add_view(ModelView(LineString, db.session, category='Lines'))
-admin.add_view(ModelView(MultiLineString, db.session, category='Lines'))
+admin.add_view(ModalModelView(Point, db.session, category='Points'))
+admin.add_view(ModalModelView(MultiPoint, db.session, category='Points'))
+admin.add_view(ModalModelView(Polygon, db.session, category='Polygons'))
+admin.add_view(ModalModelView(MultiPolygon, db.session, category='Polygons'))
+admin.add_view(ModalModelView(LineString, db.session, category='Lines'))
+admin.add_view(ModalModelView(MultiLineString, db.session, category='Lines'))
 
 if __name__ == '__main__':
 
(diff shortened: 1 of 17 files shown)


samuelhwilliams:
In the long run we probably want some tests for this. But given it's all interactive/JS-based stuff, that probably means Playwright or similar, and so I'm leaving out of scope for now.

But it's a clear gap in our testing and we should fix it. Will raise an issue...

hasansezertasan:
> In the long run we probably want some tests for this. But given it's all interactive/JS-based stuff, that probably means Playwright or similar, and so I'm leaving out of scope for now.
> 
> But it's a clear gap in our testing and we should fix it. Will raise an issue...

I totally agree with you. I believe playwright should fit our purpose.

Tomorrow, I'll take a look at the changes you made and leave a review.

hasansezertasan (approved):
Since it's not a problem I've encountered before, it seems like I missed it in #2407 😞. There are a lot of moving parts in the frontend code and I have to admit we need to be more careful on these PRs 🤓.

I reviewed the changes you made and everything looks fine to me 🚀.

I also tried to run `examples/geo_alchemy/app.py` but failed. I can't seem to run the "Setup the database" section properly 🦾.

But I do trust the logic here 🥇. In any case, one more approval might be better.

hasansezertasan:
> I also tried to run `examples/geo_alchemy/app.py` but failed. I can't seem to run the "Setup the database" section properly 🦾.

I don't have to test it with the `geo_alchemy` example anymore. I tested it with the `sqla-association_proxy` and witnessed the problem.

#2407, prevents the `flask_admin/static/admin/js/form.js` functionality on modals, causing JS initialized form fields to break - not initialize at all.

I believe we'll need a browser-based testing solution for such problems.

PR #2474, opened 2024-07-23
Remove bootstrap 2 and 3

samuelhwilliams:
Removes Bootstrap2 and Bootstrap3 themes, both of which are now comfortably replaced by Bootstrap4 (and, really, 5 when we add it).

We want to support add support for Bootstrap5 soon. This will probably be easier without the thought of having to support anything from bs2 and bs3.


hasansezertasan:
> We want to support add support for Bootstrap5 soon. This will probably be easier without the thought of having to support anything from bs2 and bs3.

Yes and no, X-Editable doesn't work with BS5 😞.

flask_admin/theme.py
```diff
@@ -11,12 +11,10 @@ class Theme:
...
-    folder: typing.Literal['bootstrap2', 'bootstrap3', 'bootstrap4']
+    folder: typing.Literal['bootstrap4']
     base_template: str = 'admin/base.html'
     swatch: str = 'default'
     fluid: bool = False
 
 
-Bootstrap2Theme = partial(BootstrapTheme, folder='bootstrap2')
-Bootstrap3Theme = partial(BootstrapTheme, folder='bootstrap3')
 Bootstrap4Theme = partial(BootstrapTheme, folder='bootstrap4')
```
hasansezertasan, line 20: What about declaring Bootstrap4Theme without partial?

```suggestion
class Bootstrap4Theme(BootstrapTheme):
    folder='bootstrap4'
```

hasansezertasan (approved):
Looks good to me, good job 👍💯👍.

flask_admin/theme.py
samuelhwilliams: I think, because `BootstrapTheme` is a dataclass, it's preferable to me to just instantiate it than subclass it :)

samuelhwilliams:
Ugh, deleted my fork unintentionally 🙃  I will re-open and merge, referencing this.

PR #2483, opened 2024-07-26
SQLAlchemy 2.0 Compatibility

cjmayo:
Here are the SQLAlchemy 2.0 from #2467, which will pass here against 1.4...

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -26,7 +26,11 @@ jobs:
           - python: '3.12'
             tox: 'py312-noflaskbabel'
           - python: '3.12'
-            tox: 'py312-flaskmongoengine'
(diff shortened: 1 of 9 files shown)


samuelhwilliams:
This would be great to get in 👍 

Could we also please:

* Remove the sqlalchemy warnings filtered out in `pyproject.toml`
* Remove the `<2` limitation in the `sqlalchemy` extra in `pyproject.toml`
* Add a test run for py38-sqlalchemy1 and py312-sqlalchemy1 via tox, since after those changes above we'll be running against sqlalchemy >=2.0.0 by default

samuelhwilliams (approved):

samuelhwilliams:
Great work, thank you :)

PR #2484, opened 2024-07-28
Fix test_different_bind_joins UnboundExecutionError

cjmayo:
sqlalchemy.exc.UnboundExecutionError: Bind key 'other' is not in 'SQLALCHEMY_BINDS' config.

---

An update of a commit from #2328.

I'm no fixtures expert so maybe there is a better way, although this does seem to be a one-off at this time.

### DIFF
--- a/flask_admin/tests/sqla/test_basic.py
+++ b/flask_admin/tests/sqla/test_basic.py
@@ -2637,12 +2637,11 @@ def test_multipath_joins(app, db, admin):
         assert rv.status_code == 200
 
 
-# TODO: Why this fails?
-@pytest.mark.xfail(raises=Exception)
-def test_different_bind_joins(app, db, admin):
-    app.config['SQLALCHEMY_BINDS'] = {
-        'other': 'sqlite:///'
-    }
+def test_different_bind_joins(request, app):
+    app.config['SQLALCHEMY_BINDS'] = {'other': 'sqlite:///'}
+
+    db = request.getfixturevalue('db')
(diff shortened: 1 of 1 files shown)


flask_admin/tests/sqla/test_basic.py
```diff
@@ -2637,12 +2637,16 @@ class Model2(db.Model):
...
+def test_different_bind_joins(app, babel):
+    from flask_admin import Admin
+    from flask_sqlalchemy import SQLAlchemy
+
+    app.config['SQLALCHEMY_DATABASE_URI'] = 'sqlite:///'
+    app.config['SQLALCHEMY_ECHO'] = True
+    app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = False
+    app.config['SQLALCHEMY_BINDS'] = {'other': 'sqlite:///'}
+    db = SQLAlchemy(app)
+    admin = Admin(app)
```
samuelhwilliams, line 2649:
```suggestion
def test_different_bind_joins(request, app):
    app.config['SQLALCHEMY_BINDS'] = {'other': 'sqlite:///'}

    db = request.getfixturevalue('db')
    admin = request.getfixturevalue('admin')
```

This achieves a very similar thing, but avoids having to duplicate the other fixture setup and keeps consistent with how we pull in these fixtures dynamically in a few other places where we need some custom config.

samuelhwilliams (approved):
Thanks! 🙏

PR #2485, opened 2024-07-28
Fix KeyError exception when flask-babel is installed but not used

cjmayo:
flask-babel's gettext() expects a Babel object to have been created.

Otherwise an exception is raised.

```
  File "/usr/lib/python3.12/site-packages/flask_babel/__init__.py", line 46, in get_babel
    return app.extensions['babel']
           ~~~~~~~~~~~~~~^^^^^^^^^
KeyError: 'babel'
```

---

Not completely convinced by this, but at least it illustrates the issue.

### DIFF
--- a/flask_admin/babel.py
+++ b/flask_admin/babel.py
@@ -1,7 +1,14 @@
+from flask import current_app, has_app_context
+
 try:
     from flask_babel import Domain
 
-except ImportError:
+    if has_app_context():
+        current_app().extensions['babel']
+    else:
+        raise KeyError
+
+except (KeyError, ImportError):
     def gettext(string, **variables):
         return string % variables
 


samuelhwilliams:
Yeah I've been chewing over this one for a little while in the back of my head. It's a bit gnarly because it mutates the entire `flask_admin.babel` module based on what's in place at the time of the first import.

I think it would be better to refactor this so that the state of the flask app, flask_admin instance, etc aren't relevant at module import time. But haven't yet dug into this enough to work out what that work look like and/or how far the repercussions would spread.

ElLorans:
This would also fix https://github.com/pallets-eco/flask-admin/issues/2609 . I would try to do what flask-security did https://github.com/pallets-eco/flask-security/blob/main/flask_security/babel.py by adding a `has_babel_ext()` function to prevent the KeyError.

ElLorans:
This was solved with https://github.com/pallets-eco/flask-admin/pull/2624 by running `uv run main.py` to launch the examples, so I would close this PR for now.

Issue #2466, opened 2024-07-21
Remove (flask-)mongoengine support?


samuelhwilliams:
[`flask-mongoengine`](https://github.com/MongoEngine/flask-mongoengine) appears to be dead, and isn't compatible with Flask>=2.3.0.

Should we stop supporting it? What're the alternatives? Is `pymongo` enough?

princerb:
**The dead frameworks should be removed. But I guess `flask-mongoengine` which uses `MongoEngine `is high-level abstraction and implemented with a lot of features. If we only stay with `PyMongo`, it is perfect itself. How about we remove it in the next release?**

Until then, currently, we add a warning message to all `MongoEngine` users to tell them that *we are dropping support for FLask-MongoEngine usage from x.x.x version, and please migrate to mongo or stay with the old version of flask-admin*?

But I also want to hear others' opinions about this)

samuelhwilliams:
I've gone back and forth on this a little bit.

Given `flask-mongoengine` appears to be unsupported/unmaintained, is incompatible with newer versions of our core dependencies (eg flask/wtforms), and has no obvious popular replacement (there are a couple, but much smaller community presence), I propose that we remove `flask-mongoengine` support from Flask-Admin in v2.0.0 onwards.

If `flask-mongoengine` becomes maintained again, or a fork picks up a clear broad base of support and is brought up-to-date, then I'd be glad and keen to add support back in.

PR #2493, opened 2024-08-01
Remove (flask-)mongoengine support :(

samuelhwilliams:
Support for `mongoengine` relies heavily on `flask-mongoengine` to
provide wtforms integration. Unfortunately, `flask-mongoengine` has
not seen much activity of support over the last two years, and is no
longer compatible with recent versions of our core dependencies (eg
flask).

Given this, it feels inappropriate and undesirable for us to try to
continue to support the functionality that project brings, as it will
increasingly apply a higher maintenance burden on us and eventually
likely cause compatibility issues for Flask-Admin users generally (if
someone wants to use both sqlalchemy and mongoengine views in
Flask-Admin, they'd be forced onto the lowest common denominator
dependency versions).

---

closes: #2466


princerb:
We hope the updates will resume soon to reintegrate it.

By the way, do we need to add a warning message to those who accidently/certainly updated `flask-admin` intentionally while using `mongoengine` in the project? The warning could advise them to *either implement their own raw method or downgrade the version.*

samuelhwilliams:
I'm not sure a warning would be very useful - if we remove the code, and they are trying to import anything from`flask_admin.contrib.mongoengine`, they will get import errors. Not something we could easily throw a warning for at that point - it might just be more apparent what's happening for them to see the ImportError?

If we were doing another minor/patch version release before 2.0.0 it would make sense to add DeprecationWarnings, but since we aren't and the next release will be a major one, there may not be a good touchpoint.

PR #2496, opened 2024-08-05
Replace filter support for sqlalchemy_enum34 with sqlalchemy.Enum

cjmayo:
PEP-435-style enumerated classes supported by sqlalchemy.Enum since SQLAlchemy 1.1.

---

We have no tests using sqlalchemy_enum34. The last SQLAlchemy-Enum34 release was in 2021.

sqlalchemy_enum34 is no longer necessary in applications and appears to be straightforward to replace.


samuelhwilliams (approved):
Nice spot - this feels sensible 👍

Issue #1892, opened 2019-07-26
Setting timezone (again)


tmladek:
Hello,
Flask-Admin seems to show all `timestamp` fields as if they were UTC (which is also set in the locale of the server). However, I would like to show the dates in GMT+2, which is the timezone of all users of my app - without having to reconfigure the server it is running on.

I found #1246, which is now closed, referring to #772, which in fact discusses a different issue, and I have to concur with the last poster, that the solution there doesn't work - nothing at all changes in my forms if I override the widgets like described.

I also tried using Flask-BabelEx, setting the locale to 'cs' using `@babel.localeselector`, which successfully changed the form labels, but not the timezone - and changing the timezone directly, using `@babel.timezoneselector`, which also had no effect.

Is there a way to set Flask-Admin's timezone?

chakritp:
Does anyone have an update or workaround on this?

ElLorans:
> Does anyone have an update or workaround on this?

@hmanicka posted a [gist ](https://gist.github.com/hmanicka/82834f2e0df016947730dbaf19d25c28/revisions) 2 years ago, but I would rather let the client do at least the conversion to local time. My approach has 2 advantages: bringing some load to the client rather than the server, and automatically getting the user's timezone. Unfortunately, I am not able to make it work with datepicker. This is not an issue to me, as nowadays browser's UI for datetime-local is excellent.

`templates/admin/edit.html`

```
{% extends 'admin/model/edit.html' %}
{% block tail %}
    <script>
        fetch('/set_timezone', {
            method: 'POST',
            headers: {
                'Content-Type': 'application/json'
            },
            body: JSON.stringify(Intl.DateTimeFormat().resolvedOptions().timeZone)
        })

        function convertToUserLocalTime() {
            const inputs = document.querySelectorAll('input[data-date-format]');
            const userTimezone = Intl.DateTimeFormat().resolvedOptions().timeZone;
            inputs.forEach(input => {
                const localizedTime = new Date(input.getAttribute("value") + "Z");
                const formattedDate = `${localizedTime.getFullYear()}-${(localizedTime.getMonth() + 1)
                    .toString()
                    .padStart(2, '0')}-${localizedTime.getDate().toString().padStart(2, '0')}T${localizedTime
                    .getHours()
                    .toString()
                    .padStart(2, '0')}:${localizedTime.getMinutes().toString().padStart(2, '0')}`;
                input.value = formattedDate;
            });
        }

[...]

PR #2502, opened 2024-08-08
Example: display datetime fields in client's timezone and save user's input in UTC

ElLorans:
This example shows how to make Flask-Admin display all datetime fields in client's timezone.
Timezone conversion is handled by the frontend in /static/js/timezone.js, but an
automatic post request to /set_timezone is done so that flask session can store the
client's timezone and save datetime inputs in the correct timezone.

fixes: https://github.com/pallets-eco/flask-admin/issues/1892


examples/datetime-timezone/requirements.txt
```diff
@@ -0,0 +1 @@
+Flask-Admin
```
samuelhwilliams, line 1: The example also needs sqlalchemy and flask-sqlalchemy-lite installed, so can we change this please:

```suggestion
.[sqlalchemy]
flask-sqlalchemy-lite
```

It also raises a slightly interesting question. Installing `flask-admin[sqlalchemy]` will install `flask-sqlalchemy` as a dependency - but this uses `flask-sqlalchemy-lite` instead. From a quick look, I think us installing `flask-sqlalchemy` automatically is not actually required. If that's true, we should probably remove it from our dependencies. I'll need to check a bit closer though.

examples/datetime-timezone/app.py
```diff
@@ -0,0 +1,119 @@
...
+
+# Flask views
+@app.route('/')
+def index():
+    return '<a href="/admin/article">Click me to get to Admin!</a>'
+
+
+@app.route('/timezone.js/')
+def timezone_js():
+    return send_from_directory('static/js', 'timezone.js')
```
samuelhwilliams, line 91: We shouldn't need a custom handler for this if we configure Flask correctly. It should be something like changing the Flask instance creation to `app = Flask(__name__, static_folder='static', static_url_path='/static')` and then changing the `extra_js` line to `extra_js = ['/static/js/timezone.js']`

examples/datetime-timezone/requirements.txt
ElLorans: Thanks! Flask-Admin does not install flask-sqlalchemy as a dependency. Just tested with conda:
`conda create -n flask-admin-venv`
`conda install flask-admin`
`conda list`
gives
```

_libgcc_mutex
_openmp_mutex
blinker
bzip2
ca-certificates
click
flask
flask-admin
importlib-metadata
itsdangerous
jinja2
ld_impl_linux-64
libexpat
libffi
libgcc-ng
libgomp
libnsl
libsqlite
libuuid
libxcrypt
libzlib
markupsafe
ncurses
openssl
pip
python
python_abi
readline
setuptools
tk
tzdata
werkzeug
wheel
wtforms
xz
zipp
```

examples/datetime-timezone/app.py
ElLorans: You are correct: we can remove the route and changing the extra_js line to `extra_js = ['/static/js/timezone.js']`. I have also leveraged moment.js to remove some code from /static/js/timezone.js

examples/datetime-timezone/requirements.txt
ElLorans: I have added flask-sqlalchemy-lite as a dependency who has sqlalchemy as a required dependency. https://github.com/pallets-eco/flask-admin/pull/2502/commits/73cff4366a3434897634553e1a3c975b9ec4e81c

samuelhwilliams: Ah - currently it doesn't (v1.6.1), but as of v2.0.0 it's listed as a dependency, so would be installed automatically.

examples/datetime-timezone/app.py
samuelhwilliams: I think you've missed updating Flask - the example is broken now

ElLorans: are you running it locally? What error do you get?

examples/datetime-timezone/requirements.txt
ElLorans: > Ah - currently it doesn't (v1.6.1), but as of v2.0.0 it's listed as a dependency, so would be installed automatically.

Ah! I missed that. flask-sqlalchemy is used only in /flask_admin/tests, so I agree it should not be a dependency. https://github.com/pallets-eco/flask-sqlalchemy-lite is a very new project from july but is part of pallets-eco. Obviously I can use an except ImportError to use the more established flask-sqlalchemy instead

samuelhwilliams: My mistake - didn't spot that you'd added a separate TimezoneAware Article tab now.

examples/datetime-timezone/app.py
samuelhwilliams: My mistake - didn't spot that you'd added a separate TimezoneAware Article tab now.

ElLorans: Yes, I was trying to make it more clear. Maybe I should change the link at / to point to the TimezoneAware tab? Or should I simply drop the regular naive timezone one?

samuelhwilliams: I think it's fine to have separate tabs 👍  Maybe the column name for the TimezoneAware tab could be "Last edit (local time)" - or could display the timezone code?

ElLorans: I have added (UTC) and (local time) to the labels in https://github.com/pallets-eco/flask-admin/pull/2502/commits/1439622aae142f51a14f81f567bc32826fe5d476 (I have also changed the url to follow snake case: now it's http://127.0.0.1:5000/admin/timezone_aware_article/ and not http://127.0.0.1:5000/admin/timezoneaware_article/). I have chosen to display "(local time)" because it's not immediate to show the timezone code as session["timezone"] might not be set yet when the user visits the page, so we would need to change the label with js.

examples/datetime-timezone/requirements.txt
samuelhwilliams: Yeah, I've seen it mentioned a lot recently. We should probably investigate adding direct support for it (if we need to do anything).

I think that anyone using Flask-admin with SQLAlchemy will have been using Flask-SQLAlchemy. For now, could we keep the example consistent with the rest and just use Flask-SQLAlchemy instead of the new `-Lite` project?

Something for us/me to look into in the future around adding proper support/removing the Flask-SQLAlchemy dependency.

examples/datetime-timezone/app.py
samuelhwilliams: Looks good 👍

examples/datetime-timezone/requirements.txt
ElLorans: done in https://github.com/pallets-eco/flask-admin/pull/2502/commits/cf679e04684b6268e1d0fb863db30d4f4ec3ebcc .

>Yeah, I've seen it mentioned a lot recently. We should probably investigate adding direct support for it (if we need to do anything). [...] Something for us/me to look into in the future around adding proper support/removing the Flask-SQLAlchemy dependency.

I expect close to no work in Flask-Admin to support flask-sqlalchemy-lite as flask-sqlalchemy is imported only in 3 tests and pagination is handled using native sqlalchemy rather than flask-sqlalchemy one. I am trying out flask-sqlalchemy-lite and works fine with flask-admin.
For now, I would only advocate for dropping flask-sqlalchemy as a required flask-admin dependency since it's not required (e.g. pewee does not require flask-sqlalchemy and is supported by flask-admin).

samuelhwilliams (approved):
Thanks 🙏

PR #2506, opened 2024-08-26
Remove unnecessary execute bits

cjmayo:
This touches quite a lot of files, if there any clashes it is easy to regenerate.


samuelhwilliams (approved):
Looks fine, thanks.

PR #2507, opened 2024-08-26
Update references to the project URL

cjmayo:
This changes quite a lot of files, if there any clashes it is easy to regenerate.


flask_admin/translations/zh_Hans_CN/LC_MESSAGES/advanced.po
```diff
@@ -702,7 +702,7 @@ msgid ""
 "admin/flask-admin/tree/master/examples/peewee."
```
samuelhwilliams, line 702: Unfortunately the URL is split across lines here so would also need to update these - might need a bit of manual checking for any other instances 👀

samuelhwilliams (approved):
Thanks for this - very helpful 🙏

PR #2509, opened 2024-08-27
[pallets-eco#2505] fixing inconsistent modal headers

### LINKED ISSUE #2505: Inconsistent headers in modals
Bootstrap4 modals have a minor bug, which is still present in [the current version of the repo](https://github.com/pallets-eco/flask-admin/blob/a5c3baaa2b59d9face72ebd033f7562aebc7033c/flask_admin/templates/bootstrap4/admin/model/modals/details.html#L7). Namely, both `model/modals/edit.html` and `model/modals/create.html` contain the following code in `header_text` block:
```html
<h5 class="modal-title">{{ … }}</h5>
```
Meanwhile, **`model/modals/details.html`** contains the following instead:
```html
<h3>{{ … }}</h3>
```

…Which as you might guess results in different styling:
edit header
view header

LeXofLeviafan:
Fixes #2505:
* "View Record" modal header styling is no longer inconsistent with "Edit Record" & "Create New Record"

Before:
create
edit
view-h5
After:
view-h5
edit
create

### DIFF
--- a/flask_admin/templates/bootstrap4/admin/model/modals/details.html
+++ b/flask_admin/templates/bootstrap4/admin/model/modals/details.html
@@ -4,7 +4,7 @@
 {% block body %}
   <div class="modal-header">
     {% block header_text %}
-      <h3>{{ _gettext('View Record') + ' #' + request.args.get('id') }}</h3>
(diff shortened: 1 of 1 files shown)


samuelhwilliams:
Nice spot - thanks!

samuelhwilliams (approved):

PR #2516, opened 2024-09-03
Remove unused coveralls requirement

cjmayo:
Added in:
bb93ee7c ("Test python 3.5, and use coveralls and pylint.", 2016-03-10)

---

Looks like it was added for use with Travis. If we want to use a similar service on GitHub then there are Actions available.

### DIFF
--- a/requirements-skip/tests-min.in
+++ b/requirements-skip/tests-min.in
@@ -1,4 +1,3 @@
-coveralls
 flake8
 pylint
 pytest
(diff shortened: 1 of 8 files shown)


samuelhwilliams:
They need to be regenerated (pip-compiled) locally, it's not automatic

samuelhwilliams:
I think that's for a different purpose (but is also broken and needs fixing)

To pin new requirements - `pip-compile <file>` - which needs `pip-tools` available

samuelhwilliams (approved):

samuelhwilliams:
Somewhat expected - that tox command uses `change_dir` which was only added in tox v4+.

Possibly `dev.in` could be updated to clarify this.

PR #2518, opened 2024-09-04
Test without sqlalchemy-citext

cjmayo:
The last sqlalchemy-citext update was in 2021 and it has various compatibility issues.

SQLAlchemy now has CITEXT support.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,6 @@ sqlalchemy = [
 sqlalchemy-with-utils = [
     "Flask-Admin[sqlalchemy]",
     "sqlalchemy_utils>=0.38.0",
-    "sqlalchemy-citext>=1.8.0",
     "colour>=0.1.5",
     "email_validator>=2",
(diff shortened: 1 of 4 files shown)


samuelhwilliams:
This looks sensible. Would you mind adding an entry in `doc/changelog.rst` under a new v2.0.0a2 heading please?

pyproject.toml
```diff
@@ -36,7 +36,6 @@ sqlalchemy = [
 sqlalchemy-with-utils = [
     "Flask-Admin[sqlalchemy]",
     "sqlalchemy_utils>=0.38.0",
-    "sqlalchemy-citext>=1.8.0",
```
samuelhwilliams, line 39: I've had a bit more of a think about this since my last comment - sorry.

We still theoretically support `sqlalchemy<2.0.19` - this feels like it would break users on SQLAlchemy v1 and I'm not sure we've made the decision to drop support for that yet.

PR #2519, opened 2024-09-04
Filter unclosed database warning from sqlite3

cjmayo:
The Python 3.13 sqlite3 module added a new warning which is being issued during testing.
I do not know if there is an actual fix.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -107,6 +107,9 @@ filterwarnings = [
 
     # `flask.testing` accesses this attribute; remove when they have updated their code.
     "default:The '__version__' attribute is deprecated and will be removed in Werkzeug 3\\.1\\.:DeprecationWarning",
+
+    # Python 3.13 sqlite3 module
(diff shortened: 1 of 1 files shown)


samuelhwilliams:
Have you looked into anything that would fix them properly before we just ignore them?

I feel like we or our tests must be doing something wrong here.

samuelhwilliams:
I've seemingly fixed all but one of these errors with this changeset: https://github.com/samuelhwilliams/flask-admin/pull/24/commits/ce6cca543bf764f001770757b29828918f5330c8

The one outstanding failure seems to jump between different unit tests every now and again, which suggests something a little flaky/unstable. Haven't yet narrowed down what this is. I suspect our tests are not well isolated and/or our conftest setup is doing something wrong.

ElLorans:
Given https://github.com/pallets-eco/flask-sqlalchemy/issues/1379, I think we can filter out the warning

samuelhwilliams:
Fixed/superceded by https://github.com/pallets-eco/flask-admin/pull/2638 (sorry)

PR #2521, opened 2024-09-07
fix translations_path

LeXofLeviafan:
The `translations_path` constructor parameter appears to have _never_ actually worked with `flask-babel`, because `flask_babel.Domain` [**does not have**](https://github.com/python-babel/flask-babel/blob/v4.0.0/flask_babel/__init__.py#L615) a `get_translations_path()` method to override ([unlike `flask-babelex`](https://github.com/mrjoes/flask-babelex/blob/0.9.4/flask_babelex/__init__.py#L516)).


samuelhwilliams (approved):
Thank you - good catch. We definitely need some better tests around some of these areas - could you raise an issue to track that (and/or if you're willing to write some tests around babel, that would be very useful) 🙏

samuelhwilliams:
Would you mind also adding a line to `docs/changelog.rst` recording this bug fix?

Along the lines of:

```
2.0.0a1
-------

Fixes:

* Fix `translations_path` attribute when Flask-Admin is used with Flask-Babel.
```

LeXofLeviafan:
I've added a test for `translations_path`… though it's rather low-level :sweat_smile: (The upside is that it doesn't require doing actual environment setup as it only checks whether the specified directory gets used when building the translations cache.)

As for raising an issue – I suspect you'd know how to word it better (and what else needs to be tested) as you're more familiar with the codebase :thinking:

samuelhwilliams (approved):

PR #2524, opened 2024-09-15
Fix `Apply` filter button show/hide

### LINKED ISSUE #2520: Filter 'Apply' Button Visibility Issue Introduced in PR #2501
After the merge of #2501, the [filter "Apply" button](https://github.com/pallets-eco/flask-admin/blob/5d07c88eb953ebc801106b184bc8de4f3726919a/flask_admin/templates/bootstrap4/admin/model/layout.html#L49) is visible even when no filter option has been selected. This button should only appear when a filter option is selected. Multiple class attributes were applied to the button element. Previously, `style="display: none;"` was used to toggle the visibility.

### Steps to Reproduce

1. Go to a page where the filtering is enabled.
2. The "Apply" button is visible despite no filters being selected.

### Expected Behavior
The "Apply" button should be hidden when no filters are selected and only appear when a filter option is selected.

### Possible Solution
The issue might be resolved by revising the Javascript code from bs4_filters.js.

Environment:

- Python version: 3.10
- Flask version: 3.0.3
- Flask-Admin version: 2.0.0a0

samuelhwilliams:
This became permanent visibly after https://github.com/pallets-eco/flask-admin/pull/2501 because jQuery's show/hide
methods work toggle inline styles for `display: none`, but that PR
switched over to the `d-none` class from bootstrap.

This patch adds/removes the `d-none` class now.

Fixes #2520


pajusmar:
LGTM!

Issue #1974, opened 2020-04-07
Translations not extracted


lnlrbr:
Some sentences are not extracted in .PO files, like "Delete record". Seems to re-run a pybabel extract

forlAIght:
Does this mean that  some fields are translated , some are not?(Most of them is the defaulted template, like "list" "edit" "create")
I am also troubled by this problem.   （Mark!

LionelR:
Yes. The majority of the sentences are translated but some aren't.

PR #2526, opened 2024-09-23
Re-run babel-extract

samuelhwilliams:
fixes https://github.com/pallets-eco/flask-admin/issues/1974


samuelhwilliams:
well that's a lot of test failures for something innocuous 🤔

samuelhwilliams:
Right, I think that's all sorted now.

@ElLorans @LeXofLeviafan - would you mind having a glance over this and seeing if you're happy with the updates? It shouldn't affect any translation strings other than:

* Removing old strings that are no longer present
* Adding new msgids with empty msgstrs, ready for translations to be put in
* Adding equivalent msgstrs to the `en` file.

I haven't used Babel before to do any translation work, so would appreciate someone else looking over this unless I've done something unexpected/undesirable.

LeXofLeviafan:
Note that unless you override the `POT-Creation-Date` timestamp within the template file with a static value (e.g. using `sed` or any equivalent after executing `pybabel extract`), this timestamp will be updated in _all_ translations every time you run the generation script (even if that particular translation was not changed and you only wanted to make sure all translations are compiled); which would also modify _all compiled translations_ as well.

Also, here's some (ad-hoc) translations for the "Delete record" string that [were merged in buku](https://github.com/jarun/buku/pull/782/files) (…can't say for sure if they're accurate though :sweat_smile:)
```po
# de
msgid "Delete record"
msgstr "Datenzatz löschen"

# fr
msgid "Delete record"
msgstr "Supprimer l'enregistrement"

# ru
msgid "Delete record"
msgstr "Удалить запись"
```

samuelhwilliams:
> Note that unless you override the POT-Creation-Date timestamp within the template file with a static value (e.g. using sed or any equivalent after executing pybabel extract), this timestamp will be updated in all translations every time you run the generation script (even if that particular translation was not changed and you only wanted to make sure all translations are compiled)

Interesting - good shout out - let me look at handling that 🤔 

I'll pull in those translations, thanks!

babel/babel.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/sh
 pybabel extract -F babel.ini -k _gettext -k _ngettext -k lazy_gettext -o admin.pot --project Flask-Admin ../flask_admin
+
+# Uncomment below if new messages need to be pulled in
+# pybabel update -i admin.pot -d ../flask_admin/translations -D admin -N
```
LeXofLeviafan, line 5: Wouldn't it make more sense to do this via CLI parameter? E.g.:
```sh
[ "$1" = '--update' ] && pybabel update -i admin.pot -d ../flask_admin/translations -D admin -N
```

samuelhwilliams: Yes 👌

flask_admin/translations/af/LC_MESSAGES/admin.po
```diff
@@ -368,274 +358,242 @@ msgstr "Integriteit fout. %(message)s"
...
+#: ../flask_admin/templates/bootstrap4/admin/model/list.html:195
 msgid "Please select at least one record."
 msgstr "Kies ten minste een rekord."
 
-#: ../flask_admin/templates/bootstrap2/admin/model/row_actions.html:34
-#: ../flask_admin/templates/bootstrap3/admin/model/row_actions.html:34
-msgid "Are you sure you want to delete this record?"
-msgstr "Is jy seker jy wil hierdie rekord verwyder?"
+#~ msgid "ObjectId equals"
+#~ msgstr ""
```
LeXofLeviafan, line 589: Incidentally, these commented out strings are obsolete; if you no longer need these translations (as opposed to "slightly modified"/"temporarily renamed") you can remove them.

samuelhwilliams: Yep I spotted those. I'm happy to leave them - easy to restore later if somehow they come back, and there aren't so many that they're really causing a problem (AFAIK).

LeXofLeviafan: Actually if they _do_ come back they won't get automatically restored (and if they get obsoleted again even for a single run, the old translations will be removed automatically)… I'd suggest relying on git history instead, like it's done with regular code :sweat_smile:

samuelhwilliams: Noted, thanks

ElLorans: > Actually if they _do_ come back they won't get automatically restored (and if they get obsoleted again even for a single run, the old translations will be removed automatically)… I'd suggest relying on git history instead, like it's done with regular code 😅

That's really not my experience. If the previously obsolete string reappears in the source code and you run babel extract again, babel extract will:
1. Recognize that the string is present in the source code again.
2. Uncomment the translation and restore it to the active list of translatable strings
3. The translation itself (if already provided) will also be restored, assuming the context hasn't changed (same msgid)

samuelhwilliams:
Going to merge this as-is now. If anything needs tweaking or there are more translations, another PR is just a few clicks away :)

ElLorans:
> Right, I think that's all sorted now.
> 
> @ElLorans @LeXofLeviafan - would you mind having a glance over this and seeing if you're happy with the updates? It shouldn't affect any translation strings other than:
> 
>     * Removing old strings that are no longer present
> 
>     * Adding new msgids with empty msgstrs, ready for translations to be put in
> 
>     * Adding equivalent msgstrs to the `en` file.
> 
> 
> I haven't used Babel before to do any translation work, so would appreciate someone else looking over this unless I've done something unexpected/undesirable.

I confirm that now the strings are translated and everything looks right to me!

PR #2529, opened 2024-09-29
Fixing 'Save and Continue Editing' URL in edit form

### LINKED ISSUE #2311: "Save and Continue Editing" doesn't retain the backlink on edit page
Unlike both ["Save and Add Another" in the same form](https://github.com/flask-admin/flask-admin/blob/v1.6.0/flask_admin/model/base.py#L2145) and [the very same action in the create form](https://github.com/flask-admin/flask-admin/blob/v1.6.0/flask_admin/model/base.py#L2096), the [redirect URL for "Save and Continue Editing" in edit form](https://github.com/flask-admin/flask-admin/blob/v1.6.0/flask_admin/model/base.py#L2147) doesn't include a `url=return_url` parameter (which means at least losing page index when switching back to the list).

LeXofLeviafan:
fixes #2311

Action:
action

Before the fix (note the URL in location bar):
before

After the fix
after

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -10,6 +10,7 @@ Fixes:
 * Fix `translations_path` attribute when Flask-Admin is used with Flask-Babel
 * Some translation updates.
 * Date fields no longer override `widget` if set in `form_args`
+* “Save and Continue Editing” button no longer discards the “return URL” (allowing to retain filters when switching back to the list)
 
 2.0.0a0
 -------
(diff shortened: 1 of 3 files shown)


samuelhwilliams:
Hey @LeXofLeviafan  - please could you rebase and fix styling as we've started enforcing that now.

Agree in principle that this seems like something that should be fixed 👍 

If you could include a note in `doc/changes.rst` that would be great.

samuelhwilliams (approved):

PR #2532, opened 2024-10-01
Add horizontal form support via BootstrapTheme

cjmayo:
Allow horizontal forms to be used as described in https://getbootstrap.com/docs/4.0/components/forms/#horizontal-form and used in #2368, although here no extra div is added to the template and it is configured from the Theme object. That was not quite so simple as there is also template rendering going on in form.widgets.RenderTemplateWidget - so this PR adds theme to current_app ???

Also.
Boostrap 4 renamed .control-label to .col-form-label. 
https://getbootstrap.com/docs/4.0/migration/#forms-1

Closes #2368.


samuelhwilliams:
I think this may need a bit more thought - if an app has multiple admin instances with different themes, these settings will get mixed up.

Also, injecting the Flask-Admin theme as `theme` on the global app config seems to have a high possibility of overlapping with userspace config. We should probably prefix it with `flask_admin_` or something similar.

PR #2533, opened 2024-10-02
Publish constructor docstrings

cjmayo:
There is a lot of info in `__init__` docstrings that is not published with Sphinx. This changes that using `autoclass_content = 'both'`.

There are some instances in https://flask-admin.readthedocs.io/en/latest/api/mod_form_rules/ where there is duplication between the class signature and `__init__`, fix that.

And in https://flask-admin.readthedocs.io/en/latest/api/mod_form_upload/ the `__init__` and class signatures are different so need both.

### DIFF
--- a/doc/api/mod_form_rules.rst
+++ b/doc/api/mod_form_rules.rst
@@ -4,34 +4,19 @@
 .. automodule:: flask_admin.form.rules
 
 	.. autoclass:: BaseRule
-		:members: __init__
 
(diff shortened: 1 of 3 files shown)


samuelhwilliams:
Thanks for this - just a tiny fix required if you have a moment as styling rules are now enforced.

samuelhwilliams (approved):

PR #2534, opened 2024-10-03
Remove duplicate and unused imports

cjmayo:


samuelhwilliams:
I think we're better off getting the style/formatting stuff turned on, so will close this in favour of that. Thanks for the nudge though.

PR #2535, opened 2024-10-03
Use flask Sphinx theme from pallets_sphinx_themes

cjmayo:
pallets_sphinx_themes is already included in requirements/docs.in, just start using it.
A bit of CSS from the old theme needed copying over to use the image on the home page.


samuelhwilliams (approved):
thanks!

PR #2539, opened 2024-10-13
Fixes for bootstrap4/admin/lib.html

cjmayo:
Bootstrap 4 renamed .control-label to .col-form-label. https://getbootstrap.com/docs/4.0/migration/#forms-1

---

The original fix commit from #2532


samuelhwilliams:
Could we do a full check of bs4 deprecations here? Looking at the link you shared, we also have a reference to `help-block` that needs to be updated. There may be others 🙏

samuelhwilliams (approved):
nice one! thanks

PR #2543, opened 2024-10-19
Switch to hatchling and exclude tests from installation

cjmayo:
We're installing all the tests into site-packages.

I've checked and with flit-core you have to move the tests directory out of the package directory, and that is not going to change.

That would mean a bunch of changes. 

Instead switch to hatchling. The tests are still included in the sdist for those (distros) that want them but not installed for ordinary usage.


samuelhwilliams:
The build system is currently consistent with the rest of the pallets-eco projects and I'm not sure we'd want to deviate here. Probably worth dropping into #pallets-eco in discord if you think this is a sensible change.

PR #2564, opened 2024-11-12
Fix commands in README for running the example

pamelafox:
The pip install command only works when you're inside the folder, so I added that as a step to the README.


samuelhwilliams:
Agree, I think there is a bit of work on the readme/docs that could definitely be done for how to eg run the examples. This is a fair improvement but possibly something more wholesale would be a useful extension in the future, too 🙇

samuelhwilliams (approved):

PR #2569, opened 2024-11-13
Update pyproject.toml testpaths directory

pamelafox:
Pytest can't find the tests when it's specified as "tests", since it's under flask_admin. This fixes that.

It's possible that coverage.run also needs an update? I haven't tried it yet:

[tool.coverage.run]
branch = true
source = ["flask_admin", "tests"]


samuelhwilliams (approved):
Good spot, thanks for fixing :)

PR #2571, opened 2024-11-16
Fix errors caused in jinja StrictUndefined mode

### LINKED ISSUE #2522: Multiple issues when using  `StrictUndefined` in app.jinja_env.undefined
I like to use StrictUndefined to catch more errors in my templates. This raises many errors in flask-admin. You can simply check by enabling it in tests.

```python
from jinja2 import StrictUndefined
app.jinja_env.undefined = StrictUndefined
```

Environment:

- Python version:3.12
- Flask version:3.0.3
- Flask-Admin version:2.0.0a0

For now I'm using a custom class to ignore these (I configure all flask-admin blueprints to start with `admin2`):

```python
class MyStrictUndefined(StrictUndefined):
    """Work like StrictUndefined except in flask-admin views/blueprints."""

    @property
    def is_admin2_blp(self) -> bool:
        if flask.request and flask.request.blueprint:
            return bool(flask.request.blueprint.startswith("admin2"))
        return False

    def __eq__(self, other: t.Any) -> bool:  # type: ignore[override]
        if self.is_admin2_blp:
            return type(self) is type(other)
        self._fail_with_undefined_error()

    def __ne__(self, other: t.Any) -> bool:  # type: ignore[override]
        if self.is_admin2_blp:
            return not self.__eq__(other)
        self._fail_with_undefined_error()

    #
    # def __hash__(self) -> int:
    #    return id(type(self))
    #
    # def __str__(self) -> str:
    #    return ""
    #
    # def __len__(self) -> int:
    #    return 0
    #
    # def __iter__(self) -> t.Iterator[t.Any]:
    #    yield from ()
    #
    # async def __aiter__(self) -> t.AsyncIterator[t.Any]:
    #    for _ in ():
    #        yield

    def __bool__(self) -> bool:  # type: ignore[override]
        if self.is_admin2_blp:
            return False
        self._fail_with_undefined_error()

    # def __repr__(self) -> str:
    #    return "Undefined"

```

pamelafox:
Fixes #2522

This PR addresses errors that arise when running jinja in StrictUndefined mode (like mypy for Jinja!). There were different ways that I addressed the issues depending on the attribute:

* extra_css/extra_js: I added these as attributes to the BaseView, since their usage and documentation implies that they should always exist as empty lists.
* form.csrf_token: This is only defined if the dev is using SecureForm as the base class, which isn't always the case, so I used the pattern of "form.csrf_token is defined and form.csrf_token" which seems to be the recommended way in Jinja to ensure a variable exists before using it.
* return_url: From what I could tell, this is only passed in sometimes? So I made an if to only use it if it exists. But I might be wrong about that.
* widget.input_type: Only the classes that output HTML input fields have this defined. I considered adding input_type: Optional[str] = None to all the other classes, but that would have been a larger change, so I added the "is defined" check to templates instead.
* config.FLASK_ADMIN*: Since these will only be defined by devs who want to use that particular functionality, I used the "is defined" pattern for them.

Generally, for most cases, I ended up using the "is defined" pattern, but if you think there are more situations where .py changes would be warranted, let me know!


doc/changelog.rst
```diff
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+2.0.0a3
+-------
+
+Fixes:
+
+* Jinja templates can now be loaded in StrictUndefined mode.
```
pamelafox, line 9: I wasn't sure if 1) this change warranted an entry,
2) this was the correct version header

examples/sqla/app.py
```diff
@@ -3,6 +3,7 @@
 
 from admin import app
 from admin.data import build_sample_db
+from jinja2 import StrictUndefined
```
pamelafox, line 6: This is not the optimal import sorting, given that "admin" is a local folder, but ruff does not know that. I could use "." to make it clearer, or we could just be okay with it.
This seems to be the only example complex enough to have both an app.py and a folder of other modules.

doc/changelog.rst
samuelhwilliams: This looks great - that'll be the next version, so perfect :)

examples/sqla/app.py
samuelhwilliams: I am personally fine with it. If someone was so inclined, we could probably add some ruff config that would help it sort "correctly" - but I consider it broadly unimportant given the small scope, as you note.

samuelhwilliams (approved):
Great work on this - thank you :)

PR #2573, opened 2024-11-21
Upgrade Azure Blob Storage SDK to v12

### LINKED ISSUE #2566: Upgrade azure-storage-blob to >12
This package currently brings in a potentially old azure-storage-blob (2.1.0), we'd like to update to >12 which may have some SDK interface changes.

pamelafox:
Fixes #2566

This PR upgrades the azure-storage-blob SDK from v2 to v12, which involved a lot of interface changes. I followed the migration guide @ https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/migration_guide.md and was able to get all the previous functionality working, at least according to the tests.

For ease of testing, I added a simple example app and a devcontainer.azure.json which brings in the Azurite local emulator. That means you can open this repo inside a Codespace or Dev Container with that configuration, and Azure Blob Storage will be running for you.


pamelafox:
There's an admin test that is failing in CI that isn't failing locally, so I'm going to put some debugging in in the next commits to try to figure out what's happening.

samuelhwilliams:
FYI, it _is_ failing for me locally too 👀  Not that it helps much ... but I can possibly investigate a bit myself and see if I find anything.

samuelhwilliams:
If it helps:

```
/Users/sam/work/personal/flask-admin/flask_admin/contrib/fileadmin/azure.py(220)read_file()
-> blob = self._container_client.get_blob_client(path).download_blob()
(Pdb) ll
215         def read_file(self, path):
216             path = self._ensure_blob_path(path)
217             if path is None:
218                 raise ValueError("No path provided")
219             breakpoint()
220  ->         blob = self._container_client.get_blob_client(path).download_blob()
221             return blob.readall()
(Pdb) pp path
'dummy.txt'
(Pdb) n
azure.core.exceptions.DeserializationError: Unable to deserialize response data. Data: undefined, bytearray
> /Users/sam/work/personal/flask-admin/flask_admin/contrib/fileadmin/azure.py(220)read_file()
-> blob = self._container_client.get_blob_client(path).download_blob()
```

pamelafox:
Thanks! Might be a statefulness thing, I'll start a fresh env.

LeXofLeviafan:
...Wouldn't it be better to include actual information in that `error` variable? I.e. make it an optional string (containing the error message if an error occurred).

The check would work the same, and the error could be logged properly before redirect (...as it probably should be, anyway.)

pamelafox:
@samuelhwilliams It was an error due to the tests using an older Azure storage emulator, I've updated them to the official Microsoft hosted emulator now, and they're passing fine. 
@LeXofLeviafan I agree that it would be nice if the errors were easier to debug. They should probably be logged as either warning or error, depending on whether they're likely user errors or server errors. That would be better in a different PR though, I think. I'm trying to only affect the Azure module for this one.

.devcontainer/azure/devcontainer.json
```diff
@@ -0,0 +1,14 @@
+// For format details, see https://aka.ms/devcontainer.json.
+{
+	"name": "flask-admin (Python + Azurite)",
```
pamelafox, line 3: In the future, I could add Postgres and Mongo to this dev container too, to have a single container that can run all the tests. It should be fairly easy given tests.yaml has the services setup, just copying that to docker-compose.yaml.

flask_admin/contrib/fileadmin/azure.py
```diff
@@ -1,13 +1,19 @@
...
-    from azure.storage.blob import BlockBlobService
-except ImportError:
-    BlobPermissions = BlockBlobService = None
+    from azure.core.exceptions import ResourceExistsError
+    from azure.storage.blob import BlobProperties
+    from azure.storage.blob import BlobSasPermissions
+    from azure.storage.blob import BlobServiceClient
+    from azure.storage.blob import generate_blob_sas
+except ImportError as e:
+    raise Exception(
```
pamelafox, line 12: I changed the handling of the importerror to make mypy happier. I looked at other optional modules and they all seemed to handle importerrors a bit differently. I found one that did it this way. This seems fine since you should import the module unless you're using it.

pamelafox:
Hm, it appears I still don't have an approach for handling unimportable Azure modules that makes mypy happy. Let me know if you have any thoughts about the best practice for importing extra modules in tests (and skipping tests if they don't exist). I'll take another look Monday otherwise.

samuelhwilliams:
I don't think we need to support tests running without azure installed - I'd probably be fine with you removing the try/except around that test import.

pamelafox:
Okay, I've made it so that the tests assume you've got azure-blob-storage installed. I've also made the tests devcontainer bring in Postgres and Mongo too, so I was able to get all the tests passing in my dev container environment, without additional setup.

flask_admin/contrib/fileadmin/azure.py
```diff
@@ -47,28 +53,29 @@ def __init__(self, container_name, connection_string):
...
-
-        if not BlockBlobService:
-            raise ValueError(
-                "Could not import `azure.storage.blob`. "
-                "Enable `azure-blob-storage` integration "
-                "by installing `flask-admin[azure-blob-storage]`"
-            )
-
         self._container_name = container_name
         self._connection_string = connection_string
```
samuelhwilliams, line 57: One of the changes I made on the s3 admin side when bringing it up to date was to have `__init__` take a client instance rather than parameters that get passed the client.

Do you think we should do something similar here and accept an instance of `BlobServiceClient`, or is it still fine to just use the connection string?

samuelhwilliams:
I'm not sure the `hstore` extension is being set up correctly for the devcontainer. At least, I just opened vscode for this PR and ran the tests, and it complained that `hstore` didn't exist. After manually creating it and re-running the tests seem good. Is that something you could check?

samuelhwilliams:
I also think I'm getting that same error in the devcontainer you were getting previously with `FAILED flask_admin/tests/fileadmin/test_fileadmin.py::TestLocalFileAdmin::test_file_admin - assert 200 == 302` 🤔

pamelafox:
@samuelhwilliams I put the hstore setup in the postAttachCommand, but that doesnt' run until really late in the process, so it's possible you ran the tests before that had been run. I can make it more robust by instead bringing in a Dockerfile for the postgis and putting the CREATE EXTENSIOn in there.

@samuelhwilliams Do you get that error when there are no changes locally? Sometimes errors happen with partially aborted tests due to dangling changes to dummy files.
And is that in Codespaces or VS Code Dev Containers locally?

samuelhwilliams:
Apologies - both of these problems were on my end, my checkout of your latest changes failed so I still had an old version of the PR cached.

After resetting that, unfortunately I get a new error that looks unrelated - not able to install `requirements/dev.txt`, presumably due to some pinned package incompatibilities. I think that probably deserves to be out-of-scope for fixing in this PR though... After manually changing devcontainer to install `requirements/tests.txt` instead, which _does_ work, there's still some test failures due to an apparent incompatibility somewhere related to [Flask 3.1.0's partitioned cookies](https://github.com/pallets/flask/pull/5499).

But I think, yeah, neither of these things are this PR's problem...

samuelhwilliams:
Running the example app locally, after uploading a file, when I click the file to trigger a download I'm getting an error:

image

Can you tell if this a fundamental issue or just related to the example app?

flask_admin/contrib/fileadmin/azure.py
samuelhwilliams: It would be, but this is scheduled to go out for the v2 release where we're making a bunch of breaking changes, so I'm ok with it.

If you'd be happy to, feel free :) 🙏

pamelafox:
@samuelhwilliams I ran into both install issues. I've been commenting out numpy and manually installing Flask==3.0.0. I wasn't sure whether to tackle those in the PR or not.

pamelafox:
@samuelhwilliams I'll check on the download error, that sounds like a flow I didn't do. (And perhaps isnt covered by tests, given they're passing). And add a link.

samuelhwilliams:
> @samuelhwilliams I ran into both install issues. I've been commenting out numpy and manually installing Flask==3.0.0. I wasn't sure whether to tackle those in the PR or not.

Ah ok that's reassuring for me at least. I think it's fine to leave it for a separate PR; I'll raise an issue on it...

> @samuelhwilliams I'll check on the download error, that sounds like a flow I didn't do. (And perhaps isnt covered by tests, given they're passing). And add a link.

Brill, thanks!

pamelafox:
The error with download_file looks like this Azurite issue:
https://github.com/Azure/Azurite/issues/656
So I'm actually going to first see if the issue happens with a prod Azure account, and assuming it doesn't, I can shift focus to configuring Azurite correctly (as described in the issue).
That bit of the code was the trickiest to upgrade, so it's possible there's a lingering issue there.

pamelafox:
Update:

* The download file code was originally generating a SAS URL and redirecting to that URL. To generate a SAS URL securely, you need to use keyless authentication, which is tricky to set up with Azurite. However, I realized we could just use flask's send_file after downloading the blob, since users dont need the URL, they just need the data, so I changed the code to use that. That's what I use for downloading blobs in my work projects.
* I changed the constructor to take in a BlobServiceClient, and put examples of two ways of making the client in the Azure example (one for prod with keyless auth, one for local with connection string).

pamelafox:
FYI: there are no type annotations for AzureFileAdmin and S3FileAdmin currently. When we add them in, we get a mypy error about "multiple values for keyword storage". I think it's confused by the args/kwargs around it. So I have left off type annotations on AzureFileAdmin for now, though I added them to the storage class in a few places.

doc/changelog.rst
```diff
@@ -6,6 +6,7 @@ Changelog
 
 Breaking changes:
 
+* Azure Blob Storage SDK has been upgraded from the legacy version (v2) to the latest version (v12). AzureFileAdmin now accept `blob_service_client` rather than `connection_string` to give more flexibility with connection types.
```
samuelhwilliams, line 9: Tiny one: this is now in the wrong block because the main branch has changed under this PR. If you're happy to rebase/merge in master, that'd be helpful. Otherwise I can tidy this up after the fact.

pamelafox:
Are you getting those with the local Azure connection or prod? I thought I
tested that rename flow for both local/prod but I can go through it again
and add a pytest so that CI should catch these issues.

samuelhwilliams:
Using `AZURE_STORAGE_CONNECTION_STRING=DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://127.0.0.1:10000/devstoreaccount1;`

pamelafox:
Hm, I haven't been able to replicate yet, on this branch in GitHub Codespaces with the dev container configuration.

One thing I could do early next week is to ask my colleagues to do a bug bash on this branch, with both local and prod, to see if they encounter any issues (since we all have Azure accounts). That'd be a fun excuse to intro them to flask-admin anyway.

Here's my successful rename:

Screenshot 2024-12-01 at 3 03 06 PM

And my logs - which aren't super useful except that they show which URL is being requested. From what I've read, Azurite "start_copy_from_url" should work as long as the server is the same (http://127.0.0.1:10000 in this case).

```
INFO:azure.core.pipeline.policies.http_logging_policy:Request URL: 'http://127.0.0.1:10000/devstoreaccount1/fileadmin-tests?restype=REDACTED&comp=REDACTED&prefix=REDACTED'
Request method: 'GET'
Request headers:
    'x-ms-version': 'REDACTED'
    'Accept': 'application/xml'
    'User-Agent': 'azsdk-python-storage-blob/12.24.0 Python/3.12.7 (Linux-6.5.0-1025-azure-x86_64-with-glibc2.36)'
    'x-ms-date': 'REDACTED'
    'x-ms-client-request-id': '29312128-b038-11ef-8796-0242ac120002'
    'Authorization': 'REDACTED'
No body was attached to the request
INFO:azure.core.pipeline.policies.http_logging_policy:Response status: 200
Response headers:
    'Server': 'Azurite-Blob/3.33.0'
    'content-type': 'application/xml'
    'x-ms-client-request-id': '29312128-b038-11ef-8796-0242ac120002'
    'x-ms-request-id': '86f5013e-ee43-4231-aa18-c934836d4e0d'
    'x-ms-version': 'REDACTED'
    'date': 'Sun, 01 Dec 2024 23:01:12 GMT'
    'Connection': 'keep-alive'
    'Keep-Alive': 'REDACTED'
    'Transfer-Encoding': 'chunked'
INFO:azure.core.pipeline.policies.http_logging_policy:Request URL: 'http://127.0.0.1:10000/devstoreaccount1/fileadmin-tests?restype=REDACTED&comp=REDACTED&prefix=REDACTED'
Request method: 'GET'
Request headers:
    'x-ms-version': 'REDACTED'
    'Accept': 'application/xml'
    'User-Agent': 'azsdk-python-storage-blob/12.24.0 Python/3.12.7 (Linux-6.5.0-1025-azure-x86_64-with-glibc2.36)'
    'x-ms-date': 'REDACTED'
    'x-ms-client-request-id': '2932fd7c-b038-11ef-8796-0242ac120002'
    'Authorization': 'REDACTED'
No body was attached to the request
INFO:azure.core.pipeline.policies.http_logging_policy:Response status: 200
Response headers:
    'Server': 'Azurite-Blob/3.33.0'
    'content-type': 'application/xml'
    'x-ms-client-request-id': '2932fd7c-b038-11ef-8796-0242ac120002'
    'x-ms-request-id': '2bf381a6-acfa-4fd0-98a6-dbd53aa6518a'
    'x-ms-version': 'REDACTED'
    'date': 'Sun, 01 Dec 2024 23:01:12 GMT'
    'Connection': 'keep-alive'
    'Keep-Alive': 'REDACTED'
    'Transfer-Encoding': 'chunked'
INFO:azure.core.pipeline.policies.http_logging_policy:Request URL: 'http://127.0.0.1:10000/devstoreaccount1/fileadmin-tests?restype=REDACTED&comp=REDACTED&prefix=REDACTED'
Request method: 'GET'
Request headers:
    'x-ms-version': 'REDACTED'
    'Accept': 'application/xml'
    'User-Agent': 'azsdk-python-storage-blob/12.24.0 Python/3.12.7 (Linux-6.5.0-1025-azure-x86_64-with-glibc2.36)'
    'x-ms-date': 'REDACTED'
    'x-ms-client-request-id': '2935dd08-b038-11ef-8796-0242ac120002'
    'Authorization': 'REDACTED'
No body was attached to the request
INFO:azure.core.pipeline.policies.http_logging_policy:Response status: 200
Response headers:
    'Server': 'Azurite-Blob/3.33.0'
    'content-type': 'application/xml'
    'x-ms-client-request-id': '2935dd08-b038-11ef-8796-0242ac120002'
    'x-ms-request-id': '38197744-d851-4f7b-80c1-0925836a038b'
    'x-ms-version': 'REDACTED'
    'date': 'Sun, 01 Dec 2024 23:01:12 GMT'
    'Connection': 'keep-alive'
    'Keep-Alive': 'REDACTED'
    'Transfer-Encoding': 'chunked'
INFO:azure.core.pipeline.policies.http_logging_policy:Request URL: 'http://127.0.0.1:10000/devstoreaccount1/fileadmin-tests/test.jpg'
Request method: 'PUT'
Request headers:
    'x-ms-requires-sync': 'REDACTED'
    'x-ms-copy-source': 'REDACTED'
    'x-ms-version': 'REDACTED'
    'Accept': 'application/xml'
    'User-Agent': 'azsdk-python-storage-blob/12.24.0 Python/3.12.7 (Linux-6.5.0-1025-azure-x86_64-with-glibc2.36)'
    'x-ms-date': 'REDACTED'
    'x-ms-client-request-id': '2938575e-b038-11ef-8796-0242ac120002'
    'Authorization': 'REDACTED'
No body was attached to the request
INFO:azure.core.pipeline.policies.http_logging_policy:Response status: 202
Response headers:
    'Server': 'Azurite-Blob/3.33.0'
    'etag': '"0x1DF02BBD3F33C30"'
    'last-modified': 'Sun, 01 Dec 2024 23:01:12 GMT'
    'x-ms-client-request-id': '2938575e-b038-11ef-8796-0242ac120002'
    'x-ms-request-id': 'a40b52e4-32ed-44bf-8fef-9b28c9662e2e'
    'x-ms-version': 'REDACTED'
    'date': 'Sun, 01 Dec 2024 23:01:12 GMT'
    'x-ms-copy-id': 'REDACTED'
    'x-ms-copy-status': 'REDACTED'
    'Connection': 'keep-alive'
    'Keep-Alive': 'REDACTED'
    'Content-Length': '0'
INFO:azure.core.pipeline.policies.http_logging_policy:Request URL: 'http://127.0.0.1:10000/devstoreaccount1/fileadmin-tests/2QMXPCB2-102372.JPG'
Request method: 'DELETE'
Request headers:
    'x-ms-version': 'REDACTED'
    'Accept': 'application/xml'
    'User-Agent': 'azsdk-python-storage-blob/12.24.0 Python/3.12.7 (Linux-6.5.0-1025-azure-x86_64-with-glibc2.36)'
    'x-ms-date': 'REDACTED'
    'x-ms-client-request-id': '293ab90e-b038-11ef-8796-0242ac120002'
    'Authorization': 'REDACTED'
No body was attached to the request
INFO:azure.core.pipeline.policies.http_logging_policy:Response status: 202
Response headers:
    'Server': 'Azurite-Blob/3.33.0'
    'x-ms-client-request-id': '293ab90e-b038-11ef-8796-0242ac120002'
    'x-ms-request-id': '360678d6-3589-463d-b133-bbd55c354b83'
    'x-ms-version': 'REDACTED'
    'date': 'Sun, 01 Dec 2024 23:01:12 GMT'
    'x-ms-delete-type-permanent': 'REDACTED'
    'Connection': 'keep-alive'
    'Keep-Alive': 'REDACTED'
    'Content-Length': '0'
INFO:werkzeug:127.0.0.1 - - [01/Dec/2024 23:01:12] "POST /admin/azurefileadmin/rename/?path=2QMXPCB2-102372.JPG HTTP/1.1" 302 -
INFO:azure.core.pipeline.policies.http_logging_policy:Request URL: 'http://127.0.0.1:10000/devstoreaccount1/fileadmin-tests?restype=REDACTED&comp=REDACTED&prefix=REDACTED'
Request method: 'GET'
Request headers:
    'x-ms-version': 'REDACTED'
    'Accept': 'application/xml'
    'User-Agent': 'azsdk-python-storage-blob/12.24.0 Python/3.12.7 (Linux-6.5.0-1025-azure-x86_64-with-glibc2.36)'
    'x-ms-date': 'REDACTED'
    'x-ms-client-request-id': '295d70de-b038-11ef-a4ed-0242ac120002'
    'Authorization': 'REDACTED'
No body was attached to the request
INFO:azure.core.pipeline.policies.http_logging_policy:Response status: 200
Response headers:
    'Server': 'Azurite-Blob/3.33.0'
    'content-type': 'application/xml'
    'x-ms-client-request-id': '295d70de-b038-11ef-a4ed-0242ac120002'
    'x-ms-request-id': '5beb3d5b-a4b6-4fb1-8d88-0c673fc83806'
    'x-ms-version': 'REDACTED'
    'date': 'Sun, 01 Dec 2024 23:01:13 GMT'
    'Connection': 'keep-alive'
    'Keep-Alive': 'REDACTED'
    'Transfer-Encoding': 'chunked'
```

pamelafox:
I'm bug bash'ing this branch with some colleagues tomorrow, so hopefully we'll replicate that issue.

flask_admin/contrib/fileadmin/azure.py
```diff
@@ -160,80 +152,78 @@ def get_breadcrumbs(self, path):
...
-            expiry=now + self._send_file_validity,
-            start=now - self._send_file_lookback,
+        path = self._ensure_blob_path(file_path)
+        if path is None:
+            raise ValueError("No path provided")
+        blob = self._container_client.get_blob_client(path).download_blob()
+        if not blob.properties or not blob.properties.has_key("content_settings"):
+            raise ValueError("Blob has no properties")
+        mime_type = blob.properties["content_settings"]["content_type"]
+        blob_file = io.BytesIO()
```
tonybaloney, line 169: I think the `blob_file` should be closed here. It will delete the buffer. `flask.send_file` doesn't seem to close it. It will be closed when the variable is GC'ed but if you're getting lots of files it could consume a lot of memory

pamelafox:
We just did a little bug bash. Possible issues:

- [ ] Directory name with "**" failed: Should that succeed? We think probably not?
- [ ] Directory name with "  " succeeded: Should that actually fail?
- [ ] The date of new folders is 1970-01-01 00:00:00, should probably be actual date
- [ ] Dates of files are in UTC/GMT, is that expected?
- [ ] Memory leak with send_file (see comment)
- [ ] Connection error when using a connection string with a *prod* storage account, versus keyless auth (which works fine).

Issues that are probably Codespace-related and shouldn't happen outside of Codespaces:

* HTTP 413 from nginx when uploading files > 100 MB
* 404 when clicking "Admin" link, due to :5000 at end of URL

Issues outside of this PR's scope:
- Anthony thinks the UI would be clearer if ".." was at top of the UI in directory view. That's not an issue for this PR.

I'll move the PR to WIP while I triage those issues.

We did not replicate the rename-file issue with Azurite, all attendees were able to rename.

samuelhwilliams:
I can still reproduce an error on file renaming that looks like this

image

From a bit of playing around, it seems limited to _large_ files (4MB+ or so) rather than JPG or a specific filetype. Can you reproduce with that?

pamelafox:
I tried a larger file, no luck, but can you attach your large file and also tell me what you tried renaming it to? In theory, I should be able to replicate, given we're in a containerized environment. If I still fail, then I'll DM next week on Discord and see if we can hop on a screen share or some thing.

flask_admin/contrib/fileadmin/azure.py
pamelafox: FYI, according to comments from David Lord of Flask issues, the file does eventually get closed: https://github.com/pallets/flask/issues/2468
I also did try the context manager approach, but immediately got an error from werkzeug being unable to access a closed file. So I've left this as is.

pamelafox:
Following up on bug bash issues:

* Timestamps of directories: that was an existing issue, where directories always passed down a last modified of 0, for both Azure and S3. However, I went ahead and changed the logic for Azure to grab the timestamp, since that looks nicer.
* UTC timezones: that's expected for flask-admin, samuel has an example checked in that shows how to add user-specific timezone rendering for those who want it
* Possible memory leak with send_file: According to my research, there shouldn't be a leak.

pamelafox:
I am still working through an issue with "rename" when connecting to a prod account using the connection string (versus using OAuth token-based connection).

samuelhwilliams:
> I tried a larger file, no luck, but can you attach your large file and also tell me what you tried renaming it to? In theory, I should be able to replicate, given we're in a containerized environment. If I still fail, then I'll DM next week on Discord and see if we can hop on a screen share or some thing.

Unfortunately all of the files I've ended up reproducing it with are personal ones that I'd rather not share (eg a passport scan) 😂 I'm willing to accept that this might be something about my setup...

pamelafox:
@samuelhwilliams I worked a bunch on this today, with help from an Azure Blob SDK engineer, and ended up changing the rename method so that it uses an asynchronous copy behind the scenes (versus synchronous).

Now the code works for me for:
* Local Azurite
* Prod Azure account with keyless auth
* Prod Azure account with connection string

I tested with files of up to a size of 5 GB on my prod Azure account, and was able to rename them all.
I don't know if this would resolve the issue you were seeing with Azurite, however, since I was never able to replicate that. It's possible.

I do not currently have any other planned changes, so I've marked this as ready for review again.

flask_admin/contrib/fileadmin/azure.py
```diff
@@ -160,80 +151,94 @@ def get_breadcrumbs(self, path):
...
-        copy = self._client.copy_blob(self._container_name, dst, src_url)
-        while copy.status != "success":
-            sleep(self._copy_poll_interval_seconds)
-            copy = self._client.get_blob_properties(
-                self._container_name, dst
-            ).properties.copy
+        src_blob_client = self._container_client.get_blob_client(src)
+        dst_blob_client = self._container_client.get_blob_client(dst)
+        copy_result = dst_blob_client.start_copy_from_url(src_blob_client.url)
+        if copy_result.get("copy_status") == "success":
```
samuelhwilliams, line 219: Is it well known in what conditions blob storage will decide to do this synchronously vs asynchronously? Is it just based on file size, or some other information?

I'm fairly uncomfortable with the sleep(10) below in the async journey. It was previously sleep(1), which is better, but I guess I'd prefer to avoid sleeping in a request at all.

I suppose if we're keeping the rename operation as a copy+delete we probably can't just move on immediately if the copy hasn't finished, so maybe this is an OK workaround if async is only going to happen for really large files. It mightstill be nice to poll for updates more frequently than every 10 seconds though?

samuelhwilliams (commented):
@pamelafox really nice work - sorry it's taken a bit of time to get around to testing this out locally.

When I upload files in a sub-directory, then go back to the root, the directory shows up multiple times (seemingly once for itself and once for each item in the directory). This is with local azurite:



Just spent a bit of time trying to test with a real azure account but I'm getting errors currently when trying to sign up for a storage account, so that's great 🙃

This feels close though - only spotted that one issue, really. A few other comments but nothing major.

LeXofLeviafan:
…There seems to have been no activity here for a few weeks by now. Was the work on this pull-request paused? :thinking:

pamelafox:
Sorry, just had some busy weeks! I will hopefully have time this week, if not, certainly next week.

samuelhwilliams:
Really appreciate your hard work and persistence on this @pamelafox - thank you 🙇

pamelafox:
@samuelhwilliams I'm pushing a fix for the double-directory issue. That was introduced when I tried to remedy the issue of the Date field displaying a timestamp of 0:

Screenshot 2025-02-11 at 11 58 16 AM

However, from what I can see in the S3 code, it might have the same issue, and I think this has always been an issue for Azure. Removing that 1970 date would be a larger change involving the BaseFileAdmin and possibly the templates, depending on how sorting works.

For the record, Azure shows blank dates for directories, since directories aren't real entities:
Screenshot 2025-02-11 at 11 21 29 AM

I assume that I should just leave that be, for now? Let me know what you think.

flask_admin/contrib/fileadmin/azure.py
pamelafox: Good question! I was never able to get it to go down this path, even with large files. I tested up to 20 GB, and it still didn't go down the polling path. The engineer said the Blob service decides behind the scenes when to use async vs sync, and that perhaps it's always using sync for a same-account copy. I've asked if we can find that out from the Blob service team for sure.
I'm also asking if we can decrease the sleep timeout.

pamelafox: Update: The SDK engineer said sleep(1) is good, so I've changed to that.

samuelhwilliams:
I think it's fine to leave the timestamp issue for directories for now, agreed. That can be addressed separately 👍

pamelafox:
Okay, then the only pending question is if we can get clarity on when/whether the async path for copying gets used. I've sent an inquiry to the Blob storage team, but don't know if I'll definitely get a response.

I've filed the date-for-directories issue here:
https://github.com/pallets-eco/flask-admin/issues/2598

samuelhwilliams:
I'm also happy to move ahead for now with the async path question unanswered, since that does still match existing behaviour.

samuelhwilliams:
Great work + perseverance on this @pamelafox 🙏

samuelhwilliams (approved):

PR #2578, opened 2024-11-28
The security context needs a theme attribute

mapio:
To use the new version of Flask-Security not only `admiin.base_template` must be replaced by `armin.theme.base_template` but the context dictionary must contain a `theme` attribute that should point to `admin.theme`.


samuelhwilliams (approved):
Good spot - thanks for sending in this update 🙏

PR #2581, opened 2024-12-10
upgrade azure blob storage to >= 12

ibigbug:
update package dependency


ibigbug:
closed in favor of #2573

PR #2586, opened 2024-12-26
fixed error caused by the fallback implementation of gettext

LeXofLeviafan:
Turns out that when used with keywords _in templates_, `gettext()` gets invoked by Jinja _**without**_ keyword params (guess they're applied to the returned string instead). Thus, the current fallback implementation causes an error in such a case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Flask-Admin
 
-Flask-Admin is now part of Pallets-Eco, an open source organization managed by the 
+Flask-Admin is now part of Pallets-Eco, an open source organization managed by the
 Pallets team to facilitate community maintenance of Flask extensions. Please update
 your references to `https://github.com/pallets-eco/flask-admin.git`.
 
--- a/flask_admin/babel.py
+++ b/flask_admin/babel.py
@@ -4,11 +4,11 @@ try:
 except ImportError:
 
     def gettext(string, **variables):
-        return string % variables
+        return string if not variables else string % variables
 
     def ngettext(singular, plural, num, **variables):
         variables.setdefault("num", num)
-        return (singular if num == 1 else plural) % variables
+        return gettext((singular if num == 1 else plural), **variables)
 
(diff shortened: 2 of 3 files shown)


flask_admin/babel.py
```diff
@@ -4,11 +4,11 @@
 except ImportError:
 
     def gettext(string, **variables):
-        return string % variables
+        return string if not variables else string % variables
```
LeXofLeviafan, line 7: This is based directly on [flask-babel implementation](https://github.com/python-babel/flask-babel/blob/master/flask_babel/__init__.py#L639).

README.md
```diff
@@ -1,6 +1,6 @@
 # Flask-Admin
 
-Flask-Admin is now part of Pallets-Eco, an open source organization managed by the 
+Flask-Admin is now part of Pallets-Eco, an open source organization managed by the
```
LeXofLeviafan, line 3: Autoformatted by `pre-commit`

samuelhwilliams:
Do you have a reproduction example or a test case we could add?

LeXofLeviafan:
Server code during app init (to enable app-level translations in templates):
```py
app.jinja_env.add_extension('jinja2.ext.i18n')
app.jinja_env.install_gettext_callables(gettext, ngettext, newstyle=True)
```
If the `gettext` passed into this is the _fallback implementation_ from flask-admin (e.g. when flask-babel is **not** installed within the env), a template expression like this would raise a `KeyError: 'bar'` at page render:
```jinja
{{ _('foo %(bar)s baz', bar='<code>bar</code>'|safe) }}  {# usecase: injecting HTML into text #}
```
As mentioned in the PR description, trying to debug the error reveals that when `gettext()` is invoked, it receives the translation string but _no_ keywords.

samuelhwilliams (approved):
I've tested this locally for functionality and it does seem correct - nice catch.

PR #2589, opened 2025-01-04
Update the introiduction section to add Flask-Basicauth example

howiezhao:
This PR adds an example of integrating Flask-Basicauth in the Introduction section of the documentation.

### DIFF
--- a/doc/introduction.rst
+++ b/doc/introduction.rst
@@ -121,6 +121,30 @@ under development, before you want the whole world to see it.
 Have a look at `Flask-BasicAuth <https://flask-basicauth.readthedocs.io/>`_ to see just how
 easy it is to put your whole application behind HTTP Basic Auth.
 
+Below is an example of integrated Flask-Basicauth::
+
+    from flask_admin import Admin, AdminIndexView, expose
+    from flask_admin.contrib.sqla import ModelView
+    from flask_basicauth import BasicAuth
+
+    basic_auth = BasicAuth(app)
+
+    class MicroBlogAdminIndexView(AdminIndexView):
+        @expose('/')
+        @basic_auth.required
+        def index(self):
+            return super().index()
+
+    class MicroBlogModelView(ModelView):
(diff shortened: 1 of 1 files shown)


samuelhwilliams (commented):
Could we perhaps do this as a working example in the `examples/` directory and point to that? This snippet implies a few things already exist (app, Post, db) - and it'd be nice to just be able to see a functioning example straight away.

howiezhao:
> Could we perhaps do this as a working example in the `examples/` directory and point to that? This snippet implies a few things already exist (app, Post, db) - and it'd be nice to just be able to see a functioning example straight away.

Hi @samuelhwilliams Thank you for your reply. The reason why I did not put it in the `examples/` is that I think this example is very simple. Its core code is only the following lines:
```python
class MicroBlogAdminIndexView(AdminIndexView):
    @expose('/')
    @basic_auth.required
    def index(self):
        return super().index()

class MicroBlogModelView(ModelView):
    def is_accessible(self):
        return basic_auth.authenticate()

    def inaccessible_callback(self, name, **kwargs):
        return basic_auth.challenge()
```
This example is very concise compared to the [Flask-Login](https://flask-admin.readthedocs.io/en/latest/introduction/#rolling-your-own) and [Flask-Security](https://flask-admin.readthedocs.io/en/latest/introduction/#using-flask-security) examples.

As for the `app, Post, db`, etc. you mentioned, they have actually been mentioned in the [context](https://flask-admin.readthedocs.io/en/latest/introduction/#adding-model-views) of this document. So, if the user reads this document completely, I don’t think there will be any gaps, what do you think?

samuelhwilliams:
I think it's alright to rely on context further up in the readme, but that a snippet of code in the README that worked if grabbed just by itself would generally be a better user experience.

I'll accept this PR in a couple days if you don't want to change it, but if you want to update the code snippet to work in isolation that would be great too 🙏

PR #2601, opened 2025-02-17
Typo fix base.py: add forgotten closing quote (") to the message {Export type "<type>" is not supported.}

princerb:
Typo (syntax) fix in flash message:
```py
flash(
  gettext('Export type "%(type)s not supported.', type=export_type),
  ...                  ^       ^
)
```


ElLorans:
This is pretty self explanatory and I would merge

ElLorans:
@samuelhwilliams can I merge this?

samuelhwilliams:
Yes, although we should regenerate the translation files as well since it's updating a translatable string. I'm doing a quick early pass in https://github.com/pallets-eco/flask-admin/pull/2617 and then it should be easier to see the incremental changes.

Issue #2609, opened 2025-03-29
Examples `auth`, `auth-flask-login`, `datetime-timezone`, and `sqla-association_proxy` fail to run: `KeyError: 'babel'`


bubanoid:
Running the `auth`, `auth-flask-login`, or `datetime-timezone` examples with `flask-babel` installed in the virtual environment results in the following error:
```
...
  File ".../python3.13/site-packages/flask_babel/__init__.py", line 46, in get_babel
    return app.extensions['babel']
           ~~~~~~~~~~~~~~^^^^^^^^^
KeyError: 'babel'
```

ElLorans:
Cannot check right now. Do you have flask-babel installed in your env? What happens if you uninstall it?

bubanoid:
Please check the PR https://github.com/pallets-eco/flask-admin/pull/2610/files, which solves the issue. There are all the answers.

1) The `requirements.txt` related to the examples should be extended with the instruction to install the `flask-babel`
2) Besides, the examples require registration of Babel: `babel = Babel(app)`.

ElLorans:
I am sorry but there are no answers there.
1. The examples do not need flask-babel
2. The examples do not need registration of Babel

Can you share your environment with the libraries installed?

bubanoid:
```
arrow==1.3.0
babel==2.17.0
blinker==1.9.0
click==8.1.8
colour==0.1.5
dnspython==2.7.0
email_validator==2.2.0
Flask==3.1.0
-e git+https://github.com/pallets-eco/flask-admin.git@87f7c0bbf40cd1d506ea201b047cc45488eb740a#egg=Flask_Admin
flask-babel==4.0.0
Flask-Login==0.6.3
Flask-Principal==0.4.0
Flask-Security==5.6.1
Flask-SQLAlchemy==3.1.1
Flask-WTF==1.2.2
greenlet==3.1.1
idna==3.10
importlib_resources==6.5.2
itsdangerous==2.2.0
Jinja2==3.1.6
MarkupSafe==3.0.2
passlib==1.7.4
python-dateutil==2.9.0.post0
pytz==2025.2
ruff==0.11.2
setuptools==78.1.0
six==1.17.0
SQLAlchemy==2.0.40
sqlalchemy-citext==1.8.0
SQLAlchemy-Utils==0.41.2
tablib==3.8.0
types-python-dateutil==2.9.0.20241206
typing_extensions==4.13.0
Werkzeug==3.1.3
[...]

PR #2610, opened 2025-03-29
Fix crash on startup in `auth`, `auth-flask-login`, `sqla-association_proxy`, and `datatime-timezone` examples (#2609)

bubanoid:
Fix crash on startup in `auth`, `auth-flask-login`, `sqla-association_proxy`, and `datatime-timezone` example (#2609) by registering `app` with `Babel`.


bubanoid:
The examples are valid and require neither the installation of the `flask-babel` package nor the initialization of Babel within the Flask application. However, if the environment has `flask-babel` installed but the apps in the examples don't initialize it, the examples crash.

PR #2617, opened 2025-06-03
Update translation files

samuelhwilliams:
Re-run `babel.sh --update`


ElLorans:
We are missing the msgstr for "This field requires at least one item." in every language

samuelhwilliams:
Yep, looks like we are - I suspect that is just a new string that hasn't been translated yet, unless you think something is specifically broken?

ElLorans:
No absolutely not. Then I'll try to find the time and open a PR with missing translations, ok?

samuelhwilliams:
Amazing, thank you :)

PR #2619, opened 2025-06-07
Tweak testing instructions

samuelhwilliams:
- Recommend running via docker compose and `uv`, using the `.devcontainer/tests` docker compose file. This should simplify getting tests working locally because it provides postgres/azurite/mongo out of the box and sets environment variables for them.

I might leave explicit support for running the tests on your machine directly on the table for now, but it should still be very possible by setting those services up manually.


samuelhwilliams:
@ElLorans mind having a look at (and testing) this to see if it's an improvement / clear?

samuelhwilliams:
@pamelafox also would appreciate your eyes on this if you have time - I've tweaked the devcontainer setup so be good to know I haven't broken it for however you use it, as I don't 🙇

README.md
```diff
@@ -140,6 +140,19 @@ uv sync --extra all
 
 Tests are run with *pytest*. If you are not familiar with this package, you can find out more on [their website](https://pytest.org/).
 
+The tests rely on some external services running, including postgres, azurite and mongo. Currently, themost reliable way to run tests,
```
pamelafox:
```suggestion
The tests rely on some external services running, including postgres, azurite and mongo. Currently, the most reliable way to run tests,
```

.devcontainer/tests/Dockerfile
```diff
@@ -4,3 +4,11 @@ FROM mcr.microsoft.com/devcontainers/${IMAGE}
 RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     && apt-get -y install --no-install-recommends postgresql-client \
      && apt-get clean -y && rm -rf /var/lib/apt/lists/*
+
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
+
+WORKDIR /workspace
+
+COPY README.md LICENSE.txt ../../uv.lock pyproject.toml setup.cfg .python-version .
```
pamelafox: I'm curious why the partial copy here / how you choose what to copy? Going forward, when would this need to be adjusted with additional files?

README.md
```diff
@@ -140,6 +140,19 @@ uv sync --extra all
...
+without setting these services up manually on your machine, is to use the tests devcontainer docker compose configuration.
+
+```shell
+make test
```
pamelafox: Does this command also work *inside* the dev container? (I'm starting up a dev container now to see)
If not then I assume the instruction would be simply "uv run pytest" inside the container.

pamelafox:
I'm just testing this branch out in a Codespace (with the dev container config), and getting an error on the postCreateCommand:

"pip install --use-pep517 -r requirements/dev.txt"

That requirements/dev.txt file no longer exists. I assume it existed at the point where I wrote that command. Was it all moved into pyproject.toml?

.devcontainer/tests/Dockerfile
samuelhwilliams: I only want to copy the smallest set of files required to install requirements, so that docker can cache layers more effectively.

If we copy the entire repo inside at this point, any change to any file will mean docker wants to rebuild the following layers. In this case it's probably not a huge concern, but would be technically less efficient.

samuelhwilliams: Yep this is a mistake my side, leftover from when I wasn't doing the entire `flask_admin` copy. Good spot thanks

samuelhwilliams:
Getting everything lined up to run 1) natively, 2) inside docker compose and 3) inside devcontainers in a hands-off way is getting quite fiddly 🫠  Going to keep plumbing away a little bit longer but might start adding some constraints soon if not.

samuelhwilliams:
@pamelafox I think those are all addressed now - inside the devcontainer you can just run `uv run pytest` or `uv run tox` 👀

Have out-of-scoped running the tests outside of devcontainers in a fully-hands-off way for now, but I'd like to think about this some more. Possibly we could use `testcontainers` via pytest to spin up postgis/azurite/mongo for us, instead of the devcontainer doing it. I think that would let us run tests more easily both natively and inside devcontainers 🤔  I'll think about it separately.

ElLorans:
> @ElLorans mind having a look at (and testing) this to see if it's an improvement / clear?

I was on holidays! Hope to tackle it in the next 2 weeks.

ElLorans:
1. I would remove the part for running pytest locally, since it's too complex to set up and there are no instructions.
2. docker noob here, but maybe that's what we need to make the tests easy to run for everyone:

I have docker.io installed. `make test-in-docker` gives 
```
docker compose -f .devcontainer/tests/docker-compose.yaml run --remove-orphans app uv run pytest
unknown shorthand flag: 'f' in -f
```

After following https://docs.docker.com/engine/install/ubuntu/#install-using-the-repository, `make test-in-docker` gives

```
docker compose -f .devcontainer/tests/docker-compose.yaml run --remove-orphans app uv run pytest
unable to get image 'mcr.microsoft.com/azure-storage/azurite:latest': permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.49/images/mcr.microsoft.com/azure-storage/azurite:latest/json": dial unix /var/run/docker.sock: connect: permission denied
make: *** [Makefile:157: test-in-docker] Error 1
```


I first have to start docker `sudo systemctl start docker` and then `make test-in-docker` fails
```
docker compose -f .devcontainer/tests/docker-compose.yaml run --remove-orphans app uv run pytest
unable to get image 'mcr.microsoft.com/azure-storage/azurite:latest': permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.49/images/mcr.microsoft.com/azure-storage/azurite:latest/json": dial unix /var/run/docker.sock: connect: permission denied
make: *** [Makefile:157: test-in-docker] Error 1
```

Running `sudo make test-in-docker` works. What is the best alternative?
3. Can we set COMPOSE_BAKE=true for better performance?
4. I get `sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedObject) type "hstore" does not exist`. I am not sure how to fix it.

pamelafox:
@samuelhwilliams 

When I start in the Codespace and it runs the uv command, it then prompts for the password:

Screenshot 2025-06-23 at 9 49 51 AM

Is there a reason that we need to prompt for password, given its the local docker postgres? Seems like we could pass in the password.

samuelhwilliams:
Ah yes passing it in should be totally fine 👌

ElLorans:
I added a SQL script that includes `CREATE EXTENSION IF NOT EXISTS hstore;`—this was necessary to get the tests passing on my end. I went ahead and committed it here: https://github.com/pallets-eco/flask-admin/pull/2619/commits/b94ab156660ea66ed2d4ef525a0781bcb7acf1f1. Let me know if you'd prefer a different approach. Apologies for being a bit forward with the commit.

EDIT: all issues solved for me. I get  `187 passed, 1 skipped, 5 xfailed in 24.33s`. The skipped test is in `test_fileadmin_s3.py`. I would make the instructions more clear since I needed to:

1. `docker.io` errors out; Docker needs to be installed from the official Docker website
2.  ensure you are running docker, otherwise run `sudo systemctl start docker` . 
3. `make test-in-docker`. If your user does not have access to docker, you either need sudo or must: run `sudo usermod -aG docker $USER` , then restart the machine (logging out was not sufficient on Ubuntu)

samuelhwilliams:
Nice, thanks for checking! Adding the commit is totally fine, this one is definitely a communal PR :) All working on my end as well.

README.md
```diff
@@ -140,64 +139,38 @@ uv sync --extra all
...
-    Ran 102 tests in 13.132s
-
-    OK
-
-**NOTE!** For all the tests to pass successfully, you'll need several services running locally:
-Postgres (with the postgis and hstore extension), MongoDB, and Azurite.
-You'll also need *libgeos* available.
-See tests.yaml for Docker configuration and follow service-specific setup below.
+```shell
+uv run pytest
```
ElLorans, line 149: I think I would remove this line, since it's complex to set up locally and there are no instructions. The default should be `make test-in-docker`

README.md
```diff
@@ -140,64 +139,38 @@ uv sync --extra all
...
-# Install postgis and geos
-brew install postgis
-brew install geos
+If you don't use devcontainers and simply want to run the basic test suite against the default python installation, you can use:
 
-# Set up a PostgreSQL user
-createuser -s postgresql
-brew services restart postgresql
+```shell
+make test-in-docker
```
ElLorans, line 163: Does it make sense to add

```
sudo systemctl start docker
make test-in-docker
```

If you get a permission error, run `sudo usermod -aG docker $USER` to grant your user access to docker, or use `sudo make test-in-docker`

samuelhwilliams: I don't think so - I think it should be reasonable to assume that developers have docker working (we can add a line saying this, but maybe don't need to include the command - esp as systemctl is OS-specific).

README.md
samuelhwilliams: I think we are also fine to support running the tests within the devcontainer, which is what this line is about, rather than running them natively.

samuelhwilliams:
@ElLorans @pamelafox are either/both of you happy with this as-is or need anything else done here? :)

PR #2620, opened 2025-06-08
Refactor: Replace javascript:void(0) with # for CSP compliance

bubanoid:
I've replaced all instances of `javascript:void(0)` in href attributes with `#` within the Bootstrap 4 templates. This change is part of the effort to make the package more Content Security Policy (CSP) compliant.

I modified the following files:
- `flask_admin/templates/bootstrap4/admin/actions.html`: Updated in the dropdown macro for both the main link and individual action links.
- `flask_admin/templates/bootstrap4/admin/layout.html`: Updated in the menu macro for dropdown menu items.
- `flask_admin/templates/bootstrap4/admin/lib.html`: Updated in the pager macro for disabled pagination links.

Existing JavaScript `onclick` handlers are preserved, and the use of `#` as the href value for links that primarily trigger JavaScript actions or serve as placeholders should not impact functionality.

I manually tested dropdowns, actions, and pagination and confirmed that they continue to work as expected. I also ran automated tests, and any failures were determined to be unrelated to these changes (e.g., pre-existing issues, database connection timeouts, or missing cloud credentials in the test environment).


flask_admin/tests/fileadmin/files/dummy2.txt
ElLorans, line 1: what is this file's purpose?

samuelhwilliams: It gets created/managed by the tests somehow - I think if tests fail in some ways it doesn't get cleaned up properly. Probably should be .gitignored.

bubanoid: This file is an artifact from a previously failed test. All tests are passing now, so I don’t think we should add it to .gitignore - tests should pass before a PR is merged.

UPD
The file was removed.

samuelhwilliams:
Not against this in principle; can try and review if we get this updated 👍

samialfattani:
Related #2715

Issue #2628, opened 2025-06-27
ModuleNotFoundError


ElLorans:
due to 7a51179a21b1c7ab92f7e89161200fbdf07d6a2c, sqlalchemy is now a required library even if it is not listed in dependencies

```
Traceback (most recent call last):
  File "/home/orens/Desktop/MyStuff/Py/Web/flask-admin-use-uv/examples/peewee/app.py", line 20, in <module>
    from flask_admin.contrib.peewee import ModelView
  File "/home/orens/.cache/uv/environments-v2/app-d3726ec04d808c0d/lib/python3.9/site-packages/flask_admin/contrib/peewee/__init__.py", line 11, in <module>
    from .view import ModelView
  File "/home/orens/.cache/uv/environments-v2/app-d3726ec04d808c0d/lib/python3.9/site-packages/flask_admin/contrib/peewee/view.py", line 21, in <module>
    from flask_admin.contrib.peewee import filters
  File "/home/orens/.cache/uv/environments-v2/app-d3726ec04d808c0d/lib/python3.9/site-packages/flask_admin/contrib/peewee/filters.py", line 4, in <module>
    from flask_admin.model import filters
  File "/home/orens/.cache/uv/environments-v2/app-d3726ec04d808c0d/lib/python3.9/site-packages/flask_admin/model/__init__.py", line 2, in <module>
    from .base import BaseModelView
  File "/home/orens/.cache/uv/environments-v2/app-d3726ec04d808c0d/lib/python3.9/site-packages/flask_admin/model/base.py", line 22, in <module>
    from sqlalchemy.orm import InstrumentedAttribute
ModuleNotFoundError: No module named 'sqlalchemy'

```

PR #2629, opened 2025-06-27
remove sqlalchemy import

ElLorans:
fixes https://github.com/pallets-eco/flask-admin/issues/2628 . due to https://github.com/pallets-eco/flask-admin/commit/7a51179a21b1c7ab92f7e89161200fbdf07d6a2c, sqlalchemy became a required library even if it is not listed in dependencies. Moving the import to `_types.py` under `TYPE_CHECKING` guard fixes the issue.

### DIFF
--- a/flask_admin/_types.py
+++ b/flask_admin/_types.py
@@ -49,6 +49,7 @@ if t.TYPE_CHECKING:
     import sqlalchemy  # noqa
     from sqlalchemy import Column as T_SQLALCHEMY_COLUMN
     from sqlalchemy import Table as T_TABLE  # noqa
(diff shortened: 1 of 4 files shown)


ElLorans:
I also removed imports for sqlalchemy-utils, as it was before https://github.com/pallets-eco/flask-admin/commit/7a51179a21b1c7ab92f7e89161200fbdf07d6a2c . Is the issue now that the flask_admin/contrib/sqla/validators.py can cause ImportError at runtime (as it was before)? See https://github.com/pallets-eco/flask-admin/blob/707cca65b6ee9d6cafa8b8316d0c7b9f386f4812/flask_admin/contrib/sqla/validators.py#L92

Issue #2630, opened 2025-06-29
babel.sh --update requires setuptools


ElLorans:
EDIT:
must run `uv sync --group docs`, then `./babel.sh --update`. I get 
```
Traceback (most recent call last):
  File "/home/orens/Desktop/MyStuff/Py/Web/flask-admin-use-uv/.venv/bin/pybabel", line 10, in <module>
    sys.exit(main())
  File "/home/orens/Desktop/MyStuff/Py/Web/flask-admin-use-uv/.venv/lib/python3.9/site-packages/babel/messages/frontend.py", line 998, in main
    return CommandLineInterface().run(sys.argv)
  File "/home/orens/Desktop/MyStuff/Py/Web/flask-admin-use-uv/.venv/lib/python3.9/site-packages/babel/messages/frontend.py", line 924, in run
    return cmdinst.run()
  File "/home/orens/Desktop/MyStuff/Py/Web/flask-admin-use-uv/.venv/lib/python3.9/site-packages/babel/messages/frontend.py", line 522, in run
    for filename, lineno, message, comments, context in extracted:
  File "/home/orens/Desktop/MyStuff/Py/Web/flask-admin-use-uv/.venv/lib/python3.9/site-packages/babel/messages/extract.py", line 217, in extract_from_dir
    yield from check_and_call_extract_file(
  File "/home/orens/Desktop/MyStuff/Py/Web/flask-admin-use-uv/.venv/lib/python3.9/site-packages/babel/messages/extract.py", line 282, in check_and_call_extract_file
    for message_tuple in extract_from_file(
  File "/home/orens/Desktop/MyStuff/Py/Web/flask-admin-use-uv/.venv/lib/python3.9/site-packages/babel/messages/extract.py", line 324, in extract_from_file
    return list(extract(method, fileobj, keywords, comment_tags,
  File "/home/orens/Desktop/MyStuff/Py/Web/flask-admin-use-uv/.venv/lib/python3.9/site-packages/babel/messages/extract.py", line 446, in extract
    raise ValueError(f"Unknown extraction method {method!r}")
ValueError: Unknown extraction method 'jinja2'
```

PR #2631, opened 2025-06-29
Update babel translations

samuelhwilliams:
- Update babel.sh and makefile to use `uv`
- Manually update all admin.po files to fix the msgid changed by https://github.com/pallets-eco/flask-admin/pull/2601
- Re-run `babel.sh --update` from `babel` dir to update all translations
- Comment out translations for our docs; these haven't ever been generated+committed before so don't believe this is supported or used anywhere right now.

fixes: https://github.com/pallets-eco/flask-admin/issues/2630


ElLorans:
Amazing! Do we want to remove crowdin mention from the readme and the crowdin files in /babel, too? I am updating the translations but we have a lot of languages. Does it make sense to have Spanish Catalan and so many others?

samuelhwilliams:
I think removing crowdin does make sense as well, yeah - it's not been used for a long time and I think is not intended to be used any more.

Will need to work out how to disable flask-admin on crowdin altogether to stop any wasted effort 👀

samuelhwilliams:
I'm gonna run the crowdin stuff as a separate PR/issue: https://github.com/pallets-eco/flask-admin/issues/2633

PR #2654, opened 2025-08-08
type hints

ElLorans:
Partially addresses https://github.com/pallets-eco/flask-admin/issues/2440

- [x] Re-enable warn_unused_configs in pyproject.toml
- [x] Re-enable warn_redundant_casts in pyproject.toml
- [x] Re-enable warn_unused_ignores in pyproject.toml
- [x] Re-enable strict_equality in pyproject.toml
- [x] Re-enable strict_concatenate in pyproject.toml
- [x] Re-enable check_untyped_defs in pyproject.toml

Way too many `type: ignore` still required, due to highly dynamic dependencies and complex codebase. `examples/` had to be modified from `kebab-case` to `snake_case` due to mypy requirements.

- [x] **Bonus:** run mypy on examples, too
A simple `files = ["flask_admin",  "examples"]` cannot be used because examples are self contained and use relative imports which do not work from the parent folder.
The only solution I could find was to add a new step in the tox typing matrix.
```
[tool.tox.env.typing]
description = "run static type checkers"
dependency_groups = ["typing"]
commands = [
    ["mypy", "--python-version", "3.9"],
    ["mypy", "--python-version", "3.13"],
    ["mypy", "--python-version", "3.9", "--config-file", "examples/mypy-examples-config.toml", "examples", ]
```

I used a [new config file](https://github.com/pallets-eco/flask-admin/blob/fix/type-hints/examples/mypy-examples-config.toml) to allow differentiating vs the rest of the repo. Another possibility would be to dump the second config file and explicitate the different options, which right now are:
```
# this is the difference vs pyproject.toml
# relative imports in examples cause errors when mypy is run from proj folder
ignore_missing_imports = true
disable_error_code = "name-defined" # otherwise flask_sqlalchemy db.Model needs to be ignored, littering examples
warn_unused_ignores = false # otherwise we get conflict with first run of mypy on proj folder
```

Any feedback on the toml config is welcome.


pyproject.toml
```diff
@@ -177,13 +177,19 @@ source = ["flask_admin", "*/site-packages"]
 
 [tool.mypy]
 python_version = "3.9"
-files = ["flask_admin"]
+files = ["flask_admin", ]
```
hasansezertasan: Why do we have an extra comma here 🤓?

hasansezertasan (approved):
Great effort @ElLorans, thank you 🫶.

hasansezertasan:
> * [x]  **Bonus:** run mypy on examples, too
>   A simple `files = ["flask_admin",  "examples"]` cannot be used because examples are self contained and use relative imports which do not work from the parent folder.
>   The only solution I could find was to add a new step in the tox typing matrix.
> 
> ```
> [tool.tox.env.typing]
> description = "run static type checkers"
> dependency_groups = ["typing"]
> commands = [
>     ["mypy", "--python-version", "3.9"],
>     ["mypy", "--python-version", "3.13"],
>     ["mypy", "--python-version", "3.9", "--config-file", "examples/mypy-examples-config.toml", "examples", ]
> ```
> 
> I used a [new config file](https://github.com/pallets-eco/flask-admin/blob/fix/type-hints/examples/mypy-examples-config.toml?rgh-link-date=2025-08-08T11%3A48%3A36Z) to allow differentiating vs the rest of the repo. Another possibility would be to dump the second config file and explicitate the different options, which right now are:

Just saying, WDYT about [uv workspaces](https://docs.astral.sh/uv/concepts/projects/workspaces/)? Do you think it would work without a new config file 🤔?

ElLorans:
I think uv workspaces solve a different problem. Here, we would need to access each example's folder and run mypy from there (which should be feasible and slightly better than what I opted for)

pyproject.toml
ElLorans: It comes from when I was trying to add examples there. I’ll remove it tomorrow

flask_admin/contrib/pymongo/view.py
```diff
@@ -224,15 +235,15 @@ def _search(self, query, search_term):
...
         self,
-        page,
+        page: t.Optional[int],
         sort_column,
-        sort_desc,
-        search,
-        filters,
+        sort_desc: bool,
+        search: t.Optional[str],
+        filters: t.Optional[t.Sequence[T_FILTER]],
```
princerb, line 244: I think `typing.Sequence` is deprecated. See [here](https://docs.python.org/3/library/typing.html#typing.Sequence).

flask_admin/contrib/sqla/view.py
```diff
@@ -143,7 +145,7 @@ class MyModelView(ModelView):
           used.
     """
 
-    column_filters: t.Optional[t.Sequence[str]] = None
+    column_filters: t.Optional[t.Collection[t.Union[str, BaseFilter]]] = None
```
princerb, line 148: [`typing.Collection`](https://docs.python.org/3/library/typing.html#typing.Collection) is so as well.

flask_admin/contrib/pymongo/view.py
ElLorans: I think there are some issues with Python 3.9.0 and Python 3.9.1, so this will be fixed automatically by Ruff when we drop Python 3.9 and move to 3.10. https://stackoverflow.com/questions/65858528/is-collections-abc-callable-bugged-in-python-3-9-1 , https://github.com/astral-sh/ruff/issues/2690

flask_admin/contrib/sqla/view.py
ElLorans: see above!

flask_admin/contrib/pymongo/view.py
princerb: Yeah I see your concern

ElLorans:
@samuelhwilliams does this approach for type checking the examples too make sense to you?

samuelhwilliams:
I'd prefer if we used `Mapped` and `mapped_column` in sqlalchemy, since that's generally a newer/cleaner way of declaring columns with good type hints. Can you explain what problem you've had with that which made you switch away?

ElLorans:
> I'd prefer if we used `Mapped` and `mapped_column` in sqlalchemy, since that's generally a newer/cleaner way of declaring columns with good type hints. Can you explain what problem you've had with that which made you switch away?

They are not supported by the min sqlalchemy version, so I had to revert the examples to the old syntax in order to make mypy happy.

samuelhwilliams:
I think it's fine(/preferrable) for us to require the examples to be on the latest version of flask-admin and dependencies. I don't see a huge amount of value in having examples for older versions of things - it feels like it sets a bad precedent?

If the problem is mypy, then we can make mypy only run in the latest python version mode against examples :)

Flask-Admin needs to work with python 3.9; the examples don't, we can definitely just say "these are built for python 3.13"

ElLorans:
> I think it's fine(/preferrable) for us to require the examples to be on the latest version of flask-admin and dependencies. I don't see a huge amount of value in having examples for older versions of things - it feels like it sets a bad precedent?
> 
> If the problem is mypy, then we can make mypy only run in the latest python version mode against examples :)
> 
> Flask-Admin needs to work with python 3.9; the examples don't, we can definitely just say "these are built for python 3.13"

Perfect, I'll amend in the next few weeks

ElLorans:
I brought back `mapped` in the 2 examples I had modified.
 
Now the third mypy check crashes on the examples because the base env does not have a sqlalchemy version with mapped_column.
This can be solved with a `type: ignore[attr-defined]`, but the right fix is specifying tox to use py313 to run the third command. 
@samuelhwilliams I do not know how to change the third mypy command to use the env py313. I tried to split tool.tox.env.typing in 2 and use depends to call only tox -e typing, but then I get a pytest error because the `commands` arg is imported from base env. 
EDIT:
I was wrong, the issue was due to old stubs in the requirements.

ElLorans:
I removed the outdated https://pypi.org/project/types-Flask/ and https://pypi.org/project/types-Flask-SQLAlchemy/ since 
1. the former should be used only for Flask < 2.0 while we use Flask >= 2.0 and with Flask-SQLAlchemy we use Flask>=2.2 
2. Flask-SQLAlchemy includes type annotations from 3.0.1 and the min version is 3.0 so it was close enough for me.
If you think it makes sense, we can also bump Flask-SQLAlchemy min version to 3.0.1 but I see little value.
I also removed the .toml for examples and specified the different settings directly in the tox mypy command:
```
["mypy", "--python-version", "3.13",
        "--ignore-missing-imports",
        "--disable-error-code", "name-defined", # allow flask_sqlalchemy's db.Model in examples
        "--no-warn-unused-ignores",  # prevent conflict with mypy 1st run on src folder
        "examples", ]
```

samuelhwilliams (approved):

ElLorans:
@hasansezertasan does it look good to you as well?

hasansezertasan:
> @hasansezertasan does it look good to you as well?

I loved what you have done 🫶. Great job 🚀 🚀 🚀.

hasansezertasan (approved):

Issue #2525, opened 2024-09-16
Compatibility with flask async routes


williamhakim10:
Currently, flask-admin does not easily support flask routes declared as async (see https://flask.palletsprojects.com/en/3.0.x/async-await/). While this isn't a dealbreaker, if you have a bunch of async code (e.g., if you're using flask-admin alongside a Quart app), then it's pretty annoying to have to manually wrap all your code in `asyncio.run(...)` (or `loop.run_until_complete(...)`). My understanding is that the limitation stems from the `expose` decorator which does not call `ensure_sync`. It would be nice if `expose` could wrap async functions, or if `expose_async` existed as an alternative. I'd be happy to help with a PR if either of these approaches seem reasonable.

ElLorans:
A PR adding ensure_sync to expose would be welcome and should be enough. The PR should include a test addressing the issue to prevent regressions.

ElLorans:
Thinking back about this, there are too many other issues:

1. template rendering is synchronous (Jinja2 async mode is optional but not integrated)
2. Database/model operations (SQLAlchemy, MongoEngine, etc.) are synchronous by default.
3. Middleware, hooks, and utilities (get_list, on_model_change, etc.) are all sync
4. WTForms validation is sync.

Even if you make @expose async-friendly, the rest of the call chain isn’t ready for await.

Am I getting it wrong?

hasansezertasan:
I'm interested in this. It's not an easy task. There's some work to be done to achieve this 🤓. However, other Flask Admin-inspired tools like SQLAdmin and Starlette Admin already use async/await and I'd like to have that with Flask Admin too 😎.

After some thought and planning, we can prioritize everything and go from there.

ElLorans:
I fear that's not feasible. You cannot really have a single codebase for both sync and async. We have other priorities and not enough resources.

ElLorans:
I was discussing on Pallets' Discord Channel, and the consensus appears to be that ensure_sync is sufficient in Flask. So we can accept a PR, if somebody wants to work on it.

ElLorans:
I found some time, @williamhakim10 can you try https://github.com/pallets-eco/flask-admin/tree/feature/async?

PR #2659, opened 2025-09-08
support async

ElLorans:
fixes https://github.com/pallets-eco/flask-admin/issues/2525

### DIFF
--- a/flask_admin/base.py
+++ b/flask_admin/base.py
@@ -88,7 +88,7 @@ def _wrap_view(f: t.Callable) -> t.Callable:
         if abort is not None:
             return abort
 
-        return self._run_view(f, *args, **kwargs)
+        return self._run_view(current_app.ensure_sync(f), *args, **kwargs)
 
     inner._wrapped = True  # type:ignore[attr-defined]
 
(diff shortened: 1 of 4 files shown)


ElLorans:
Can I do anything to push this forward?

samuelhwilliams:
I am not super au fait with async - if you are both happy that this is all that's needed, I will take your work for it 👍

samuelhwilliams:
Do we need any documentation updates around this support?

ElLorans:
> I am not super au fait with async - if you are both happy that this is all that's needed, I will take your work for it 👍

I asked on Pallets' Discord Channel, and the consensus was that `ensure_sync `is sufficient to make any extension compatible with Flask async. This is confirmed by the test I added which are failing without the code edit, and passing now. The method was added in Flask 2.0, which is the min version we support. We can link this fork on Pallet's discord and ask for feedback/tests.
No additional docs are required, except stating in the changelog that Flask-Admin can now be used natively with Flask async (without wrapping in `asyncio.run(...)` (or `loop.run_until_complete(...)`), even though it remains sync. https://flask.palletsprojects.com/en/stable/async-await/#extensions

samuelhwilliams:
Thanks for explaining :)

Issue #2372, opened 2023-07-31
pymongo: integration of get_query into get_list method


MihailMiller:
In the current `pymongo` implementation of the `get_list` method in the `ModelView` class, the default query is defined as `{}`. There is no way to override this query without overriding `get_list`, which makes the code less maintainable. For example, the `sqla` contribution allows you to override the `get_query` when subclassing the `ModelView`. 

I propose to include a `get_query` call in the `get_list` method. This change would allow developers to override only the `get_query` method and pre-filter the database so that only relevant results are displayed according to the specific parameters of the use case.

### Proposed change:
Revise `get_list` in `ModelView` to call `get_query` when no specific query is specified:

```python
def get_query():
    return {}

def get_list(self, page, sort_column, sort_desc, search, filters, execute=True, page_size=None):
    query = self.get_query()
    # continue existing logic...
```

I will be happy to make a pull request if you would like to integrate this into the `pymongo` contribution.

PR #2661, opened 2025-09-09
pymongo: add pymongo.view.ModelView.get_query()

ElLorans:
fixes https://github.com/pallets-eco/flask-admin/issues/2372

### DIFF
--- a/flask_admin/contrib/pymongo/view.py
+++ b/flask_admin/contrib/pymongo/view.py
@@ -235,6 +235,9 @@ class ModelView(BaseModelView):
 
         return query
 
+    def get_query(self) -> dict[str, t.Any]:
+        return {}
+
     def get_list(  # type: ignore[override]
         self,
         page: t.Optional[int],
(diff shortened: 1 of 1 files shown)


ElLorans:
In the current pymongo implementation of the `get_list` method in the `ModelView` class, the default query is defined as `{}`. There is no way to override this query without overriding `get_list`, which makes the code less maintainable. For example, the sqla contribution allows you to override the `get_query` when subclassing the `ModelView`.

This PR includes a `get_query` call in the `get_list` method. This change allows developers to override only the get_query method and pre-filter the database so that only relevant results are displayed according to the specific parameters of the use case.

Maybe we should ensure that the different ModelView adhere to the API standard to prevent this. `@abstractmethod` or `typing.Protocol` would work.

samuelhwilliams (approved):

samuelhwilliams:
> Maybe we should ensure that the different ModelView adhere to the API standard to prevent this. @abstractmethod or typing.Protocol would work.

Agree - this feels very sensible

PR #2663, opened 2025-09-14
fix s3 cloud compatibility

samialfattani:
- this change needed whever the current os platform is different thn s3 cloud platform. In this case the os.path.pathnorm() will not work properly since it treats storage as it is placed in the current system.

- Before this PR, browsing deep path in FileAdmin view such as /admin/fileadmins3/b/xx/yy/zz responses 404 code. This can happend only when the current os is "Windows" and the cloud s3 is "non-Windows" and vise versa.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,0 @@
-{
-  "python.testing.pytestArgs": [
-    "flask_admin"
-  ],
-  "python.testing.unittestEnabled": false,
-  "python.testing.pytestEnabled": true
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+  "python.testing.pytestArgs": [
+    "flask_admin"
+  ],
+  "python.testing.unittestEnabled": false,
+  "python.testing.pytestEnabled": true
+}
--- a/examples/s3/main.py
+++ b/examples/s3/main.py
@@ -52,8 +52,8 @@ if __name__ == "__main__":
             S3FileAdmin(
                 bucket_name=bucket_name,
                 s3_client=s3_client,
-                # explicitly set the OS based on the platform of s3-cloud 
-                on_windows = False
+                # explicitly set the OS based on the platform of s3-cloud
+                on_windows=False,
             )
         )
 
--- a/examples/s3/main.py
+++ b/examples/s3/main.py
@@ -52,7 +52,7 @@ if __name__ == "__main__":
             S3FileAdmin(
                 bucket_name=bucket_name,
                 s3_client=s3_client,
-                # explicitly set the OS based on the platform of s3-cloud
+                # explicitly set the OS based on the platform of s3-cloud storage
                 on_windows=False,
             )
         )
--- a/examples/s3/main.py
+++ b/examples/s3/main.py
@@ -52,6 +52,8 @@ if __name__ == "__main__":
             S3FileAdmin(
                 bucket_name=bucket_name,
                 s3_client=s3_client,
+                # explicitly set the OS based on the platform of s3-cloud 
+                on_windows = False
             )
         )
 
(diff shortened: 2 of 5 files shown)


ElLorans:
Thanks for the PR! Looks like you used python 3.10 or above so tests on 3.9 are failing because of the pipe operator `|` . In python 3.9, you must use `typing.Union`. Do you think we can try to solve the issue without an additional argument by detecting local os and cloud os?

ElLorans:
From a quick google search, `posixpath.normpath(base_path)` looks promising.
Can you try something like this, without using a new `on_windows` parameter?
```
def is_in_folder(self, base_path: str, directory: T_PATH_LIKE) -> bool:
        """
        Verify that `directory` is in `base_path` folder
        :param base_path:
            Base directory path
        :param directory:
            Directory path to check
        """
        base_path = posixpath.normpath(base_path)
        directory = posixpath.normpath(directory)
        return directory.startswith(base_path)
```

samialfattani:
> From a quick google search, `posixpath.normpath(base_path)` looks promising. 
> 


as i said , the `os.path.normpath()`  returns path with `'/'` or `'\\'` based on the current os no mater what the os of s3 cloud is !! if we assume both of them are Windows or both of them are Linux, _then no need for this PR_! The test fails only if:

current os (Windows) with s3 cloud (Linux)
or
current os (Linux) with s3 cloud (Windows)

in these cases we should initiate the `S3FileAdmin()` object specifying what is the operating system is used on the clould. I think such information is hard to retrive automatically unless if `boto3` client support such function.

 posixpath.normpath() returns the path in Unix style even if the os is Windows which is not what we are looking for!

what we need is letting s3FileAdmin view works based on s3 cloud's os while other local FileAdmin  views works based on the local os which can be defined automatically by `platform.system()`

ElLorans:
I didn't know you could have Windows on S3. Do you have any docs about it? My understanding **was** that S3 has only one path convention, so posixpath would solve the issue since it returns always the same result from both Windows and Linux

samialfattani:
i don't know either,  i am just asumming worst case senario.

however, [this article](https://aws.amazon.com/blogs/storage/how-to-create-smb-file-shares-with-aws-storage-gateway-using-hyper-v/) shows that it is possible using what so called FSx which is something different from S3 as  i believe.


Anyway, on_windows parameter Must be added. even if S3 uses posixpath then we have to tell `BaseFileAdmin` to behave differently. Another option, we can just add extra check of the `storage` type in the `BaseFileAdmin.__init__() ` where it sets `self._on_windows` to False if the storage type is S3Storage

samialfattani:
keep in mind that the changes of normpath() are made on `BaseFileAdmin` not on `S3FileAdmin` class which make it mandatory to specify on_windows explicitly since the project might have 2 FileAdmin views one to browse local folder and the other for browsing s3 storage.

samialfattani:
i have made all changes we discussed before and it is ready to test now, could you please move this PR on so that i can upload another related PRs. those other PRs are ready but i need this one to be merged first.

samuelhwilliams:
Thanks for the contributions @samialfattani.

I believe this PR is doing two things:

1. Fixing a bug when Flask-Admin is running on a Windows machine and S3 FileAdmin is in use
2. Adding some 'prefix' functionality to FileAdmin

Is that right? If so please can we split these two things out? I think this PR should focus on the windows bug.

I don't think we need the `on_windows` parameter - Flask-Admin should be able to automatically detect this and do the correct thing, rather than us making the user configure FileAdmin correctly? Is that right?

samialfattani:
yes you are right , i mixed up two changes in one PR, my fault !! 
the question is how can i separate them into 2 PRs , should just closed this one and raise another two new PRs?

ElLorans:
> yes you are right , i mixed up two changes in one PR, my fault !! the question is how can i separate them into 2 PRs , should just closed this one and raise another two new PRs?

As you prefer, I think you can use this one as basis for the first one and then open a second one.

samialfattani:
since this branch is opened before your release 2.0.0 then it is better to close this one and open 2 branches

PR #2666, opened 2025-10-02
fix Menu icon

### LINKED ISSUE #1695: category icon
I can't find how to add an icon alongside a menu category.

Try with 
```python
category_icon_classes = dict(config='fa fa-screwdriver')
admin = Admin(..., category_icon_classes=category_icon_classes)
admin.add_view(DefaultView(models.Item, session, category='config', name='Items'))
```
and 
```python
admin = Admin(...)
menucat = MenuCategory(name='config', icon_type='fa', icon_value='fa-screwdriver')
admin._menu_categories['config'] = menucat
admin._menu.append(menucat)
admin.add_view(DefaultView(models.Item, session, category='config', name='Items'))
```
The item view is in the category menu, but no icon on the category menu

### LINKED ISSUE #1800: menu with category_icon_classes bug
`
category_icon_classes = {"System": "glyphicon glyphicon-wrench"}
   
admin = Admin(
template_mode="bootstrap3",
category_icon_classes=category_icon_classes)`
screen shot 2019-01-25 at 5 21 02 pm

The related code is:
https://github.com/flask-admin/flask-admin/blob/280c76b5f62fed98bbb91fa5f7a1f050262a0370/flask_admin/templates/bootstrap3/admin/layout.html#L27

samialfattani:
1. Fix category menu icons, now by setting category_icon_classes on Admin object it will show the icon properly and it is no longer duplicated. 
2. fix Menu item classes for all menu-item types including links.

fixes #1800 
fixes #1695 

* **Tests and Docs:** 
  - Test code is placed in `examples/bootstrap4/`,  Without this change, running the example shows duplicated icon for any category and css classes for most of menu items is not getting the effect.
  - No need to update any docs.

* **Final Look:** 
image


examples/bootstrap4/main.py
```diff
@@ -51,6 +58,9 @@ class UserAdmin(CustomView):
     column_filters = ("name", "email")
     can_export = True
     export_types = ["csv", "xlsx"]
+    can_set_page_size = True
```
ElLorans, line 61: why is this required? Just to show more functionalities, or is there a specific goal?

examples/bootstrap4/main.py
```diff
@@ -185,12 +195,72 @@ def build_sample_db():
...
+        CustomView(
+            Page,
+            db.session,
+            name="Page-with-icon",
+            endpoint="page2",
+            menu_class_name="text-danger",
+            menu_icon_type="fa",
+            menu_icon_value="fa-file",
+        )
+    )
```
ElLorans, line 222:
```suggestion
    views = [
        (
            UserAdmin,
            User,
            {
                "category": "Menu",
                "menu_icon_type": "fa",
                "menu_icon_value": "fa-users",
                "menu_class_name": "text-warning",
            },
        ),
        (CustomView, Page, {"category": "Menu"}),
        (
            CustomView,
            Page,
            {
                "name": "Page-with-icon",
                "endpoint": "page2",
                "menu_class_name": "text-danger",
                "menu_icon_type": "fa",
                "menu_icon_value": "fa-file",
            },
        ),
    ]

    for cls, model, kwargs in views:
        admin.add_view(cls(model, db.session, **kwargs))
```

ElLorans: Probably better to keep as it is, I wanted to see if a for loop could simplify

examples/bootstrap4/main.py
```diff
@@ -185,12 +195,72 @@ def build_sample_db():
...
+        MenuLink(
+            name="External link",
+            url="http://www.example.com/",
+            category="Links",
+            class_name="text-success",
+        )
+    )
+    admin.add_link(
+        MenuLink(name="External link", url="http://www.example.com/", category="Links")
+    )
```
ElLorans, line 259:
```suggestion
    admin.add_sub_category(name="Links", parent_name="Menu")

    links = [
        {
            "name": "link1",
            "url": "http://www.example.com/",
            "class_name": "text-warning bg-danger",
            "icon_type": "fa",
            "icon_value": "fa-external-link",
        },
        {
            "name": "link2",
            "url": "http://www.example.com/",
            "class_name": "text-danger",
        },
        {"name": "Link3", "url": "http://www.example.com/"},
        {
            "name": "External link",
            "url": "http://www.example.com/",
            "category": "Links",
            "class_name": "text-info",
            "icon_type": "fa",
            "icon_value": "fa-external-link",
        },
        {
            "name": "External link",
            "url": "http://www.example.com/",
            "category": "Links",
            "class_name": "text-success",
        },
        {
            "name": "External link",
            "url": "http://www.example.com/",
            "category": "Links",
        },
    ]

    # Loop over links
    for link in links:
        admin.add_link(MenuLink(**link))
```

ElLorans: same as above, probably not worth it. But I would definitely move the `add_sub_category` to the end or beginning of this code bloc. Otherwise it gets "hidden" in all the `add_link`

ElLorans:
I can reproduce both the issue and the fix in the example, thanks! 2 minor considerations: the example is now much more complex (probably fine); we do not have a test case. @samuelhwilliams, would it make sense to add a simple test checking the HTML?

examples/bootstrap4/main.py
samuelhwilliams: Yeah, I don't like this change, it makes it harder to read and probably decreases the ability of intellisense-like things to understand track correct use of paramd.

examples/bootstrap4/main.py
samialfattani: yes, just to show more functionality of pagination since the whole example is about bootstrap theme

examples/bootstrap4/main.py
samuelhwilliams: Let's keep this as it was - the for loop doesn't improve the code here IMO

examples/bootstrap4/main.py
samialfattani: ok boss 👍

samuelhwilliams:
Thanks for the fix @samialfattani! Really appreciate it. I'm probably not too stressed about a unit test for this, but one might be nice if you have the time 🙇

samialfattani:
my pleasure, nice to hear that. Actually making a succesful unit test in this case seems to be challenging, even if we make sure that the HTML icon code is placed properly but still we don't know if the icon is really visible in the browser or not. For example check this icon that is placed in many examples in `/exmples/`
`
<a class="btn btn-primary" href="/"><i class="glyphicon glyphicon-chevron-left"></i> Back</a>
`
if you run the example you won't see the icon in the button.

I'm pretty sure that Glyphicon has some problem. However i'm thinking to make another PR adopting Bootstrap Icons as an alternative of Glyphicon

samuelhwilliams:
fair enough! Thanks for thinking it out. Happy to accept :)

PR #2667, opened 2025-10-03
Bootstrap icons

samialfattani:
Currently, [_Glyphicons_ ](https://www.glyphicons.com/) shows unstable behavior in most cases of the whole project where sometimes icons are just no longer visible. To the best of my knowleadge, glyphicon is not for free any more except for a very limited package. Thus,  here in this PR, i'm suggesting to use [Bootstrap Icons ](https://icons.getbootstrap.com/) where it is totally free and compatible with Bootstrap theme.

This PR fixes all disappeared icons that are written with glyphicon and they are replaced with Bootstrap ones.

* **Test and Docs**
* Test code is placed in `/examples/bootstrap4` showing variety of bootstrap usage.
* all comments docs are changed to mention the availabitlity of Bootstrap option.

**Note:** In some parts of the code, glyphicon is already prefixed with Fontawesome icons css classes, so that the icon can be visible properly. I never change those cases since the icon is looking just fine. However _I suggest to deprecate Glyphicon_ at all and keep Fontawesome and Bootstrap

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -1,83 +1,21 @@
 Changelog
 =========
 
-2.0.0rc2
---------
-
-* Serbian translation updates
-
-2.0.0rc1
---------
-
-Fixes:
-* SQLAlchemy boolean filters now convert "1" and "0" to Python booleans, which fixes a type coercion issue with psycopg(3).
-
-
-2.0.0a5
--------
-
-New functionality:
-* Mongonengine support has been added back (no longer using `flask-mongoengine`): https://github.com/pallets-eco/flask-admin/pull/2611
-* The default query for pymongo can be overridden: https://github.com/pallets-eco/flask-admin/pull/2661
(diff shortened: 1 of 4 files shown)


samialfattani:
my PR is ready to test , why there is no action yet from your side friends!

samuelhwilliams:
I will try to look at this over the weekend.

samuelhwilliams:
There are still a lot of references across the codebase to glyphicons. I think it would be good to address this issue fully, so if glyphicons are no longer free and our implementation of them is poor, and bootstrap icons are free and available, we should just switch to those completely.

Do you have capacity to do this?

samialfattani:
yes sure , i will submit another commit within a few days

ElLorans:
Why did you close this one?

samialfattani:
after i merged master into this branch it started failing tests with different versions of python 3.10, 3.11...etc. , i will checkout to older commit in master and create a new branch for bootstrap_icons and then make another PR. unless you have another suggestion i'm happy to hear from you.

PR #2669, opened 2025-10-09
Support python 3.14

samuelhwilliams:


ElLorans:
Testing this locally looks painfully slow since numpy and psycopg do not have wheels for py3.14 yet

ElLorans:
> Testing this locally looks painfully slow since numpy and psycopg do not have wheels for py3.14 yet

And yet, I see the precompiled wheels here: https://pypi.org/project/numpy/#files . There must be a mismatch between my OS and the whl, will try to understand why. Psycopg2 does not provide wheels for linux unless you use https://pypi.org/project/psycopg2-binary/#files which is is a practical choice for development and testing but in production it is advised to use the package built from sources. The smart solution would be to test locally without doing a force reinstall.

PR #2670, opened 2025-10-09
Drop SQLAlchemy 1.x support

samuelhwilliams:
With the aim to release Flask-Admin v2 shortly, now is a reasonable time to drop some out-dated dependencies to help improve maintainability into the future. SQLAlchemy v2 was released in [January 2023](https://www.sqlalchemy.org/blog/2023/01/26/sqlalchemy-2.0.0-released/), 2.5ish years ago.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -29,10 +29,6 @@ jobs:
             tox: 'py39-min'
           - python: '3.13'
             tox: 'py313-noflaskbabel'
-          - python: '3.9'
-            tox: 'py39-sqlalchemy1'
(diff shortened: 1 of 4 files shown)


ElLorans:
This would seriously prevent a lot of legacy apps (which is unfortunately, I fear, the majority of Flask apps), to update to v2. I would postpone this to a v3.

samuelhwilliams:
Yeah, I'm inclined to agree. For some reason I thought sqlalchemy 2 had been out a lot longer 🤔 I don't think it's bad to be a bit more conservative here and do it for v3.

PR #2673, opened 2025-10-14
update translation references

ElLorans:
Paves the work for cleaner transl PRs


samuelhwilliams:
We probably will need to add a CI check that works out if translations are (meaningfully) out-of-date so that we don't gradually fall out of sync as new PRs come in and tweak words. Tracking line number references will be so noisy that I'm not sure if it's super valuable unless some text has actually changed as well?

ElLorans:
This PR is only because otherwise a PR for translations brings a lot of changes even when only the references did. This one is definitely superseded by the one for Python 3.9. I will close it.

PR #2690, opened 2025-10-31
Replace Glyphicons with Bootstrap icons

samialfattani:
Currently, [_Glyphicons_ ](https://www.glyphicons.com/) shows unstable behavior in most cases of the whole project where sometimes icons are just no longer visible. To the best of my knowleadge, glyphicon is not for free any more except for a very limited package. Thus,  here in this PR, i'm suggesting to use [Bootstrap Icons ](https://icons.getbootstrap.com/) where it is totally free and compatible with Bootstrap theme.

This PR fixes all disappeared icons that are written with glyphicon and they are replaced with Bootstrap ones. Moreover, Glyph icons is completely removed from the project.

* **Test and Docs**
* Test code is placed in `/examples/bootstrap4` showing variety of bootstrap usage.
* all comments docs are changed to mention the availabitlity of Bootstrap option.

**Note:** In some parts of the code, glyphicon is already prefixed with Fontawesome icons css classes, so that the icon can be visible properly. I never change those cases since the icon is looking just fine. However _I suggest to deprecate Glyphicon_ at all and keep Fontawesome and Bootstrap


samialfattani:
what do we need to merge this PR?

samuelhwilliams:
This changes doesn't seem like a backwards-compatible one (somewhat at my request, but this initially came in at an awkward time near the v2 release so I was hoping it might get in for that, but didn't).

Someone in pallets-eco probably needs to take the time to understand the problem and back up the approach here, and then work out what kind of release we schedule this for (v2.0.2 / v2.1 / v3).

I haven't had time to do that yet I'm afraid; very happy for anyone else to do so.

PR #2685, opened 2025-10-20
python 3.14

ElLorans:
fixed typo pointing to https://github.com/pallets-eco/flask-sqlalchemy/issues/137 in place of https://github.com/pallets-eco/flask-sqlalchemy/issues/1379 . Added a new exception to filter a db open error from flask-sqlalchemy. Had an issue building shapely 2 locally, can anyone try to relock the uv.lock?

PR #2691, opened 2025-10-31
add example of form rules

samialfattani:
I have spent long time looking for the right way of using form rules; current examples are not clear enough to cover all types of form rules. Even the AI could not give a correct example of using Container rule! However, This PR providing a very good example of from rules with supported sippet codes.

Now people can follow clear examples of all types of form rules.
- fix the documentation 

* **Test and Docs**
* Test code is placed in `/tests/sqla/test_form_rules.py` showing variety of all typles of rules. Confirmed that each rule is tested in a separate test case.
* docs section is changed including all types of form rules with supported examples.

### DIFF
--- a/flask_admin/tests/test_base.py
+++ b/flask_admin/tests/test_base.py
@@ -245,6 +245,7 @@ def test_add_views(admin):
     assert len(admin.menu()) == 3
 
 
+@pytest.mark.filterwarnings("ignore:unclosed file:ResourceWarning")
 def test_add_category(admin):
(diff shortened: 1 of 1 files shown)


ElLorans:
You are facing a very similar error of https://github.com/pallets-eco/flask-admin/pull/2685.

samialfattani:
> You are facing a very similar error of #2685.

could you guide me please how can i fix this? i read #2685 but i could not find out how to resove the issue.

ElLorans:
I have no clue, it looks to me like a bug in a dependency or in Python

PR #2697, opened 2025-11-03
test examples

### LINKED ISSUE #2470: Add a test that runs all of the examples
A number of our example apps have had typos and small bugs in them, preventing them from actually running.

We should add a test/some tests that run all of the examples to make sure that they actually work (with their associated requirements files).

A single tox environment to run all of the examples would probably be good. In the first instance, just testing that the apps all run and a page is available on `localhost:5000/admin` is probably good enough. We could firm them up from there.

This test should dynamically collect all packages in the `examples` directory so that if any new examples are added, they are tested without having to change any config files anywhere (ie to explicitly add the new example to the test suite).

ElLorans:
fixes #2470

### DIFF
--- a/flask_admin/tests/test_examples.py
+++ b/flask_admin/tests/test_examples.py
@@ -1,0 +1,81 @@
+"""
+Test suite for flask-admin examples.
+
+This test suite validates that each example in the examples directory
+can be started successfully and that the /admin endpoint returns HTTP 200.
+
+Each example is run using the env in examples/<example>/pyproject.toml.
+"""
+
+import subprocess
+import time
+from pathlib import Path
+
+import pytest
+import requests  # type: ignore
+
+EXAMPLES_DIR = Path(__file__).parent.parent.parent / "examples"
+
+
+def get_example_directories() -> list[tuple[str, Path]]:
+    """
+    Get all subdirectories in the examples directory that contain a main.py file.
+
+    Returns:
+        List of tuples (example_name, example_path)
+    """
+    if not EXAMPLES_DIR.exists():
+        return []
+
+    examples = []
+    for item in EXAMPLES_DIR.iterdir():
+        if item.is_dir():
+            main_py = item / "main.py"
+            if main_py.exists():
+                examples.append((item.name, item))
+
+    return sorted(examples)
+
+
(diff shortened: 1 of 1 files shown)


ElLorans:
Locally the new tests are much faster, not sure how to tackle this

samuelhwilliams:
6 hours 😂  😭  yeah something doesn't feel quite right ha. Not sure when I can help out with investigating this but can try to 👍

ElLorans:
> 6 hours 😂 😭 yeah something doesn't feel quite right ha. Not sure when I can help out with investigating this but can try to 👍

locally most tests pass (I still have to verify that they do not if I add an error to the examples), probably `subprocess.Popen` and `requests.get` are not supported by github actions. I need to find a new way, I probably just overnengineered it and I can simply build a dynamic test matrix with tox.

samialfattani:
I have tried an approach like this but i could not make it to succesfful run, have a look it might be an idea. Assuming that all examples should have create_app() function that returns the Flask app object.


```
@pytest.mark.parametrize("example_name,example_path", get_example_directories())
def test_example_runs(example_name: str, example_path: Path):
    """
    Test that the example can be started and that the /admin endpoint returns HTTP 200.

    Args:
        example_name: Name of the example.
        example_path: Path to the example directory.
    """

    def run_tester(app):
        try:
            # Wait for the server to start
            time.sleep(2)

            Check if the /admin endpoint is reachable
            response = requests.get("http://localhost:5000/admin")
            assert (
                response.status_code == 200
            ), f"/admin endpoint returned {response.status_code}"

        finally:
            app.sutdown()
            print(f"Example '{example_name}' ran successfully.")

    sys.path.append(str(example_path))
    from main import create_app

    app = create_app()

    import threading

    thread = threading.Thread(target=run_tester, args=(app), daemon=True)
    thread.start()

    # app.run(debug=True)
```

ElLorans:
Unfortunately, this does not work because each example needs its own env

hasansezertasan:
> Unfortunately, this does not work because each example needs its own env

WDYT about https://github.com/pallets-eco/flask-admin/issues/2470#issuecomment-4168471034?

PR #2706, opened 2025-11-12
add page title + icon

samialfattani:
This PR enabels developer to override the page title at the begining of each view 

image

- tested pages: `model/list.html, model/edit.html, model/create.html, model/details.html, fileadmin/list.html`
- `page_title` block is added to docs so that it can be overriden by child templates.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -259,6 +259,8 @@ ignore = [
     "UP007",  # Use X | Y for type annotations; doesn't work with some of our (stringy) type definitions - should aim to resolve this
 ]
 
+ignore-init-module-imports = true
+
 [tool.ruff.lint.isort]
 force-single-line = true
 order-by-type = false


.vscode/settings.json
ElLorans, line 1: I think this should be gitignored, shouldn't it?

samialfattani: yes you are right, just added it to the PR by mistake, is it ok to add it to gitignore in this PR?

examples/bootstrap4/data.py
```diff
@@ -0,0 +1,129 @@
...
+        "Davis",
+        "Davies",
+        "Rodriguez",
+        "Cox",
+        "Alexander",
+    ]
+
+    for i in range(len(first_names)):
+        user = User()
+        user.name = first_names[i] + " " + last_names[i]
```
ElLorans, line 66:
```suggestion
        user.name = f"{first_names[i]} {last_names[i]]"
```

ElLorans:
If you solve the conflicts, I am happy to merge

pyproject.toml
```diff
@@ -258,7 +258,6 @@ select = [
 ignore = [
     "UP007",  # Use X | Y for type annotations; doesn't work with some of our (stringy) type definitions - should aim to resolve this
 ]
-ignore-init-module-imports = true
```
ElLorans, line 261: sorry for being picky, but this should go in a separate PR

samialfattani:
done ✅

PR #2708, opened 2025-11-14
Fix/popperjs map

samialfattani:
`popper.min.js.map` file supposed to be placed beside the `popper.min.js` file so that it avoid the error of `404` to be raised from time to time.

source: https://getbootstrap.com/docs/4.2/getting-started/download/


ElLorans:
I cannot replicate the error nor I find the documentation about it with a very quick search, can you help me understand?

samialfattani:
> I cannot replicate the error nor I find the documentation about it with a very quick search, can you help me understand?

actually, replicating the error is a bit awkward. you need to use the error_handler and put a break point to catch the error of 404. However, the error happens when ever I repeatedly navigate between 2 heavy view pages in a quick way. It seems like this error happens when you navigate to another view while the current page is not fully loadded!

ElLorans:
If you need a break point to get the error, does it mean this happens only in dev?

Nicholas-Read:
Hello, we too are seeing errors related to this.

I might be being a fool here but the popper [included in flask admin ](https://github.com/pallets-eco/flask-admin/blob/master/flask_admin/static/bootstrap/bootstrap4/js/popper.min.js#L5) includes `sourceMappingURL=popper.min.js.map`, but that file doesn't seem to exist in the repo at all?

It's very likely I am being an idiot though; I am not a front end developer.

ElLorans:
> Hello, we too are seeing errors related to this.
> 
> 
> 
> I might be being a fool here but the popper [included in flask admin ](https://github.com/pallets-eco/flask-admin/blob/master/flask_admin/static/bootstrap/bootstrap4/js/popper.min.js#L5) includes `sourceMappingURL=popper.min.js.map`, but that file doesn't seem to exist in the repo at all?
> 
> 
> 
> It's very likely I am being an idiot though; I am not a front end developer.

Can you test now and see if it's fixed?

PR #2715, opened 2025-12-12
remove onclick from all tags

### LINKED ISSUE #2556: Convert onclick scripts event listeners
There are a number of `onclick` attributes that have inline javascript: https://github.com/search?q=repo%3Apallets-eco%2Fflask-admin%20onclick&type=code

These don't support CSP nonces and so may be blocked in applications that apply strict CSP rules.

We should migrate all of the `onclick` attributes to event listeners set up in some JS files.

samialfattani:
Fixes #2556


samialfattani:
it shows this error

```
    assert rv.status_code == 200
E   assert 404 == 200
E    +  where 404 = <WrapperTestResponse streamed [404 NOT FOUND]>.status_code
________ test_csp[ModalFileAdmin rename dir-modalfa-rename/?path=dir1] _________
flask_admin/tests/csp/test_csp.py:273: in test_csp
    assert rv.status_code == 200
E   assert 404 == 200
E    +  where 404 = <WrapperTestResponse streamed [404 NOT FOUND]>.status_code
```

whie it is locally passes the test successfullly, !!! any suggestion ?

ElLorans:
I silenced the warning in master, can you pull the latest changes?

flask_admin/templates/bootstrap4/admin/file/list.html
```diff
@@ -192,4 +194,21 @@
                          actions,
                          actions_confirmation) }}
   <script {{ admin_csp_nonce_attribute }} src="{{ admin_static.url(filename='admin/js/bs4_modal.js', v='1.0.0') }}"></script>
+
+
+  <script {{ admin_csp_nonce_attribute }}>
+
+    $('.btn-delete-file').on('click', function (e)
```
ElLorans: Can we use `data-msg` as a selector and define only 1 function?

flask_admin/templates/bootstrap4/admin/model/inline_list_base.html
```diff
@@ -40,6 +40,20 @@
...
+        data-fieldid="{{ field.id }}"
+        >{{ _gettext('Add') }} {{ field.label.text }}</a>
+
+    <script {{ admin_csp_nonce_attribute }} >
+
+      document.addEventListener('DOMContentLoaded', function ()
+      {
+        $(`#{{ field.id }}-button`).on('click', function (e)
+        {
+          //e.preventDefault();
```
ElLorans: don't we need `preventDefault`?

flask_admin/templates/bootstrap4/admin/model/row_actions.html
```diff
@@ -31,8 +31,24 @@
...
 </form>
+
+<script {{ admin_csp_nonce_attribute }} >
+
+  document.addEventListener('DOMContentLoaded', function ()
+  {
+      $('#btn-delete-{{ row_id }}').on('click', function (e)
+      {
+          const msg = $(this).data('msg');
+          return faHelpers.safeConfirm(msg);
```
ElLorans, line 49: why faHelpers.safeConfirm vs confirm as above? Is it possible to consolidate the 2 snippets?

flask_admin/templates/bootstrap4/admin/actions.html
```diff
@@ -5,12 +5,25 @@
...
+
+    <script {{ admin_csp_nonce_attribute }} >
+
+      document.addEventListener('DOMContentLoaded', function ()
+      {
+          $('.p-action').on('click', function (e)
+          {
+              e.preventDefault();
+              var action = $(this).data('action');
+              modelActions.execute(action);
```
ElLorans, line 20: is this intrinsically unsafe? an attacker could change the HTML

flask_admin/tests/csp/test_csp.py
```diff
@@ -0,0 +1,228 @@
...
+        assert tag.attrs["nonce"] == nonce
+
+    styles = soup.select("style")
+    assert len(styles) == 0
+    for tag in styles:
+        assert tag.attrs["nonce"] == nonce
+
+
+@pytest.fixture
+def db(app):
```
ElLorans: Can we import this and avoid redefinition?

flask_admin/tests/csp/test_csp.py
```diff
@@ -0,0 +1,228 @@
...
+    db = SQLAlchemy(app)
+    yield db
+
+    with app.app_context():
+        db.session.close()
+        db.engine.dispose()
+
+
+@pytest.fixture
+def Model1(app, db):
```
ElLorans: function names should use snake_case

flask_admin/tests/csp/test_csp.py
```diff
@@ -0,0 +1,228 @@
...
+
+def fill_db(db, Model1):
+    objs = [
+        {"test1": "test1_val_1", "test2": "test2_val_1", "bool_field": True},
+        {"test1": "test1_val_2", "test2": "test2_val_2", "bool_field": False},
+        {"test1": "test1_val_3", "test2": "test2_val_3"},
+        {"test1": "test1_val_4", "test2": "test2_val_4"},
+    ]
+
+    db.session.add_all([Model1(**obj) for obj in objs])
```
ElLorans, line 67: Maybe I did not have my coffee yet, but this is a bit confusing. Model1 is a function, isn't it?

flask_admin/templates/bootstrap4/admin/model/inline_list_base.html
samialfattani: Well, it will not harm to add it but  `href="javascript:void(0)"` still can't be removed since removing it will harm the CSS style like this:
image

flask_admin/templates/bootstrap4/admin/model/row_actions.html
samialfattani: tbh, i don't know what is it for? but i always avoid ot change JS codes since we don't have enough tests for JS.

flask_admin/templates/bootstrap4/admin/actions.html
samialfattani: currently the only available action is bulk "delete", even if there are more available actions and the attacker change it, then it is a matter of the backend code that supposed to verify that the submitted action is allowed for the current user.

flask_admin/tests/csp/test_csp.py
samialfattani: We can't, since the conftest.py of the parent folder does not have it. but you have a point. I will create a conftest.py in this folder to be shared for other later test files.

flask_admin/tests/csp/test_csp.py
samialfattani: haha it is a function & class 😄 
actually it is a function returns a class of Model1, since it is a wrapped by `fixture` then the returned class is passed as a parameter on a later function.

ElLorans: I think in other tests we use a simpler approach with a normal function. Do you think it would make sense to use the same style or is this version better?

flask_admin/tests/csp/test_csp.py
samialfattani: db fixture is not in scope
and admin fixture has a special parameter assignment csp_nonce_generator

so both need to be redfined

flask_admin/tests/csp/test_csp.py
samialfattani: i don't know which is better actually, but anyway i will change it into normal calling function to avoid confusing

flask_admin/templates/bootstrap4/admin/file/list.html
samialfattani: do you mean to combine both '.btn-delete-file' and '.btn-delete-dir' in one function ? if so, I don't think this a right thing to do since it's violate the concept of (Separation of Concerns). it more readable and clear to keep them separate functions so that each one can be handled differently

samialfattani:
you can merge now

flask_admin/templates/bootstrap4/admin/file/list.html
ElLorans: don't the 2 functions do the exact same thing?

samialfattani: Yes they do, but my openion is to keep them separate so that it would be easy to handle each action differently in the future. But, i got your point, we  can just combine them now by assigning both Buttons into same css class (e.g. "confirm-first"), and later on, if some contributer wan't to add different action to any of those buttons he can just simply add extra css class and handle it.

flask_admin/templates/bootstrap4/admin/model/row_actions.html
```diff
@@ -31,8 +31,24 @@
   {% elif csrf_token is defined and csrf_token %}
   <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
   {% endif %}
-  <button onclick="return faHelpers.safeConfirm('{{ _gettext('Are you sure you want to delete this record?') }}');" title="{{ _gettext('Delete Record') }}">
+  <button
+    id="btn-delete-{{ row_id }}" class="btn-delete-row"
+    data-msg="{{ _gettext('Are you sure you want to delete this record?') }}, (id= {{ row_id }})"
```
ElLorans: this is an important change that was hard to spot among the commits. Why do we want to show the id number? I fear it might be too technical and would prefer to keep the same functionality as before. If we want to change it, it should go in a different PR.

ElLorans:
I squashed and rebased on top of origin/master to have a cleaner history. If you agree, I/you can remove the `, (id= {{ row_id }})` and we can merge if there is no other change in functionality.

flask_admin/templates/bootstrap4/admin/model/row_actions.html
samialfattani: i just forget to remove it as i put it just for testing pourpose

ElLorans:
This looks ready to merge, but I am not an expert on CSP and would appreciate someone to weigh in.

samialfattani:
any plan to merge this PR?

ElLorans:
Why did we close this one?

Issue #2668, opened 2025-10-09
Mention flask-sqlalchemy-lite as a supported engine


ElLorans:
Flask-Admin does not explicitly support Flask-Sqlalchemy-Lite, but since it basically acts the same as a Sqlalchemy model and we use only the Sqlalchemy features, there is no difference.
We should at a minimum mention that it is possible to use it, and at best add testing and type hints.

samuelhwilliams:
It would be good to officially support; if we do call it out in docs we should definitely add test coverage for it.

Issue #2585, opened 2024-12-25
Doesn't correctly close flask-sqlachemy-lite transactions/sessions 


ddorian:
Hi,

Clone this minimal repo and follow the readme to reproduce https://github.com/ddorian/admin-sqlalite-bug

You will get the following traceback 
[full_traceback.txt](https://github.com/user-attachments/files/18245025/full_traceback.txt)

Environment: it's all in the repo.

- Python version: 3.10
- Flask version: 3.1.0
- Flask-Admin version: https://github.com/pallets-eco/flask-admin/commit/a4e7be8b08c6b95179dc691f9887d96be32b68e7

Regards

ddorian:
An alternative would be to pass a scoped session to the ModelView. Seems to work fine for now.
```python
def get_scoped_session(db: SQLAlchemy, **options):
    """Scoped session for flask-admin.
    See https://github.com/pallets-eco/flask-admin/issues/2585
    """

    def _app_ctx_id() -> int:
        """Get the id of the current Flask application context for the session scope."""
        return id(app_ctx._get_current_object())  # type: ignore[attr-defined]

    scope = options.pop("scopefunc", _app_ctx_id)
    return scoped_session(db.sessionmaker, scope)

```

ElLorans:
flask-sqlachemy-lite is not directly supported, even though I had tried last year  with sqlite in local and it worked. We would need to add the tests and code for flask-sqlalchemy-lite. Maybe you can open a pull request?

PR #2717, opened 2025-12-14
Deprecate ModelView(Model, db.session) in favor of ModelView(Model, db)

ElLorans:
Prepares for https://github.com/pallets-eco/flask-admin/issues/2668

Raise
```
DeprecationWarning("Passing a session object directly is deprecated and will be "
                "removed in version 3.0."
                "Please pass the SQLAlchemy db object instead. "
                "Note: the parameter will be renamed from 'session' to 'db' in version "
                "3.0. "
                "Example: ModelView(User, db) instead of ModelView(User, db.session)"
)
```

Rationale discussed in:
1. https://github.com/pallets-eco/flask-admin/issues/2585
2. https://github.com/pallets-eco/flask-admin/pull/2680

Using db.session directly is considered an anti-pattern and can cause issues when using Flask-SQLAlchemy-Lite.
Passing the db object directly allows:

1. Supporting Flask-SQLAlchemy-Lite alongside Flask-SQLAlchemy in the same codebase
2. Proper session management and scoping without unexpected transaction leaks in Flask-SQLAlchemy

- [x]  Edit examples
- [x] Add tests for db in place of db.session

Next steps (separate PR):
- Add tests for flask-sqlalchemy-lite
- Update docs

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -175,9 +175,10 @@ filterwarnings = [
     # raised by flask-sqlalchemy https://github.com/pallets-eco/flask-sqlalchemy/issues/1379
     "ignore:unclosed database in <sqlite3\\.Connection object at 0x[0-9a-f]+>:ResourceWarning",
 
-    # raised at end of test run
+    # raised at end of test run, maybe by Werkzeug?
     "ignore:unclosed file <_io.FileIO name='/home/runner/work/flask-admin/flask-admin/flask_admin/static/bootstrap/bootstrap4/css/bootstrap.min.css' mode='rb' closefd=True>:ResourceWarning",
 
+    # remove once the deprecated ModelView(Model, db.session) API is fully retired
     "ignore:Passing a session object directly is deprecated.*:DeprecationWarning",
 ]
 
@@ -267,8 +268,6 @@ select = [
 ignore = [
     "UP007",  # Use X | Y for type annotations; doesn't work with some of our (stringy) type definitions - should aim to resolve this
 ]
-
-ignore-init-module-imports = true
 
 [tool.ruff.lint.isort]
 force-single-line = true


ElLorans:
@samialfattani, wanted your and the community's opinion before going forward. Can you review and offer your feedback?

samialfattani:
> @samialfattani, wanted your and the community's opinion before going forward. Can you review and offer your feedback?

based on the checklist above still missing 2 points:
-  Add tests for flask-sqlalchemy-lite
-  Update docs

ElLorans:
Yes, before doing that, what do we think? Is it something that makes sense? Or do we want to support both db and db.model as arguments indefinitely?

samialfattani:
Yes it is much better, clear, readable, and simple to provide only db object for all SQLAlchemy dialects, wiether to be SQL-lite or others in the future. Supporting both db and db.session is a bit confusing for code maintainers.

ElLorans:
Writing the tests for flask-sqa-lite requires a bit of time, especially since I would avoid just duplicating the tests/sqla folder and rather parametrize them to keep code DRY and ensure tests are always run on both libraries. The issue is that then flask-sqa-lite does not support sqlalchemy 1. Since I need some time to find a solution, I would postpone the last 2 milestones of this PR and merge as is. Flask-sqa-lite tests and docs will be updated later.

samialfattani:
the question is why we are still using SQLAlchemy 1 ??

ElLorans:
> the question is why we are still using SQLAlchemy 1 ??

Bevause it still receives updates and we cannot force everybody to update to 2. The impact for us is relatively limited (we have 2 dedicated test runs).

samialfattani:
Congrats for merging but I found that you could've handle the parametrized `db` tests in a better, simpler, cleaner, and easier to scale up way. this can be done using parametrized hook with `pytest_generate_tests` function provided by PyTest [readmore](https://docs.pytest.org/en/stable/how-to/parametrize.html),  like this:

``` python

# --- Database configs ---
DB_CONFIGS = {
    "sqlite": {
        "url": lambda: "sqlite:///",
        "creator": None,
    },
    "postgres": {
        "url": lambda: os.getenv(
            "SQLALCHEMY_DATABASE_URI",
            "postgresql://postgres:xxxxxxxxxxxxxxxx@localhost/flask_admin_test",
        ),
        "creator": None,
    },
    "mssql": {
        "url": lambda: os.getenv(
            "SQLALCHEMY_DATABASE_URI_MSSQL",
            "mssql+pymssql://sa:xxxxxxxxxxxxxxxxxxxx@localhost",
        ),
        "creator": None,
    },
}


# 🧪 Parametrize the DB type at session level → all tests run for each DB
def pytest_generate_tests(metafunc):
    if "db_type" in metafunc.fixturenames:
        metafunc.parametrize("db_type", list(DB_CONFIGS.keys()), scope="function")


@pytest.fixture
def generate_db_url(db_type):
    """Get a fresh DB URL (e.g., new temp SQLite file per session)."""
    return DB_CONFIGS[db_type]["url"]()


# -------------------------------

@pytest.fixture
def db(app, generate_db_url):
    app.config["SQLALCHEMY_DATABASE_URI"] = generate_db_url
    app.config["SQLALCHEMY_ECHO"] = True
    app.config["SQLALCHEMY_TRACK_MODIFICATIONS"] = False
    db = SQLAlchemy(app)
    yield db
    with app.app_context():
        db.session.remove()

@pytest.fixture
def admin(app, babel, db):
    admin = Admin(app)
    yield admin


### --------------- That's it --------------------


``` 

Using this every test function uses `db` paramter will be parametrized with ["SQLite", "Postgres", "MSSQL"]

ElLorans:
Thanks for the insight! But parametrizing the 3 db backends should go in a separate PR, I think. The parametrization I used was to test passing both `ModelView(Model, db.session)` and `ModelView(Model, db)` which was in the PR's scope. Your suggested change (which is great!) is more for https://github.com/pallets-eco/flask-admin/pull/2723, I think.

samialfattani:
> Your suggested change (which is great!) is more for https://github.com/pallets-eco/flask-admin/pull/2723, I think.


yes of course you are right , i'm doing this right now, but i just wanted to share the same snippet of code so you can follow the same approach

PR #2720, opened 2025-12-18
gitignore test files

samialfattani:
ignore some files that can be generated by running the test


ElLorans:
Gitignoring without first deleting the files will have no effect though, am I right?

samialfattani:
> Gitignoring without first deleting the files will have no effect though, am I right?

the files will be ignored wiether they are deleted or not, before or after the test

Issue #2598, opened 2025-02-12
In FileAdmin, if a directory has no date, it should display as empty (not 1970)


pamelafox:
Discussed in https://github.com/pallets-eco/flask-admin/pull/2573

Currently, directories are given a timestamp of 0, so they render as 1970:
![Screenshot 2025-02-11 at 11 58 16 AM](https://github.com/user-attachments/assets/256d2d51-235f-4761-a683-d6f6146a32c3)

It would be more accurate to show their dates as empty, like Azure storage explorer does:
![Screenshot 2025-02-11 at 11 21 29 AM](https://github.com/user-attachments/assets/ded8ee15-94af-4231-ad46-a9eb9530ce64)
(It would be good to know what S3 does, I don't have a current account. But I think the S3 code in flask-admin also sets a timestamp of 0 for directories).

This would be a change to BaseFileAdmin and possibly the templates.

PR #2721, opened 2025-12-19
FileAdmin: show empty date when directory has no date

alexprabhat99:
This PR updates FileAdmin to display an empty value instead of the default 1970-01-01 when a directory is given a timestamp of 0.

Fixes: #2598


alexprabhat99:
@ElLorans Could you please review this PR and share your feedback?
Thank you.

PR #2723, opened 2025-12-19
MSSQL Compatibility

samialfattani:
This PR address the issue that raises whenever the end user navigate to any page using MSSQL database. The MSSQL restricts using LIMIT and OFFSET without sorting, 

- Some test cases are added using the paginations with single and muliple primary keys.
- The MSSQL docker container is added on port 1433.
-

### DIFF
--- a/flask_admin/tests/sqla/test_mssql.py
+++ b/flask_admin/tests/sqla/test_mssql.py
@@ -1,6 +1,6 @@
 import sqlalchemy as sa
 
-from .test_basic import CustomModelView
+from flask_admin.contrib.sqla import ModelView
 
 
 def test_pagination(app, mssql_db, mssql_admin):
@@ -17,13 +17,14 @@ def test_pagination(app, mssql_db, mssql_admin):
         mssql_db.drop_all()
         mssql_db.create_all()
 
-        view3 = CustomModelView(Model3, mssql_db.session, page_size=10)
+        view3 = ModelView(Model3, mssql_db.session)
+        view3.page_size = 10
         view3.column_default_sort = ("test1", False)
         mssql_admin.add_view(view3)
 
-        for i in range(10, 100):
-            m = Model3(id=i, test1=f"test-{i:02d}")
-            mssql_db.session.add(m)
+        models = [Model3(id=i, test1=f"test-{i:02d}") for i in range(10, 100)]
+        mssql_db.session.add_all(models)
+        mssql_db.session.commit()
 
         client = app.test_client()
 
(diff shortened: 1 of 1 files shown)


samialfattani:
tests are failing due to the missing of database conneciton with MSSQL, knowing that it suppposed to configured with `devcontainer` . can any one help and guide me how to activate the MSSQL divcontainer during the test ??

ElLorans:
I silenced the warning in master, can you pull the latest changes?

flask_admin/tests/sqla/test_basic.py
```diff
@@ -449,6 +449,48 @@ def test_list_columns(app, db, admin):
...
+            def __str__(self):
+                return self.test1
+
+        db.create_all()
+
+        # test column_list with a list of strings
+        view = CustomModelView(Model3, db.session, page_size=10)
+        admin.add_view(view)
+
+        for i in range(1, 100):
```
ElLorans: This should be more efficient:
```
models = [Model3(id=i, test1=f"test-{i}") for i in range(1, 100)]
db.session.add_all(models)
```

The fastest option would be:
```
models = [Model3(id=i, test1=f"test-{i}") for i in range(1, 100)]
db.session.bulk_save_objects(models)
```
but then we have to be aware that bulk_save does **not**:
1. Run before_insert / after_insert events
2. autoflush

ElLorans:
Thanks for the great work! I do not see any change to functionality, so what was the issue?

flask_admin/tests/sqla/test_basic.py
samialfattani: i will use `db.session.bulk_save_objects(models)` since we don't wan't any of before_insert / after_insert events

samialfattani:
> Thanks for the great work! I do not see any change to functionality, so what was the issue?

to replicate the error:

1- Create a ModelView that connected to MSSQL Table includes more number of records to be paginated.
2- once the ModelView has no default sorting, the MSSQL will reject using LIMIT and OFFSET without ORDER BY clause.

ElLorans:
But I cannot see where you fixed the issue. Sorry for being dense.

samialfattani:
i can't see this either 😄 , honestly this PR gone throught many branches and local merges until i forget what i've done , maybe the issue is just addressed throught all perviouse PRs. Anyway, we can just consider this PR as enabling the use of MSSQL with a very limited test cases. later on, i will modify the /tests/sqla ALL functions to be parametrized so that they should be tested againest (postgres, mssql , and sqllite)

ElLorans:
Is pymssql required? I do not see any import.
Since the tests are currently passing, I think we can directly remove this PR and parametrize the main db fixture.

samialfattani:
> Is pymssql required? I do not see any import.

Yes it is, SQLAlchmey impots it internally. without  this PR MSSQL test won't pass since no container of MSSQL. this PR added container configuration for both devcontainer and github/workflow

ElLorans:
I would then convert/close this PR to parametrize the db fixture, since the additional test can removed.

samialfattani:
yes sure

PR #2725, opened 2025-12-20
update lock

ElLorans:
Tests on Python 3.14 take a long time to execute since numpy is built from source becausethere are no Numpy 2.2.6 wheels for Python 3.14. Updating numpy version should unlock installing wheels already compiled for Python 3.14 and grant much faster tests.


ElLorans:
Test suite down from 5 to 2 minutes!

PR #2727, opened 2025-12-23
Fix/editable form

### LINKED ISSUE #2653: [feature] Model Listing Page: column shouldn't be editable if `column_formatters`
For example, if `Post` model has a `.body` column (Text/String) which has a lot of data

and we set a formatter for it like this:
```py
class PostModelView(ModelView):
    column_formatters = {
        "body": lambda v, c, m, p: m.body[:40] + "…" if len(m.body) > 40 else m.body,
    }
```
Then, we try to edit it on the model listing page, it will set and display the formatted value, due to [this](https://github.com/pallets-eco/flask-admin/blob/37ee5fbdb17b1f4a5743366512aa140cf1cc580b/flask_admin/model/base.py#L2089):
```py
# flask_admin/model/base.py #L~2089
class BaseModelView(BaseView, ActionsMixin):
    def _get_list_value(...):
        column_fmt = column_formatters.get(name)
        if column_fmt is not None:
            value = column_fmt(self, context, model, name)
            #   ^^^^^^^^^^^^^^
        else:
            value = self._get_field_value(model, name)
        ...
```

As a result, `/ajax/update/` will receive the updated version of the <ins>**formatted value**</ins>, not the actual value (`Post.body`).

<details>
<summary>[Screenshot]</summary>

Image

</details>

### ♻️ Reproducable example:
```py
from flask import Flask
from flask_admin import Admin
from flask_sqlalchemy import SQLAlchemy

app = Flask(__name__)
app.config['SQLALCHEMY_DATABASE_URI'] = 'sqlite:///test.db'
app.config['SECRET_KEY'] = 'secret'
db = SQLAlchemy(app)

# Define a simple model
class Post(db.Model):
    body = db.Column(db.Text)

# Custom ModelView with column formatter
class PostModelView(ModelView):
    column_list = ['body']
    column_editable_list = ['body']  # Enable inline editing
    column_formatters = {
        'body': lambda v, c, m, p: m.body[:40] + '...' if len(m.body) > 40 else m.body
    }

# Set up Flask-Admin
admin = Admin(app)
admin.add_view(PostModelView(Post, db.session))

# Create database and add a test record
with app.app_context():
    db.create_all()
    long_text = "This is a very long text that exceeds forty characters in length."
    db.session.add(Post(body=long_text))
    db.session.commit()

if __name__ == '__main__':
    app.run(debug=True)
```

### Solution?
Possibly, we can either create a new endpoint for accessing a value of a single column unlike `details_view`, or raise an error - forbidding an editable feature on a formattable column.

_or leave your own thoughts!_

samialfattani:
This PR Fixes #2653. After I read the whole discussion, I think the solution is easier than I thought!  simply I have just created a new function `get_unformatted_value` that supposed to read the actual value without any format and then feed this value to the `XEditableWidget`.

i have created a dedicated test that having both column_editable_list and column_formatter for the same field and it seems works fine.

I have modified the example `examples/sqla` to include all datatypes in column_editable_list

### DIFF
--- a/examples/sqla/admin/models.py
+++ b/examples/sqla/admin/models.py
@@ -218,4 +218,4 @@ class Model1(db.Model):  # type: ignore[name-defined, misc]
         return self.test1
 
     def __str__(self):
-        return self.test1
+        return f"{self.name}"
--- a/flask_admin/tests/sqla/test_basic.py
+++ b/flask_admin/tests/sqla/test_basic.py
@@ -854,36 +854,6 @@ def test_column_editable_list(app, db, admin, session_or_db):
         pytest.param("db", id="with_db"),
     ],
 )
-def test_details_view(app, db, admin, session_or_db):
-    with app.app_context():
-        Model1, Model2 = create_models(db)
-
-        param = db if session_or_db == "session" else db.session
-        view_no_details = CustomModelView(Model1, param)
-        admin.add_view(view_no_details)
-
-        # fields are scaffolded
-        view_w_details = CustomModelView(Model2, param, can_view_details=True)
-        admin.add_view(view_w_details)
-
-        # show only specific fields in details w/ column_details_list
-        string_field_view = CustomModelView(
-            Model2,
-            param,
-            can_view_details=True,
-            column_details_list=["string_field"],
-            endpoint="sf_view",
-        )
-        admin.add_view(string_field_view)
-
-
-@pytest.mark.parametrize(
-    "session_or_db",
-    [
-        pytest.param("session", id="with_session_deprecated"),
-        pytest.param("db", id="with_db"),
-    ],
-)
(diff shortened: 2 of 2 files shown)


examples/sqla/admin/main.py
```diff
@@ -194,7 +195,12 @@ class UserAdmin(ModelView):
         "currency",
         "timezone",
     ]
-    column_formatters = {"phone_number": phone_number_formatter}
+
+    column_editable_list = ["last_name"]
```
ElLorans: column_editable_list is already defined, we cannot overwrite it. The same is true for `form_create_rules`

samuelhwilliams:
This PR has a lot of unrelated commits in - I think it needs cleaning up before it should be merged.

PR #2729, opened 2025-12-23
generate URL Filter 

### LINKED ISSUE #2703: Add method to easily generate a pre-filtered URL
I don't think it's trivial currently to generate a URL for an admin page with some filters already applied.

I'd like to be able to do something like:

```python
filtered_url = UserModelView.url_for(search="sam", filters={"age": [IntGreaterFilter(30)]})
```

samialfattani:
Fixes #2703

this PR adds a function to generate a URL with filter arguements including all types of filters (Like, NotLike, Greater....etc) and including both indexed (`flt2_3=xx`) and named (`flt3_email_like=@example.com`)

**Usage:**
```
    filtered_url = view.url_for(
        search="exam",
        filters=[
            filters.FilterLike(Model1.test1, "Email", url_value="@example.com"),
            filters.BooleanEqualFilter(Model1.bool_field, "active", url_value=False),
            filters.FloatGreaterFilter(Model1.test5, "Salary", url_value=15.3),
        ],
    )
```
  
**Output:**
`/admin/user/?search=exam&flt0_0=@example.com&flt1_21=0&flt2_16=15.3`

**OR with named parameters:**
`/admin/user/?search=exam&flt0_test1_contains=@example.com&flt1_bool_field_equals=0&flt4_test2_greater_than=15.3`


flask_admin/contrib/peewee/filters.py
```diff
@@ -32,8 +33,10 @@ def __init__(
             Fixed set of options
         :param data_type:
             Client data type
+        :param url_value:
+            URL value
         """
-        super().__init__(name, options, data_type)
+        super().__init__(column, name, options, data_type, url_value=url_value)
```
ElLorans: is having column as a first argument a breaking change?

flask_admin/contrib/peewee/filters.py
```diff
@@ -128,6 +137,9 @@ def apply(self, query: t.Any, value: t.Any) -> t.Any:
     def operation(self) -> T_TRANSLATABLE:
         return lazy_gettext("not in list")
 
+    def stringify(self, value: t.Any) -> str:
```
ElLorans: Can this method be removed? It is inherited from the parent

flask_admin/contrib/sqla/filters.py
```diff
@@ -163,6 +170,9 @@ def apply(
     def operation(self) -> T_TRANSLATABLE:
         return lazy_gettext("not in list")
 
+    def stringify(self, value: t.Any) -> str:
```
ElLorans: I would remove and inherit from parent

flask_admin/model/base.py
```diff
@@ -1341,7 +1341,7 @@ def get_filters(self) -> list[BaseFilter] | None:
         else:
             return None
 
-    def get_filter_arg(self, index: int, flt: BaseFilter) -> str:
+    def get_filter_arg_name(self, index: int, flt: BaseFilter) -> str:
```
ElLorans, line 1339: this is a breaking change. Can we revert the function name?

flask_admin/model/base.py
```diff
@@ -1384,6 +1384,73 @@ def _get_filter_groups(self) -> OrderedDict[str, FilterGroup] | None:
...
+        Given a filter `flt`, return the unique name for that filter in
+        this view.
+
+        :param flt:
+            Filter instance
+        """
+        if not self._filter_args:
+            return None
+
+        found = None
```
ElLorans: found sounds like a bool, what about renaming this var to `index_flt`?

flask_admin/model/base.py
```diff
@@ -1384,6 +1384,73 @@ def _get_filter_groups(self) -> OrderedDict[str, FilterGroup] | None:
...
+        self, search: str | None = None, filters: list[BaseFilter] | None = None
+    ) -> str:
+        url_args = {}
+        if search:
+            url_args["search"] = search
+
+        if filters is None:
+            filters = []
+
+        for i, flt in enumerate(filters):
```
ElLorans, line 1413: You aimed for code simplicity and separation of concerns (and rightly so), but does this risk to have complexity O(n*n) if filters are not in the same order? I think this search can be done with a set difference

flask_admin/model/filters.py
```diff
@@ -18,14 +18,20 @@ class BaseFilter:
 
     def __init__(
         self,
+        column: t.Any,
```
ElLorans: this is a breaking change

flask_admin/model/filters.py
```diff
@@ -18,14 +18,20 @@ class BaseFilter:
...
         name: str,
         options: T_OPTIONS = None,
         data_type: T_WIDGET_TYPE = None,
         key_name: str | None = None,
+        # FIXME: to be removed in future releases and replaced with `value`
```
ElLorans: why is that a problem?

flask_admin/model/filters.py
```diff
@@ -166,20 +221,33 @@ class BaseFloatListFilter(BaseFilter):
...
 
 class BaseDateFilter(BaseFilter):
     """
     Base Date filter. Uses client-side date picker control.
     """
 
     def __init__(
-        self, name: str, options: T_OPTIONS = None, data_type: T_WIDGET_TYPE = None
+        self,
+        column: t.Any,
```
ElLorans: this is a breaking change

ElLorans:
Thanks for this titanic effort. I will pause reviewing. Is it feasible to remove the breaking changes?

flask_admin/contrib/peewee/filters.py
samialfattani: yes

flask_admin/contrib/sqla/filters.py
samialfattani: ok

flask_admin/model/base.py
samialfattani: after reviewing the code i saw that `get_filter_arg_name` is a private function and it is used only in `BaseModelView` class. I think no harm to rename it. As I mentioned before, the whole filter process needs refactoring and code enhancement sicne it is a very low readability.

flask_admin/model/base.py
samialfattani: yes, maybe it is better to rename it into `filter_arg_key`

flask_admin/model/filters.py
samialfattani: ok i will explain. currently, a parameter `value` is passed to each of `stringify`, `validate` , and `clean` functions, where the new added `url_value` field can be used instead in all of these functions and we no longer need to pass `value` parameter. i.e `value` should be used as a field and can be shared in all functions.
However, i didn't do this in this PR to avoid breaking changes.

flask_admin/model/base.py
samuelhwilliams: get_filter_arg_name is not indicated as private therefore it is part of flask-admin's public interface. Changing the order or name of arguments is a breaking change.

samialfattani: ok i will revert it

samialfattani:
> Thanks for this titanic effort. I will pause reviewing. Is it feasible to remove the breaking changes?

Most welcome, actually i've reviewed the code, however I found that `column` is the first argument in all other Filter sub classes such as `BasePeeweeFilter` and `BasePyMongoFilter`.  that's why i decided to locate `column` to be the first arguement as well in `BaseSQLAFilter` to be consistance with other filters. Yes you are right, it is a breaking change but if I relocate it we will loose the consistancy. So, what do you think ?
`

ElLorans:
> > Thanks for this titanic effort. I will pause reviewing. Is it feasible to remove the breaking changes?
> 
> Most welcome, actually i've reviewed the code, however I found that `column` is the first argument in all other Filter sub classes such as `BasePeeweeFilter` and `BasePyMongoFilter`. that's why i decided to locate `column` to be the first arguement as well in `BaseSQLAFilter` to be consistance with other filters. Yes you are right, it is a breaking change but if I relocate it we will loose the consistancy. So, what do you think ? `

We can plan such a breaking change down the line for 3.0 **but** I would argue that `BasePeeweeFilter` and `BasePyMongoFilter` should be the ones to align to SQLA, not the opposite.

samialfattani:
> > > Thanks for this titanic effort. I will pause reviewing. Is it feasible to remove the breaking changes?
> > 
> > 
> > Most welcome, actually i've reviewed the code, however I found that `column` is the first argument in all other Filter sub classes such as `BasePeeweeFilter` and `BasePyMongoFilter`. that's why i decided to locate `column` to be the first arguement as well in `BaseSQLAFilter` to be consistance with other filters. Yes you are right, it is a breaking change but if I relocate it we will loose the consistancy. So, what do you think ? `
> 
> We can plan such a breaking change down the line for 3.0 **but** I would argue that `BasePeeweeFilter` and `BasePyMongoFilter` should be the ones to align to SQLA, not the opposite.

ok in this case we should move `column`  in SQLA to last argument for now so that we can merge without any break change. then later on we should plan to relocate column in `BasePeeweeFilter` and `BasePyMongoFilter` to match SQLA one.

flask_admin/contrib/peewee/filters.py
samialfattani: ok i will move column parameter

flask_admin/model/filters.py
samialfattani: ok , i will move column paramter

flask_admin/model/filters.py
samialfattani: ok

flask_admin/model/base.py
samialfattani: i don't get it, what do you mean by set difference! anyway i've tried to wrokarround to reduce the complexity but it seems like it won't be a simple code. I'm wondering if you can accept the current algorithm and we raise a new issue that requests to refactor it ? 

**Analysis:** the actual complexity would be O(m*n) where **m:** is number columns to be filtered, and **n:** number of filter classes for the type of each column. In regular cases m would be 1 < m < 50 and n = 9 (at most), so we are talking about 500 iteration in regular cases which is not that harmful.

samialfattani:
i think this PR is ready to review and merge, it's been so long since it is created.

samuelhwilliams:
Hey @samialfattani - agree this has been open a while and would be really useful functionality to get merged.

Can you clean up the commits please (I think effectively squash everything into one?).

I still want to review the interface generally - a single `view.url_for` call feels really good, but something's nagging in the back of my head about the `filters` interface that I still want to unpick a little bit.

+2500/-1000 PR is quite chunky to review so it may take a bit of time to get this lined up.

Really appreciate your efforts/pushing on this though.

samialfattani:
Thanks for beatiful words. I've tried to squash this branch but i think i failed to find the right way to do it. I had to create  a new branch includes the squashed of all commits if that is helping.

here is the new branch #2882

if this works fine with you, you can close this PR.

ElLorans:
Did you have any merge conflicts?

samialfattani:
> Did you have any merge conflicts?

No

samuelhwilliams:
Superceded by https://github.com/pallets-eco/flask-admin/pull/2882

PR #2730, opened 2025-12-23
fix css class generated by column name

### LINKED ISSUE #2207: Joined table column names result in invalid CSS classes in the edit list
Hey,

with a joined model like so:

```
column_list = ("order.created", ...
```

You'll end up with a list view with classes like:

```
<td class="col-order.created"> 20.12.2021 13:35 </td>
```

where the class attribute of the td is an invalid CSS class (and cannot be addressed with CSS for obvious reasons).

This probably applies to all bootstrap themes and several places, but for example in https://github.com/flask-admin/flask-admin/blob/79ad7e4e4e62898f01f582e367322dcaa2ad4e0a/flask_admin/templates/bootstrap3/admin/model/list.html#L86
should be something like (the underscore being my suggestion, could be any valid CSS separator):

```
<th class="column-header col-{{c|replace(".", "_")}}">
```

I'm happy to submit a PR if this is acceptable.

samialfattani:
Fixes #2207

actually it does NOT has any reference on the whole project but it is better to solve the issue so that it can be used properly in the future.

### DIFF
--- a/flask_admin/tests/bootstrap4/conftest.py
+++ b/flask_admin/tests/bootstrap4/conftest.py
@@ -1,19 +1,5 @@
 import pytest
-from flask import Flask
-from flask_admin import Admin
 from flask_sqlalchemy import SQLAlchemy
-
-
-@pytest.fixture(scope="function")
-def app():
-    # Overrides the `app` fixture in `flask_admin/tests/conftest.py` so that the `sqla`
-    # directory/import path is configured as the root path for Flask. This will
-    # cause the `templates` directory here to be used for template resolution.
-    app = Flask(__name__)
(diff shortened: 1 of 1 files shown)


ElLorans:
Thanks! Can you kindly add a test, too? It would be useful for when we finally implement a bootstrap5 theme...

ElLorans:
Is it possible to import the fixtures rather than redefining them?

samialfattani:
Done

ElLorans:
> Done

I still see the fixtures being redefined

samialfattani:
> > Done
> 
> I still see the fixtures being redefined

sorry i forget to remove them, now they are removed except for db() fixture, since it is not available in the parent folder

PR #2732, opened 2025-12-26
Feat/delete btn

### LINKED ISSUE #1327: Delete button on Edit/Detail view
We've been moving our admin backend from an old outdated version of superadmin to the most recent version of flask-admin and started customizing a few things to make life easier on the people using our admin interface. One feature I was surprised to see missing from flask-admin was a delete button on the edit and/or detail view. 

A common use case for this is: someone sends you a link straight to an edit view in the admin. You now want to delete this object. You need to navigate to the list view, then look for the object again to be able to delete it.

For now I've implemented this with a macro that fires a similar event as the delete button in the list view. I'd much rather see an implementation though that has this built into the ModelView. What are your thoughts on this?

I'd be open to overhauling my implementation and build something that can be shipped to everyone. If you agree on the usefulness of this feature.

samialfattani:
Fixes #1327

Tests case are added to make sure button is appeared in both edit and details views with both Modal and non-modal mode

### DIFF
--- a/flask_admin/tests/sqla/test_basic.py
+++ b/flask_admin/tests/sqla/test_basic.py
@@ -3373,7 +3373,7 @@ def test_del_btn_in_edit_and_details(app, db, admin, with_delete, modal):
 
             can_delete = with_delete
             edit_modal = modal
-            details_modal = (modal,)
+            details_modal = modal
 
         # test column_list with a list of strings
         view = MyModelView(


ElLorans:
I think you brought in some commits from the mssql PR. Can you try to remove them?

samialfattani:
i don't know how to remove them , i will just close this PR and raise a new one

PR #2733, opened 2025-12-26
add delete button on details and edit views

### LINKED ISSUE #1327: Delete button on Edit/Detail view

samialfattani:
Fixes #1327

Tests case are added to make sure button is appeared in both edit and details views with both Modal and non-modal mode

### DIFF
--- a/examples/sqla/admin/main.py
+++ b/examples/sqla/admin/main.py
@@ -225,6 +225,9 @@ class PostAdmin(ModelView):
     column_default_sort = ("date", True)
     create_modal = True
     edit_modal = True
+    can_view_details = True
+    details_modal = True
+
     column_sortable_list = [
         "id",
         "title",
--- a/flask_admin/model/base.py
+++ b/flask_admin/model/base.py
@@ -2446,8 +2446,19 @@ class BaseModelView(BaseView, ActionsMixin):
         else:
             template = self.edit_template
 
+        if self.can_delete:
+            delete_form = self.delete_form()
+        else:
+            delete_form = None
+
         return self.render(
-            template, model=model, form=form, form_opts=form_opts, return_url=return_url
+            template,
+            model=model,
+            form=form,
+            form_opts=form_opts,
+            return_url=return_url,
+            delete_form=delete_form,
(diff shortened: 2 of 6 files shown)


ElLorans:
image
Would it make sense to have the Delete button next to the "Cancel" one rather than on the top?

samialfattani:
> Would it make sense to have the Delete button next to the "Cancel" one rather than on the top?

I thought about that, but i found it more clear to have it on the top for many reasons:
1- keep it far away of cancel button to avoid clicking delete by mistake.
2- when i was trying to add the button in Modal pages, it was'nt strait forward and i found some issues with JS.

samuelhwilliams:
I think having the delete button in the bottom row with existing buttons sounds good. We could right align it to create distance. If it doesn't already, the delete process could also require a confirmation step to prevent accidental deletion.

samialfattani:
**waht about details page ?**

image

samialfattani:
**Detials**

image

image



**Edit**

image

image

hasansezertasan:
I think merging #2847 would make this feature more straightforward.

ElLorans:
> I think merging #2847 would make this feature more straightforward.

I don't think so, I think 2847 only affects inline editable on the list view, while this change is for the edit and detail views.

hasansezertasan:
> I don't think so, I think 2847 only affects inline editable on the list view, while this change is for the edit and detail views.

Oh, I actually was talking about HTMX and ease of use that come with it 🤓.

PR #2734, opened 2025-12-30
insert alert messages on create/edit/upload forms

samialfattani:
This PR shows an alert message if a record failed to create/edit/upload

image

image

**Why this PR**
Without this PR, the enduser might get confused after submitting the create/edit/upload form if the proccess is failed. Yes the error desctiption will be presented under each invalid form field, but still the enduser needs a flash message to alert him with the failure of the proccess so that he can investigate the page looking for the invalid field.


ElLorans:
This had to be reverted with https://github.com/pallets-eco/flask-admin/pull/2737 since the err msg was getting flashed even when accessing the edit form for the first time. We need the tests to check also that the msg is not flashed when it shouldn't.

samialfattani:
ok i will fix it with adding extra tests to make sure it is not showing on GET request

PR #2737, opened 2026-01-01
Revert "flash err msg on create/edit/upload forms (#2734)"

ElLorans:
This reverts commit ff320008a76872aa84bc4b4501967d9876be68e1 from https://github.com/pallets-eco/flask-admin/pull/2734
@samialfattani I am reverting since I get a flashed err msg even when I simply edit a form because is_form_submitted returns False with a GET request which is the first one to access the edit form.

```
def is_form_submitted() -> bool:
    """
    Check if current method is PUT or POST
    """
    return bool(request and request.method in ("PUT", "POST"))
```


samialfattani:
great, should i do anything from my side ?

PR #2741, opened 2026-01-10
fix QueryAjaxModelLoader super() call

### LINKED ISSUE #2740: form_ajax_refs fails with mongoengine due to missing loader.name attribute
**Bug:** `form_ajax_refs` fails with mongoengine due to missing `loader.name` attribute

The `QueryAjaxModelLoader.__init__` method calls `super()` without invoking `.__init__(name, options)`, so the parent class `AjaxModelLoader` never sets `self.name`. This causes an `AttributeError` when rendering forms with AJAX reference fields.

---

**Steps to reproduce:**

1. Create a mongoengine model with a `ReferenceField`
2. Configure `form_ajax_refs` on the `ModelView`
3. Attempt to load the create/edit form

**Traceback:**
File "flask_admin/model/widgets.py", line 39, in call
    kwargs.setdefault("data-url", get_url(".ajax_lookup", name=field.loader.name))
AttributeError: 'QueryAjaxModelLoader' object has no attribute 'name'

---

**Expected behavior:**

The form should render correctly with AJAX-enabled reference fields.

---

Environment:

- Python version: 3.11
- Flask version: 3.1.2
- Flask-Admin version: 2.0.2

karpitsky:
Fix `QueryAjaxModelLoader` to properly call parent `__init__` with `name` and `options`.

The `super()` call was missing `.__init__(name, options)`, so `self.name` was never set. This caused an `AttributeError` when the widget tried to access `field.loader.name`:

File "flask_admin/model/widgets.py", line 39, in call
    kwargs.setdefault("data-url", get_url(".ajax_lookup", name=field.loader.name))

---

fixes #2740

### DIFF
--- a/flask_admin/tests/mongoengine/test_basic.py
+++ b/flask_admin/tests/mongoengine/test_basic.py
@@ -6,6 +6,7 @@ from wtforms import form
 
 from flask_admin.contrib.mongoengine import filters
 from flask_admin.contrib.mongoengine import ModelView
+from flask_admin.contrib.mongoengine.ajax import QueryAjaxModelLoader
 
 
 class Test(Document):  # type: ignore[misc]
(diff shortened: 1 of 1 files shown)


ElLorans:
Thanks for the PR! Do you think it would be possible to add a test case?

karpitsky:
Done! Added a test that verifies the parent `__init__` is called properly.

PR #2777, opened 2026-01-29
Allow only Safe HTML avoiding XSS

samialfattani:
This PR adds the functionality of cleaning HTML from mileasouse codes based on a certain policy. This PR uses [nh3](https://nh3.readthedocs.io/) library to clean the untrusted HTML code before it is rendered. The cleaning policy is pre-defined and it can be changed during Admin setup as follows:

```python
  admin = Admin(
    app,
    name="Example",
    theme=Bootstrap4Theme(),
    safe_html_policy = dict(
      tags = {'a', 'b', 'i', 'u', 'em', 'strong', 'p', 'br', 'ul', 'ol', 'li'},
      attributes= {
        'a': {'href', 'title', 'target'},
        '*': {'class'}
      }
    )
  )

```

This PR will:
1. Fix filter names in the filter list to allow safe HTML.
2. Clean any XSS that might be injected in Column headers, Filter names, and Search placeholder.
3. users can utilize this functionlity using jinja filter, so they can write smoething like:

```python
<!-- some_template.html -->
{{ some_untrusted_html_content | make_safe | safe }}
```

image

**Test and Docs are Updated!**

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -49,6 +49,34 @@ make sure it is appropriate for your project's security needs.
 
 If you create any of your own templates for Flask-Admin pages, you will need to inject the CSP nonces yourself as appropriate.
 
+Safe HTML support
+------------------
+To support safe HTML rendering avoiding XSS attacks, Flask-Admin allows you to use `nh3 <https://pypi.org/project/nh3/>`_ to sanitize
+any HTML content before rendering it in your templates. `nh3.clean` method is embedded as a Jinja2 filter named `make_safe`.
+You can use it in your templates as follows::
+
+    {{ some_untrusted_html_content | make_safe | safe }}
+
+However, you can customize the allowed tags and attributes by injecting your own policy during the application setup
+code as follows::
+
+  admin = Admin(
+    app,
+    name="Example",
+    theme=Bootstrap4Theme(),
+    safe_html_policy = dict(
+      tags = {'a', 'b', 'i', 'u', 'em', 'strong', 'p', 'br', 'ul', 'ol', 'li'},
+      attributes= {
+        'a': {'href', 'title', 'target'},
+        '*': {'class'}
+      }
+    )
+  )
+
+please refer to `nh3 documentation <https://nh3.readthedocs.io/en/latest/>`_ for more details on how to
+customize the sanitization policy.
+
+
 Adding Custom Javascript and CSS
 --------------------------------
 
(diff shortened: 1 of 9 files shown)


ElLorans:
Safe does not allow safe HTML, it simply removes autoescaping. I think this PR makes sense, but we need to ensure this does not raise a security risk. (It shouldn't, because I don't think there is a way to let untrusted data there).
EDIT: this comment is now obsolete since the nature of the PR changed

samialfattani:
> Safe does not allow safe HTML, it simply removes autoescaping. I think this PR makes sense, but we need to ensure this does not raise a security risk. (It shouldn't, because I don't think there is a way to let untrusted data there).

I totally agree, I have changed this PR into a security change. please review the PR description

samuelhwilliams:
I don't like the use of `| safe` filters; I'd rather we don't blanket consider things passed in as safe.

I'm not sure this PR is necessary, other than perhaps documenting how to do it (as per below).

Arbitrary HTML can already be included in eg column labels by using a third-party package like `markupsafe`, then doing:

```python
    column_labels = {
        "name": markupsafe.Markup('<i class="fa fa-user"></i> Name'),
        "email": markupsafe.Markup('<i class="fa fa-envelope"></i> Email'),
    }
```

samialfattani:
> Arbitrary HTML can already be included in eg column labels by using a third-party package like `markupsafe`, then doing:

thanks for quick response 👍 
Actually I made this cahnge to avoid using `markupsafe.Markup` since it is risky where it could voilate XSS as it is mentioned on ChkMrx platform report.

image

---

if the column's label is read from an external source (e.g DB) then it might include untrusted HTML as follows:

```python
markupsafe.Markup('<script>alert('XSS!')</script> Email')
```

This PR allows user to safely control the HTML to be rendered on column's label and on any other place within any template. I believe this change gives a valuable security 

> I don't like the use of `| safe` filters; I'd rather we don't blanket consider things passed in as safe.

As you see `| safe ` is used only after `| make_safe ` which makes sure it is totally safe and approved by user.
```html
{{ some_untrusted_html_content | make_safe | safe }}
```

samuelhwilliams:
I think this responsibility should remain our users and we shouldn't try to inject responsibility to flask-admin in this one situation.

Currently flask-admin protects the user by escaping content this content in templates. If someone using our lib wants to inject HTML, especially from an untrusted source, then I think we should leave the responsibility with them to use markupsafe and/or some sanitising library.

I'm not sure this patch justifies adding the new dependency.

ElLorans:
I am also inclined towards closing

samialfattani:
This PR will be closed for the following reasons:

- `safe()` filter is not actually safe 😂
- This PR adds one more dependancy for less benifits.
- The HTML escaping should be responsible by the user, not flask-admin.

PR #2779, opened 2026-02-02
Feat/tooltips

samialfattani:
This PR adds the feature of enabling the Bootstrap [tooltip](https://getbootstrap.com/docs/4.6/components/tooltips/) where it can be inserted within adding a new view like this:

```python
    view = MyModelView(Model1, param, name="Model1", tooltip="tooltip of Model1")
    admin.add_view(view)
``` 

image

#
#
**Moreover**, users can add tooltip anywhere using normal Bootstrap tooltip as follows:

```html
<a title="tooltip message with <b>HTML enabled</b>"
     href="/admin/" 
     data-role="tooltip"
     data-toggle="tooltip" data-placement="bottom" data-html="true"
>
/admin/
</a>
```

**Tests and Docs are updated**


stajilov:
Not bad, looks useful

samialfattani:
no body reviewed it yet 🙄

ElLorans:
Sorry for the late reply, maybe it's me but I do not see a big advantage. This might also bring more chaos while we have so many PRs for a new theme.

samialfattani:
> Sorry for the late reply, maybe it's me but I do not see a big advantage. This might also bring more chaos while we have so many PRs for a new theme.

i don't think so, how could it lead to chaose where the html changes made on the current theme only. if some other themes are going to be added then they can present the tooltip in thier own way. Anyway if you intend to close this PR pleaes let me know so i won't put more effort here nor resolving the conflict.

PR #2781, opened 2026-02-06
Add audit log system example

Neha-1216:
Added an Audit Log feature that tracks all CREATE, UPDATE, and DELETE operations on database models using SQLAlchemy event listeners. Logs are stored in a dedicated AuditLog table and displayed in a read-only admin view with filters for action type and model name.

### DIFF
--- a/examples/sqla/admin/main.py
+++ b/examples/sqla/admin/main.py
@@ -496,6 +496,16 @@ class TreeView(ModelView):
         return super().render(template, foo="bar", **kwargs)
 
 
+def format_action_for_export(view, context, model, name):
+    """Format action enum for export."""
+    return model.action.value if model.action else ""
+
+
+def format_timestamp_for_export(view, context, model, name):
+    """Format timestamp for export."""
+    return model.timestamp.strftime("%Y-%m-%d %H:%M:%S") if model.timestamp else ""
+
+
 class AuditLogAdmin(ModelView):
     """Read-only view for audit log entries."""
     can_create = False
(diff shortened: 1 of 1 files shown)


examples/sqla/admin/static/css/custom-theme.css
```diff
@@ -0,0 +1,504 @@
+/* Flask-Admin Custom Theme - Matching Landing Page Design */
```
ElLorans, line 1: not sure I understand why this file

examples/sqla/admin/main.py
```diff
@@ -23,22 +24,213 @@
...
-<p><a href="/admin/?lang=pt">Click me to get to Admin! (Portuguese)</a></p>
-<p><a href="/admin/?lang=ru">Click me to get to Admin! (Russian)</a></p>
-<p><a href="/admin/?lang=pa">Click me to get to Admin! (Punjabi)</a></p>
-<p><a href="/admin/?lang=zh_CN">Click me to get to Admin! (Chinese - Simplified)</a></p>
-<p>
-  <a href="/admin/?lang=zh_TW">Click me to get to Admin! (Chinese - Traditional)</a>
-</p>
-"""
-    return tmp
+    return """<!DOCTYPE html>
```
ElLorans, line 27: Such a long html+css should be moved to static

ElLorans:
I see you only added an example, so I changed the PR name. I think adding a new example makes more sense than editing the current one. But why aren't you using something like https://github.com/sqlalchemy-continuum/sqlalchemy-continuum? What happens when a model changes?

samuelhwilliams:
Is there anything about this which is specific to Flask-Admin? This seems like a generic example of SQLAlchemy functionality only.

I'm not sure there is much value in adding this.

samuelhwilliams:
This doesn't seem like an example which demonstrates Flask-Admin functionality; this is best left to users to handle as audit systems are complex and this example does leave auditing gaps (eg from raw SQL queries).

PR #2783, opened 2026-02-06
Feat/s3 prefix

samialfattani:
This is a new feature added to the s3 to support the (Prefix) functinoality. just like a normal browsing FileAdmin view starting from a specific folder as a Root path, using this PR, s3 can support such feature and allow a specific Prefix as a Root.

There is no one try add this feature before, i hope this PR will add something valueble

### DIFF
--- a/flask_admin/contrib/fileadmin/s3.py
+++ b/flask_admin/contrib/fileadmin/s3.py
@@ -65,7 +65,9 @@ class S3Storage(BaseFileStorage):
         fileadmin_view = MyS3Admin(storage=S3Storage(...))
     """
 
-    def __init__(self, s3_client: BaseClient, bucket_name: str) -> None:
+    def __init__(
+        self, s3_client: BaseClient, bucket_name: str, prefix: str = ""
+    ) -> None:
         """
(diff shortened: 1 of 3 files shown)


flask_admin/contrib/fileadmin/s3.py
```diff
@@ -85,41 +87,47 @@ def __init__(self, s3_client: BaseClient, bucket_name: str) -> None:
...
+    @_strip_leading_slash_from("directory")
+    def get_files(self, directory: str, path: str) -> list[t.Any]:
+        def _strip_prefix(name: str) -> str:
             if name.startswith(path):
                 return name.replace(path, "", 1)
             return name
 
-        def _remove_trailing_slash(name: str) -> str:
-            return name[:-1]
+        def rcoat(x: str) -> str:
```
ElLorans: should this function be renamed to `_rcoat`?

samialfattani: yes you are right

samialfattani:
it is ready to review now, do i need to do anything from my side ?

PR #2782, opened 2026-02-06
feat: added responsive capabilities for tables, buttons, forms, form …

stajilov:
- Rebuilt the Bootstrap 4 list/filter/form stack for responsive behavior: filters render in flex cards, list tables emit `data-label` attributes, pagination hides extra items on xs screens, and the search bar/tab row flex-wraps with a fluid search control so the “List,” “Create,” “Add Filter,” and search UI stay usable on mobile. Key files: `flask_admin/static/admin/css/bootstrap4/admin.css`, `flask_admin/static/admin/js/bs4_filters.js`, `flask_admin/templates/bootstrap4/admin/model/list.html`, and `flask_admin/templates/bootstrap4/admin/model/layout.html`.
- Modernized form rendering and submit buttons: fields now use a Bootstrap grid (rows/columns) and the submit button row flexes/stacks on small screens to keep Save/Cancel reachable, while offsets remain on medium+ (`flask_admin/templates/bootstrap4/admin/lib.html` plus `.submit-row` styles).
- Added pagination-ready seed data to the Peewee example (50 posts, 20 users) so the list view’s responsive paging/card layout can be exercised locally (`examples/peewee_simple/main.py`).  
- Added CSS helpers ensuring the `.nav-tabs` controls, responsive search form, and filter cards behave across breakpoints.

**Testing**

- ~~`uv run pytest` fails locally because PostgreSQL/Mongo/S3/Azure services are unavailable in this environment; no automated UI/responsive tests exist yet.~~

PR #2787, opened 2026-02-08
upgrade Bootstrap4 to the latest v4.6.2

samialfattani:
- All CSS, JS, Swatch CSS, files have been updated to the latest version of **Bootstrap4 (4.6.2)** based on the recommedation of [Bootstrap4 EOL](https://getbootstrap.com/docs/4.6/end-of-life/) article. 
- `popper.min.js` has been also updated to version **v1.16.1**, which is compatible with BS(v4.6.2)
- Fontawesome kept without any change and it shows no harm.
- You can test the UI look using `/examples/bootstrap4`

**Why This is important;**
- includes patches for XSS vulnerabilities
- bug fixes for enhanced stability
- It might be a good step since #2782 is about to enhance the UI

### DIFF
--- a/examples/datetime_timezone/main.py
+++ b/examples/datetime_timezone/main.py
@@ -110,8 +110,6 @@ class BlogModelView(ModelView):
         "last_edit": "Last Edit (UTC)",
     }
 
-    column_filters = ["last_edit"]
-
 
 if __name__ == "__main__":
     with app.app_context():


ElLorans:
I am checking the files 1 by 1... is there any more effective way to do it? I have almost finished, missing only the swatches. 
You did not update https://github.com/pallets-eco/flask-admin/blob/master/flask_admin/static/vendor/popper.min.js . Do we need to update it or can we remove it?
What about the files in https://github.com/pallets-eco/flask-admin/tree/master/flask_admin/static/vendor/bootstrap4?

samialfattani:
>  I am checking the files 1 by 1... is there any more effective way to do it?

Thanks for your effort, if you give me more details about what kind of checking your are doing, maybe i can help you with more efficient way.

> I have almost finished, missing only the swatches. You did not update https://github.com/pallets-eco/flask-admin/blob/master/flask_admin/static/vendor/popper.min.js . Do we need to update it or can we remove it? 

Yes you can remove it, it  is never mentioned in any of the source code.

> What about the files in `https://github.com/pallets-eco/flask-admin/tree/master/flask_admin/static/vendor/bootstrap4`?

I've checked all of those files and most of them have never been refernced in the whole source code. The only files that referenced in html are:
- `dropdown.js`
- `modal.js`
- `util.js`

TBH. Whenever i use Bootstrap, i never import such files and i don't actually know what are they for? also I don't know where they are downloaded from !.  what do you suggest ?

ElLorans:
I am doing a checksum of the files to verify they match vs source.
I expect bootstrap files not to be in the readme. The fact they aren't does not imply they are not used.

samialfattani:
> I am doing a checksum of the files to verify they match vs source. I expect bootstrap files not to be in the readme. The fact they aren't does not imply they are not used.

I don't understand ?! if the filename never used in any of the project files, then how it is possible to be used ?

samialfattani:
any news ?

ElLorans:
> > I am doing a checksum of the files to verify they match vs source. I expect bootstrap files not to be in the readme. The fact they aren't does not imply they are not used.
> 
> I don't understand ?! if the filename never used in any of the project files, then how it is possible to be used ?

My bad, I had misread
>Yes you can remove it, it is never mentioned in any of the source code.

I need to finish the checksum and then I can test the frontend. Any other mantainer/user doing the latter would help.

flask_admin/static/bootstrap/bootstrap4/css/bootstrap.css
ElLorans, line 1: Do we need the unminified css? Maybe we can continue shipping only the minified version.

samialfattani: actually in the source code we are using the minified version, what i've did is downloaded the original folder as a best practice.

ElLorans: I get that, but I would personally continue shipping only the minified version since I see no advantage in shipping both

samialfattani: in this case then i think we can delete all css and js files and keep only :

`bootstrap.min.css`
`bootstrap.min.css.map`

`bootstrap.min.js`
`bootstrap.min.js.map`

shall we do this ?

ElLorans: If we validate we do not need the .js files, we can remove them

samialfattani: ok, now you can review after i've removed all unrefrenced .css an .js files

princerb:
looks clean to me, but I think it is time to mark Bs4 as deprecated, as it had reached end of life very long ago

ElLorans:
We cannot mark it as deprecated until we have a bootstrap5 version

ElLorans:
daterangepicker does not work anymore. You can test with `examples/datetime_timezone`
EDIT: also, you need to update the v parameter in `.url(filename` to avoid caching and ensure consistency between API and content

samialfattani:
> daterangepicker does not work anymore. You can test with `examples/datetime_timezone` 

i've reviewed the example and i found that daterangepicker works fine!! However if you are refferring to the issue in `create` page: 

image

Then I believe it is because of `extra_js = ["/static/js/timezone.js"]` file in that view model. I have tried the same example in the master branch and the issue still there.

> EDIT: also, you need to update the v parameter in `.url(filename` to avoid caching and ensure consistency between API and content

v parameter is fixed

ElLorans:
I cannot reproduce in master, only here

samialfattani:
> I cannot reproduce in master, only here

this is what i got when i ran the `example/datetime_timezone` on master branch
 
image

Could you please show me a screen shout from master ?

examples/datetime_timezone/main.py
```diff
@@ -110,6 +110,8 @@ class BlogModelView(ModelView):
         "last_edit": "Last Edit (UTC)",
     }
 
+    column_filters = ["last_edit"]
```
ElLorans: Why this? I think we can remove it

ElLorans:
> > I cannot reproduce in master, only here
> 
> this is what i got when i ran the `example/datetime_timezone` on master branch
> 
> image
> Could you please show me a screen shout from master ?

You are right, I opened https://github.com/pallets-eco/flask-admin/pull/2824 to fix it. I was checking create view on this branch and edit view on master.

examples/datetime_timezone/main.py
samialfattani: ok

ElLorans (approved):

ElLorans:
LGTM, I will merge next week. Would be nice if anybody else could test this branch.

PR #2788, opened 2026-02-08
fix get_one `sqla.ModelView`

### LINKED ISSUE #2915: bug(sqla): `get_one()` raises `InvalidRequestError` (HTTP 500) when `id` contains a comma on a single-PK model
# bug(sqla): `get_one()` raises `InvalidRequestError` (HTTP 500) when `id` contains a comma on a single-PK model

## Summary

`ModelView.get_one()` in `flask_admin.contrib.sqla.view` calls
`session.get(self.model, tools.iterdecode(id))` unconditionally. For a model
with a single primary key, passing an `id` query string that contains a comma
(e.g. `?id=1,2`) makes `iterdecode` return a 2-tuple, which SQLAlchemy 2.x
rejects in `Session.get()`. The exception is not caught by `details_view` /
`edit_view`, so it surfaces as an unhandled **HTTP 500**.

This was previously raised in #2788, but the discussion stalled without a
minimal reproducer. Opening a fresh issue with a PEP 723 MRE so the failure
mode is unambiguous, and so the right fix can be discussed separately from
that PR's approach.

## Expected behavior

A malformed `id` parameter on a single-PK model should produce a
user-facing error (e.g. `400 Bad Request` or the existing "Record does not
exist." flash + redirect to the list view) — not an unhandled `500`.

## Actual behavior

```
sqlalchemy.exc.InvalidRequestError: Incorrect number of values in identifier
to formulate primary key for session.get(); primary key columns are 'widget.id'
```

…rendered as a `500 Internal Server Error` response.

## Minimal reproducer (PEP 723)

Save as `repro.py` and run with `uv run repro.py`. No venv setup needed.

```python
# /// script
# requires-python = ">=3.10"
# dependencies = [
#     "flask-admin",
#     "flask-sqlalchemy",
# ]
# ///
"""MRE for flask-admin get_one() crash on `?id=1,2` with a single-PK model."""
from flask import Flask
from flask_admin import Admin
from flask_sqlalchemy import SQLAlchemy

app = Flask(__name__)
app.config["SQLALCHEMY_DATABASE_URI"] = "sqlite:///:memory:"
app.config["SECRET_KEY"] = "x"
app.config["WTF_CSRF_ENABLED"] = False
db = SQLAlchemy(app)


class Widget(db.Model):
    name = db.Column(db.String(50))


class WidgetView(ModelView):
    can_view_details = True


admin = Admin(app)
admin.add_view(WidgetView(Widget, db.session, endpoint="widget"))

with app.app_context():
    db.create_all()
    db.session.add_all([Widget(name="alpha"), Widget(name="beta")])
    db.session.commit()

client = app.test_client()

print("=== id=1 (valid) ===")
rv = client.get("/admin/widget/details/?id=1")
print("status:", rv.status_code)

print("\n=== id=1,2 (malformed for single-PK model) ===")
rv = client.get("/admin/widget/details/?id=1,2")
print("status:", rv.status_code)
print("body head:", rv.data[:120])
```

### Output

```
=== id=1 (valid) ===
status: 200

=== id=1,2 (malformed for single-PK model) ===
status: 500
body head: b'<!doctype html>\n<html lang=en>\n<title>500 Internal Server Error</title>\n<h1>Internal Server Error</h1>...'
```

The same crash is reachable on `/admin/widget/edit/?id=1,2` and any other view
that calls `get_one()`.

## Root cause

`flask_admin/contrib/sqla/view.py`:

```python
def get_one(self, id):
    session = _get_deprecated_session(self.session)
    return session.get(self.model, tools.iterdecode(id))
```

`tools.iterdecode("1,2")` returns `("1", "2")`. SQLAlchemy 2.x's
`Session.get()` validates the identifier length against the model's primary
key columns and raises `InvalidRequestError` when they don't match. Compare
with `get_pk_value()` (same file), which already branches on
`isinstance(self._primary_key, tuple)` for the write side — the read side
should mirror that.

## Notes on PR #2788

The PR proposes a fix but:

1. Uses `tools.escape(id)` for the single-PK branch, which encodes the value
   in the `iterencode` escape format (e.g. `"1,2"` → `"1\,2"`) — a value
   that can never match an integer primary key. The crash is avoided only
   because `session.get` returns `None` and the existing
   "record not found → redirect" path takes over.
2. Contains a dead `_id = tools.iterdecode(id)` line immediately overwritten
   in both branches of the `if/else`.
3. The added test uses `follow_redirects=True` and asserts on strings
   (`test1_val_1`, `test2_val_1`) that appear on the list view after the
   redirect — not on the details page. It would pass even if `get_one`
   returned `None`, so it does not actually verify the multi-id case.

## Environment

- `SQLAlchemy` 2.x.
- Python 3.10+.

```python
def get_one(self, id):
    session = _get_deprecated_session(self.session)
    if isinstance(self._primary_key, tuple):
        return session.get(self.model, tools.iterdecode(id))
    return session.get(self.model, (id,))
```

…and a test that asserts the status code / flash message on the malformed
input rather than scraping the redirected list page.

samialfattani:
A run-time error is raised when `get_one()` is called in `flask_admin.contrib.sqla.view.ModelView`

To replicate the error try to browse this: `/admin/details/?id=1,2`

**Risk:**
A security vulnerability is raised by [Tnable](https://www.tenable.com/) where it could lead to **_Blind SQL Inejection_** which is classified as a High risk vulnerability. This can be exploited in `/edit` and `/details` 

image

A test case is added to make sure that  `/admin/details/?id=1,2` is working as expected

### DIFF
--- a/flask_admin/contrib/sqla/view.py
+++ b/flask_admin/contrib/sqla/view.py
@@ -1336,7 +1336,12 @@ class ModelView(BaseModelView):
             Model id
         """
         session = _get_deprecated_session(self.session)
-        return session.get(self.model, tools.iterdecode(id))
+        if isinstance(self._primary_key, tuple):
+            _id = tools.iterdecode(id)
+        else:
+            _id = tools.escape(id)
+
+        return session.get(self.model, _id)
 
     # Error handler
     def handle_view_exception(self, exc: Exception) -> bool:
--- a/flask_admin/contrib/sqla/view.py
+++ b/flask_admin/contrib/sqla/view.py
@@ -1336,13 +1336,11 @@ class ModelView(BaseModelView):
             Model id
         """
         session = _get_deprecated_session(self.session)
-        _id = tools.iterdecode(id)
-        if isinstance(self._primary_key, tuple):
-            _id = tools.iterdecode(id)
-        else:
-            _id = (tools.escape(id),)
 
-        return session.get(self.model, _id)
+        if isinstance(self._primary_key, tuple):
+            return session.get(self.model, tools.iterdecode(id))
+
+        return session.get(self.model, (id,))
 
     # Error handler
     def handle_view_exception(self, exc: Exception) -> bool:
--- a/flask_admin/contrib/sqla/view.py
+++ b/flask_admin/contrib/sqla/view.py
@@ -1336,10 +1336,11 @@ class ModelView(BaseModelView):
             Model id
         """
         session = _get_deprecated_session(self.session)
+        _id = tools.iterdecode(id)
         if isinstance(self._primary_key, tuple):
             _id = tools.iterdecode(id)
         else:
-            _id = tools.escape(id)
+            _id = (tools.escape(id),)
 
         return session.get(self.model, _id)
 
--- a/flask_admin/tests/sqla/test_basic.py
+++ b/flask_admin/tests/sqla/test_basic.py
@@ -851,6 +851,16 @@ def test_details_view(
         assert "test2_val_1" in data
         assert "test1_val_1" in data
 
+        # test single-PK with multiple IDs in query string
+        rv = client.get(
+            "/admin/model2/details/?url=%2Fadmin%2Fmodel2%2F&id=1,2",
+            follow_redirects=True,
+        )
+        data = rv.data.decode("utf-8")
+        assert "String Field" in data
+        assert "test2_val_1" in data
+        assert "test1_val_1" in data
+
         # test column_details_list
         rv = client.get("/admin/sf_view/details/?url=%2Fadmin%2Fsf_view%2F&id=1")
         data = rv.data.decode("utf-8")
--- a/flask_admin/tests/sqla/test_basic.py
+++ b/flask_admin/tests/sqla/test_basic.py
@@ -851,22 +851,61 @@ def test_details_view(
         assert "test2_val_1" in data
         assert "test1_val_1" in data
 
-        # test single-PK with multiple IDs in query string
-        rv = client.get(
-            "/admin/model2/details/?url=%2Fadmin%2Fmodel2%2F&id=1,2",
-            follow_redirects=True,
-        )
-        data = rv.data.decode("utf-8")
-        assert "String Field" in data
-        assert "test2_val_1" in data
-        assert "test1_val_1" in data
-
         # test column_details_list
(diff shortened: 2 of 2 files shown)


ElLorans:
The value is passed to SQLAlchemy’s Session.get() using bound parameters, not interpolated SQL, so the security vulnerability is a false positive.
Passing 1,2 on a model with int primary key returns a 400 response, which is correct, so I am inclined towards closing this PR.

flask_admin/contrib/sqla/view.py
```diff
@@ -1329,7 +1329,13 @@ def get_one(self, id):
             Model id
         """
         session = _get_deprecated_session(self.session)
-        return session.get(self.model, tools.iterdecode(id))
+        _id = tools.iterdecode(id)
+        if isinstance(self._primary_key, tuple):
+            _id = tools.iterdecode(id)
+        else:
+            _id = (tools.escape(id),)
```
ElLorans: Why tools.escape?
Looks unnecessary to me

samialfattani: hmmm, what if the actual string-single-PK="1,2", is it possible to navigate `/admin/details?id=1,2` without escaping ?

samialfattani:
> The value is passed to SQLAlchemy’s Session.get() using bound parameters, not interpolated SQL, so the security vulnerability is a false positive. Passing 1,2 on a model with int primary key returns a 400 response, which is correct, so I am inclined towards closing this PR.

Actually it returns **500!**, here is what we got without this PR

image

flask_admin/contrib/sqla/view.py
ElLorans: yes, by passing another time iterdecode. Here you don't need to escape, but to undo the escaping needed to build the url.

samialfattani: what do you mean by build the url ?

ElLorans: The url that is linked in a view e.g. the one for editing will be edit/?id=1,2 but 1,2 will be escaped. You can check it by clicking on the pencil icon of a list view

samialfattani: ok, yes, it will be escaped, but what if the user  wrote `edit/?id=1,2` manually in the browser ? then 500 will be returned with error.

hasansezertasan:
> Actually it returns **500!**, here is what we got without this PR

Can you provide a MRE please?

samialfattani:
Fixes #2915

hasansezertasan (approved):

PR #2794, opened 2026-02-14
prevent XSS in renaming dirs/files

### LINKED ISSUE #1506: Stored Self XSS
When deleting a directory from the User Files menu, a confirmation prompt is displayed to the user:

`<button onclick="return confirm('Are you sure you want to delete <DIRNAME> recursively?')">`

However it is possible to create a directory with a malicious name as shown below:

```
POST /admin/filemanagerview/mkdir/ HTTP/1.1
Host: <IP>
User-Agent: Mozilla/5.0 (Windows NT 10.0; WOW64; rv:54.0) Gecko/20100101 Firefox/54.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Accept-Language: en-GB,en;q=0.5
Referer: https://<IP>/admin/filemanagerview/mkdir/
Content-Type: multipart/form-data; boundary=---------------------------32515312489068
Content-Length: 461
Cookie: <cookiename>=6e3fece9-fbbb-aaaa-bc6b-2d244560bcd0._NEuT9uTPYeJlW50K0bUIgv_LV4
DNT: 1
Connection: close
Upgrade-Insecure-Requests: 1

-----------------------------32515312489068
Content-Disposition: form-data; name="csrf_token"

ImZhNjRlMjVhZGEwNWYxOTQ0MmVlMTZiMjUwZDVmYmE0MjY4NzVkOTIi.DDZgnA.pDRqS3loHGsVzhRheKNA9KeO_Pg
-----------------------------32515312489068
Content-Disposition: form-data; name="path"


-----------------------------32515312489068
Content-Disposition: form-data; name="name"

XSSfolder'-alert(document.domain)-'
-----------------------------32515312489068--
```

This will render as...

`<button onclick="return confirm('Are you sure you want to delete \&#39;XSSfolder&#39;-alert(document.domain)-&#39;\&#39; recursively?')">`

...due to the apostrophe character being encoded as HTML entity, which allows to inject inline JavaScript.

Likely self-XSS only unless CSRF protection is disabled.

samialfattani:
Fixes #1506

This PR to prevent XSS security issue on renaming dirs or files

### DIFF
--- a/flask_admin/contrib/fileadmin/__init__.py
+++ b/flask_admin/contrib/fileadmin/__init__.py
@@ -2,7 +2,6 @@ import os
 import os.path as op
 import platform
 import posixpath
-import re
 import shutil
 import sys
 import typing as t
@@ -480,13 +479,6 @@ class BaseFileAdmin(BaseView, ActionsMixin):
         Override to implement customized behavior.
         """
 
-        def validate_name(self: type[form.BaseForm], field: Field) -> None:
-            regexp = re.compile(
-                r"^(?!^(PRN|AUX|CLOCK\$|NUL|CON|COM\d|LPT\d|\..*)(\..+)?$)[^\x00-\x1f\\?*:\";|/]+$"
-            )
-            if not regexp.match(field.data):
-                raise validators.ValidationError(gettext("Invalid name"))
-
         class NameForm(self.form_base_class):  # type: ignore[name-defined, misc]
             """
             Form with a filename input field.
@@ -496,7 +488,8 @@ class BaseFileAdmin(BaseView, ActionsMixin):
 
             name = fields.StringField(
                 lazy_gettext("Name"),
-                validators=[validators.InputRequired(), validate_name],
+                validators=[validators.InputRequired()],
+                filters=[lambda x: secure_filename(x) if x else None],
             )
             path = fields.HiddenField()
 
@@ -1030,6 +1023,7 @@ class BaseFileAdmin(BaseView, ActionsMixin):
             return redirect(self._get_dir_url(".index_view"))
 
         form = self.upload_form()
+
         if self.validate_form(form):
             try:
                 self._save_form_files(directory, path, form)
--- a/flask_admin/contrib/fileadmin/azure.py
+++ b/flask_admin/contrib/fileadmin/azure.py
@@ -271,6 +271,9 @@ class AzureStorage(BaseFileStorage):
         src = t.cast(str, self._ensure_blob_path(src))
         dst = t.cast(str, self._ensure_blob_path(dst))
 
+        if src == dst:
+            return
+
         if self.is_dir(src):
             self._rename_directory(src, dst)
         else:
--- a/flask_admin/contrib/fileadmin/s3.py
+++ b/flask_admin/contrib/fileadmin/s3.py
@@ -241,6 +241,9 @@ class S3Storage(BaseFileStorage):
     @_strip_leading_slash_from("src")
     @_strip_leading_slash_from("dst")
     def rename_path(self, src: str, dst: str) -> None:
+        if src == dst:
+            return
+
         if self.is_dir(src):
             self._check_empty_directory(src)
             src += self.separator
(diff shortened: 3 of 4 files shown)


ElLorans (approved):

ElLorans:
LGTM, will wait a couple of days before merging in case somebody wants to chime in.

samuelhwilliams (approved):
I don't think this does any damage, but I don't think it's fair to say it _fixes_ #1506 really. If there's any bad data already in someone's system then this won't help them.

We should make sure that we treat filenames as dangerous input everywhere they're rendered, and make sure they're escaped appropriately.

If #1506 is talking about [this code](https://github.com/pallets-eco/flask-admin/blob/b10a3e58cf111636d0791d277371c71f36f0de6c/flask_admin/templates/bootstrap4/admin/file/list.html#L92), then we should run it through Jinja2's escape filter or similar.

For some reason I thought that would be on by default but maybe it's not?

I think happy to merge this, but not to mark 1506 as fixed.

ElLorans:
Thanks for the input. @samialfattani , can you move to using something like [werkzeug.secure_filename](https://tedboy.github.io/flask/generated/werkzeug.secure_filename.html#werkzeug-secure-filename) ? Hopefully you can replace a part of your code.

ElLorans:
Sorry for not being clear, but I was thinking of replacing the regex with a check between the input and the result of secure_filename. Replacing the validator with `secure_filename` like you are doing represents a breaking change since you are silently renaming the file. Do we think this is a problem?

samialfattani:
> Sorry for not being clear, but I was thinking of replacing the regex with a check between the input and the result of secure_filename. Replacing the validator with `secure_filename` like you are doing represents a breaking change since you are silently renaming the file. Do we thing this is a problem?

oh !! i spend a lot of time to replace the regex i hope this will not go for nothing. anyway i don't think that renaming the filename or directory is a breaking change since this technique is actually used in /rename and other places in the project. that's why i thought it is better to use the same way.

flask_admin/contrib/fileadmin/__init__.py
```diff
@@ -1022,8 +1022,11 @@ def upload(self, path: str | None = None) -> T_RESPONSE | str:
             return redirect(self._get_dir_url(".index_view"))
 
         form = self.upload_form()
-        if form.upload.data and form.upload.data.filename:
-            form.upload.data.filename = secure_filename(form.upload.data.filename)  # type: ignore[attr-defined]
+        fname = getattr(
+            getattr(getattr(form, "upload", None), "data", None), "filename", None
+        )
+        if fname:
+            form.upload.data.filename = secure_filename(fname)  # type: ignore[attr-defined]
```
samuelhwilliams: I think this might be more cleanly done using `secure_filename` as a filter on the form?

samialfattani: yes i agree, it is changed now

flask_admin/contrib/fileadmin/azure.py
```diff
@@ -271,6 +271,9 @@ def rename_path(
         src = t.cast(str, self._ensure_blob_path(src))
         dst = t.cast(str, self._ensure_blob_path(dst))
 
+        if src == dst:
```
ElLorans, line 274: Why?

samialfattani:
any thing i can do from my side ?

flask_admin/contrib/fileadmin/azure.py
samialfattani: no need to procceed the renaming if the new name is equal to the old one. i ran the test on S3 and it shows an error if i try to rename an object to same name of it slef

`
 Failed to rename: Failed to rename path: An error occurred (InvalidRequest) when calling the CopyObject operation: This copy request is illegal because it is trying to copy an object to itself without changing the object&#39;s metadata, storage class, website redirect location or encryption attributes
`

i could not run the test on Azure locally but i believe it will make similar error.

PR #2828, opened 2026-03-11
chore [examples]: move non-app.run() locs out of if __name__ == "__main__" guard

samialfattani:
This PR just to relocate the running gurard `if __name__ == "__main__":` into the right place.

**Reason of this PR:** the running guard has to be only for  `app.run()` so that the server won't be executed twice while the other code should be out of the `if` statement. It is useful to run the example with the python debugger using VScode or other tool, however in such cases, the server will be run implicitly (using `flask run` command) without the line of `app.run()`

### DIFF
--- a/examples/auth/main.py
+++ b/examples/auth/main.py
@@ -185,9 +185,10 @@ admin.add_view(MyModelView(User, db))
 
 app_dir = os.path.realpath(os.path.dirname(__file__))
 database_path = os.path.join(app_dir, app.config["DATABASE_FILE"])
-if not os.path.exists(database_path):
-    with app.app_context():
-        build_sample_db()
 
 if __name__ == "__main__":
+    if not os.path.exists(database_path):
+        with app.app_context():
+            build_sample_db()
+
     app.run(debug=True)
--- a/examples/auth/main.py
+++ b/examples/auth/main.py
@@ -180,14 +180,14 @@ def build_sample_db():
     return
 
 
+admin.add_view(MyModelView(Role, db))
+admin.add_view(MyModelView(User, db))
+
+app_dir = os.path.realpath(os.path.dirname(__file__))
+database_path = os.path.join(app_dir, app.config["DATABASE_FILE"])
+if not os.path.exists(database_path):
+    with app.app_context():
+        build_sample_db()
+
 if __name__ == "__main__":
-    admin.add_view(MyModelView(Role, db))
-    admin.add_view(MyModelView(User, db))
-
-    app_dir = os.path.realpath(os.path.dirname(__file__))
-    database_path = os.path.join(app_dir, app.config["DATABASE_FILE"])
-    if not os.path.exists(database_path):
-        with app.app_context():
-            build_sample_db()
-
     app.run(debug=True)
--- a/examples/auth/main.py
+++ b/examples/auth/main.py
@@ -186,9 +186,10 @@ admin.add_view(MyModelView(User, db))
 app_dir = os.path.realpath(os.path.dirname(__file__))
 database_path = os.path.join(app_dir, app.config["DATABASE_FILE"])
 
-if __name__ == "__main__":
-    if not os.path.exists(database_path):
-        with app.app_context():
-            build_sample_db()
+if not os.path.exists(database_path):
+    with app.app_context():
+        build_sample_db()
 
+
+if __name__ == "__main__":
     app.run(debug=True)
--- a/examples/auth_flask_login/main.py
+++ b/examples/auth_flask_login/main.py
@@ -272,20 +272,21 @@ def build_sample_db():
     db.session.commit()
 
 
+admin = Admin(
+    app,
+    name="Example: Auth",
+    index_view=MyAdminIndexView(),
+    theme=Bootstrap4Theme(base_template="my_master.html", fluid=True),
+)
+
+admin.add_view(MyModelView(User, db))
+
+app_dir = os.path.realpath(os.path.dirname(__file__))
+database_path = os.path.join(app_dir, app.config["DATABASE_FILE"])
+if not os.path.exists(database_path):
+    with app.app_context():
+        build_sample_db()
+
+
 if __name__ == "__main__":
-    admin = Admin(
-        app,
-        name="Example: Auth",
-        index_view=MyAdminIndexView(),
-        theme=Bootstrap4Theme(base_template="my_master.html", fluid=True),
-    )
-
-    admin.add_view(MyModelView(User, db))
-
-    app_dir = os.path.realpath(os.path.dirname(__file__))
-    database_path = os.path.join(app_dir, app.config["DATABASE_FILE"])
-    if not os.path.exists(database_path):
-        with app.app_context():
-            build_sample_db()
-
     app.run(debug=True)
--- a/examples/auth_flask_login/main.py
+++ b/examples/auth_flask_login/main.py
@@ -284,10 +284,10 @@ admin.add_view(MyModelView(User, db))
 app_dir = os.path.realpath(os.path.dirname(__file__))
 database_path = os.path.join(app_dir, app.config["DATABASE_FILE"])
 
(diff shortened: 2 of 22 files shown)


ElLorans (dismissed):

ElLorans:
Shouldn't the test container LOCs be under the guard? You would not want to run those.
EDIT: same for build db functions

samialfattani:
> Shouldn't the test container LOCs be under the guard? You would not want to run those. EDIT: same for build db functions

hmmm, actaully i want to run all LOCs including test containers and db, that because the user should have Docker installed in the local machine. however, to the best of my knowledge, the gurad is for only `app.run()`, if we don't want to run any other LOCs we should look for another mechanism.

it would be better if you give a certain case that we need to run those LOCs under the gurad.

ElLorans:
The guard is for anything you don't want to be run when you import the file. I think testcontainers set up is part of that

samialfattani:
> The guard is for anything you don't want to be run when you import the file. I think testcontainers set up is part of that

I still don't get it, when we import the file definetly i want to run the testcontainers LOCs so that the container can be run on Docker !!

ElLorans:
> > The guard is for anything you don't want to be run when you import the file. I think testcontainers set up is part of that
> 
> I still don't get it, when we import the file definetly i want to run the testcontainers LOCs so that the container can be run on Docker !!

no, because maybe you want to run the main.py with a production grade server and you don't need to run docker since you have the right environment. With your PR, running docker is now a side effect when you import the file.
If you wanted to run it when you import the file, you should wrap it into a function and call it yourself after importing the function.

examples/auth/main.py
```diff
@@ -180,14 +180,14 @@ def build_sample_db():
...
-    admin.add_view(MyModelView(User, db))
+admin.add_view(MyModelView(Role, db))
+admin.add_view(MyModelView(User, db))
 
-    app_dir = os.path.realpath(os.path.dirname(__file__))
-    database_path = os.path.join(app_dir, app.config["DATABASE_FILE"])
-    if not os.path.exists(database_path):
-        with app.app_context():
-            build_sample_db()
+app_dir = os.path.realpath(os.path.dirname(__file__))
```
ElLorans, line 186: Db generation must be under guard, too

examples/auth_flask_login/main.py
```diff
@@ -272,20 +272,21 @@ def build_sample_db():
...
 
-    admin.add_view(MyModelView(User, db))
+admin.add_view(MyModelView(User, db))
 
-    app_dir = os.path.realpath(os.path.dirname(__file__))
-    database_path = os.path.join(app_dir, app.config["DATABASE_FILE"])
-    if not os.path.exists(database_path):
-        with app.app_context():
-            build_sample_db()
+app_dir = os.path.realpath(os.path.dirname(__file__))
```
ElLorans, line 284: db generation should be under guard, too

examples/babel/main.py
```diff
@@ -78,12 +78,13 @@ def index():
...
 
-if __name__ == "__main__":
-    # admin.locale_selector(get_locale)
-    admin.add_view(ModelView(User, db))
-    admin.add_view(ModelView(Post, db))
+# admin.locale_selector(get_locale)
+admin.add_view(ModelView(User, db))
+admin.add_view(ModelView(Post, db))
+
+with app.app_context():
```
ElLorans, line 87: Should be under guard, too

examples/auth_flask_login/main.py
samialfattani: ok

ElLorans:
> It is useful to run the example with the python debugger using VScode or other tool, however in such cases, the server will be run implicitly (using `flask run` command) without the line of `app.run()`

Now I read better your requirements. Would the python debugger work with testcontainers? If yes, I see your point in moving testcontainers out of the guard.

samialfattani:
yes it works fine , in that since i think even database creation should be out of th guard

ElLorans:
if using the python debugger is your objective, ok. But FYI the best practice would be different

samialfattani:
OK , you can review now

examples/geo_alchemy/main.py
```diff
@@ -66,39 +66,44 @@ class OSMModelView(ModelView):
...
-        admin.add_view(LeafletModelView(Polygon, db, category="Polygons"))
-        admin.add_view(OSMModelView(MultiPolygon, db, category="Polygons"))
-        admin.add_view(LeafletModelView(LineString, db, category="Lines"))
-        admin.add_view(OSMModelView(MultiLineString, db, category="Lines"))
-
-        with app.app_context():
-            db.create_all()
-
-        app.run(debug=True)
+    app.run(debug=True)
```
ElLorans: does this work? It's out of the with PostgresContainer block

examples/mongoengine/main.py
```diff
@@ -178,12 +178,12 @@ def create_example_data():
...
 if __name__ == "__main__":
-    with MongoDbContainer("mongo:7.0.7") as mongo:
-        mongo_uri = mongo.get_connection_url()
-        connect(host=mongo_uri)
-        create_example_data()
-        admin.add_view(UserView(User, "User"))
-        admin.add_view(TweetView(Tweet, "Tweets"))
-
-        app.run(debug=True)
+    app.run(debug=True)
```
ElLorans: same as above, does this work?

examples/pymongo_simple/main.py
```diff
@@ -107,12 +107,12 @@ def index():
...
+    db = conn.test
 
-        admin.add_view(UserView(db.user, "User"))
-        admin.add_view(TweetView(db.tweet, "Tweets"))
+    admin.add_view(UserView(db.user, "User"))
+    admin.add_view(TweetView(db.tweet, "Tweets"))
 
-        app.run(debug=True)
+if __name__ == "__main__":
+    app.run(debug=True)
```
ElLorans: same as above, does the app run out of the MongoDbContainer?

samialfattani: no , i will fix it

examples/sqla/main.py
```diff
@@ -3,14 +3,45 @@
...
-    if not os.path.exists(database_path):
-        with app.app_context():
-            build_sample_db()
+app_dir = op.join(op.realpath(os.path.dirname(__file__)), "admin")
+database_path = op.join(app_dir, app.config["DATABASE_FILE"])
+if not os.path.exists(database_path):
+    with app.app_context():
+        build_sample_db()
+
+app.jinja_env.undefined = StrictUndefined
```
ElLorans: why?

samialfattani:
now it is ready to review

examples/geo_alchemy/main.py
```diff
@@ -74,29 +101,10 @@ class OSMModelView(ModelView):
         password="password",
         dbname="main",
     ) as postgres:
-        app = Flask(__name__)
-        app.config["SECRET_KEY"] = "secret"
+        print(f"Using database URL: {app.config['SQLALCHEMY_DATABASE_URI']}")
```
ElLorans, line 104: this looks like it would raise an error: the config for this key is defined in the next line

ElLorans:
Many examples now perform db destruction and initialization or depend on runtime resources during module import (datetime_timezone, peewee_simple, geo_alchemy, mongoengine).

There are also unrelated formatting/logging changes and now examples/sqla/main.py does not have StrictUndefined anymore, which is fine but outside this PR scope.

Could we limit this PR to the guard relocation and revisit the examples that require testcontainers or destructive setup?

If you need the db initalization to happen when you run the example with the python debugger using VScode or other tool and therefore need it to happen outside the __main__ guard, I would suggest something like
```
if not os.path.exists(database_path):
    build_sample_db()
```

For examples using testcontainers, `flask run --debug` may import the application multiple times during reload cycles. Depending on how Testcontainers behaves, we may get multiple container startups or other surprises. Can you check this and revisit?

Issue #2831, opened 2026-03-14
flask-sqlalchemy-lite session handling


kesmik:
Hello,

I hope it is a correct section as a bug, but maybe it is my user error, but any help is appreciated
I am testing **flask-admin** with **flask-sqlalchemy-lite** and noticed two issues, IMO both are related to session handling.
1. When  `MODE = "lite"` is set trying to create "Art" throws an error: sqlalchemy.exc.InvalidRequestError: Object '<Team at 0x736012735490>' is already attached to session '1' (this is '2')
- It could be fixed by uncommenting `form_extra_fields` or `form_args` in `ArtsView`.
- Is this expected?
- If yes which approach I should use?
- Little explanation why it is needed would be awesome.
- It works with full sqlalchemy (`MODE = "anything"`) without `form_extra_fields` or `form_args`)
2. When `USE_CUSTOM_EQUATION = False` (You can see base class, it is just implements custom `__eq__` editing model (adding additioanl teams) throws an error that model already exist"
It happens because flask-admin Unique validator sees objects in different sessions and default object equal check fails here, because form.__obj and obj are loaded separately:
`if not hasattr(form, "_obj") or not form._obj == obj:`

**I have created simple project for easy reproduction, hope someone could check and assist:**
- https://github.com/kesmik/flask-admin-multi-sessions
- clone it and run with `uv run main.py`
- try to create and "Art" and it will throw error that Team already exist in different session
- uncomment either `form_extra_fields` or `form_args` in ArtsView and it works fine
- Try to edit Art (add extra team, not name), Unique validator fails
- Change from `USE_CUSTOM_EQUATION = False` to `USE_CUSTOM_EQUATION = True`, editing works
- if MODE = "anythong except lite" it will run full sqlalchemy and creating/editing works without changing anything

Sadly I am fresh to **sqlalchemy** and **flask-admin**, so little help from my side here, but if anyone points essential problem, I might be able to solve it and provide pull request.

Environment:
- Python version: 3.14.0
- Flask version: v3.1.3
- Flask-Admin version: github version @ 0a64283914d5a49ab68c09491c7da0c47d02a1b8

ElLorans:
Thanks for signalling this. Looks like Flask-SQLAlchemy's db.session is a "scoped session" that persists throughout a single request, while Flask-Admin with Flask-SQLAlchemy-Lite sometimes triggers multiple queries that, in certain configurations, end up using different session instances. 

[...]

PR #2832, opened 2026-03-14
fix: change condition for sqlalchemy-lite deprecated session check

kesmik:
In https://github.com/pallets-eco/flask-admin/issues/2831 I have noticed that `def _warn_session_deprecation(session, warn: bool = True):` condition to check that you are passing sqlalchemy-lite db object instead of db.session actually does not work.

This small fix addresses the issue by modifying that condition. I have decided to be specific and check for T_SESSION.
Maybe codition can be switched to `if T_SQLALCHEMY_LITE is not None and not isinstance(session, T_SQLALCHEMY_LITE ):` which should be fine probably.

### DIFF
--- a/flask_admin/contrib/sqla/_compat.py
+++ b/flask_admin/contrib/sqla/_compat.py
@@ -24,7 +24,7 @@ def _warn_session_deprecation(session, warn: bool = True):
     Raise error if session is from Flask-SQLAlchemy-Lite.
     """
     if not hasattr(session, "session"):
-        if T_SQLALCHEMY_LITE is not None and isinstance(session, T_SQLALCHEMY_LITE):
+        if T_SQLALCHEMY_LITE is not None and isinstance(session, T_SESSION):
             # see::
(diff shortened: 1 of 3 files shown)


kesmik:
Yeah, it broke a lot of tests, maybe worth to fix normally, tried to exclude some in fixture, but it is most likely not proper way to fix those tests

ElLorans:
I think the best approach is simply to add the error message to the ignored ones.

kesmik:
> I think the best approach is simply to add the error message to the ignored ones.

Did I get it right, I should go trough all test functions which are using this combination and add some exception in test itself?
IMO that would be most clean way to mark this, but will require some effort to go trough all of them.

Will try to update this, hopefully I am not breaking anything by skipping some combinations.

ElLorans:
I did it with https://github.com/pallets-eco/flask-admin/pull/2834, but it's still a WIP.

ElLorans:
Superseded by https://github.com/pallets-eco/flask-admin/pull/2834, would you take a look? Thanks for signalling this.

PR #2834, opened 2026-03-15
fix: _warn_session_deprecation not detecting sqla-lite.session

ElLorans:
`def _warn_session_deprecation(session, warn: bool = True):` does not raise error in case of SQLAlchemyLite and db.session.
This PR fixes the function and the test suite.
Since pytest allows filtering out warnings but not errors, tests now skip combination of sqla extension==lite and db.session.

- TODO: 
- [x] - replace all checks with function call to keep tests DRY
- [x] - Add test for _warn_session_deprecation

### DIFF
--- a/flask_admin/tests/conftest.py
+++ b/flask_admin/tests/conftest.py
@@ -6,7 +6,6 @@ from flask import Flask
 from jinja2 import StrictUndefined
 from sqlalchemy import event
 from sqlalchemy.engine import Engine
-from sqlalchemy.orm import clear_mappers
 
 from flask_admin import Admin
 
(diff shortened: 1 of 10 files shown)


kesmik:
- It is really ot possible to pass db.session with SQLAlchemyLite now
- Tests looks reasonable
- removing `_warn_session_deprecation` in some of the classes also looks reasonable, since it is still left in view class
- I am only missing:
> Add test for _warn_session_deprecation

Is it somehow incorporated in tests itself, or maybe i did not understood the idea here, was expecting to. On the other hand it is not mandatory IMO.

At least from my point of view, this looks fine, thank you for reacting so fast and taking over problem from me

ElLorans:
You can find it here: https://github.com/pallets-eco/flask-admin/pull/2834/changes#diff-6cd644a60f803c639b5dd6d409098b28239381cefc336c394df3d61b3c6a92c2R3426 . test_basic.py, row 3426. The diff is too large to be visualized by default.

PR #2835, opened 2026-03-15
Fix/flask sqla lite session handling

### LINKED ISSUE #2831: flask-sqlalchemy-lite session handling

ElLorans:
fixes #2831 .
Depends on https://github.com/pallets-eco/flask-admin/pull/2834
I was not able to reproduce the issue in the test suite, only in the example, so no new tests.

1. Changed Unique validator (tested under flask-sqla and flask-sqla-lite)
2. `_get_deprecated_session` must be called at the last second for flask-sqla-lite, or objects will be bound to different sessions, causing Errors

### DIFF
--- a/flask_admin/contrib/sqla/validators.py
+++ b/flask_admin/contrib/sqla/validators.py
@@ -51,8 +51,12 @@ class Unique:
         DB row are considered equal even when they were loaded by different
         Session instances (e.g. with flask-sqlalchemy-lite).
         """
+        if obj_a == obj_b:
+            return True
+
         if type(obj_a) is not type(obj_b):
             return False
+
         mapper = sa_inspect(type(obj_a))
         pk_attrs = [col.key for col in mapper.primary_key]
         return all(getattr(obj_a, attr) == getattr(obj_b, attr) for attr in pk_attrs)


kesmik:
I have checked and can confirm that my both original issues are solved with this MR.

Maybe worth for more experienced developer to inspect changes regarding session management and object comparison to not introduce anything strange, but from my point of view it looks clean and okay.

PR #2836, opened 2026-03-18
try with tox run paralell

### LINKED ISSUE #2833: make test-in-docker is slow
Running the test suite requires `make test-in-docker` or, to run the entire tox matrix as in CI, `make tox-in-docker`.
test-in-docker aliases `docker compose -f .devcontainer/tests/docker-compose.yaml run --remove-orphans app uv run pytest` which is slow. What is the best way to make test-in-docker faster?  We can keep tox as it is.

samialfattani:
Fixes #2833

### DIFF
--- a/.devcontainer/tests/Dockerfile
+++ b/.devcontainer/tests/Dockerfile
@@ -3,9 +3,11 @@ FROM mcr.microsoft.com/devcontainers/${IMAGE}
 
 ENV UV_PROJECT_ENVIRONMENT=/venv
 
-RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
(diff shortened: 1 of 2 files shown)


ElLorans:
this addresses `make tox-in-docker` not, `make test-in-docker`

samialfattani:
actually, i was just wanted to see if using `tox run-paralell` command can make any difference , but actually not. I will just close this PR.

ElLorans:
that's a shame. What about https://pypi.org/project/pytest-xdist/ or https://pypi.org/project/pytest-parallel/? But my issue is with the env creation rather than the tests per se.

Issue #2682, opened 2025-10-19
Support custom themes better


samuelhwilliams:
I tried to implement a custom theme for Flask-Admin (https://github.com/samuelhwilliams/xgovuk-flask-admin/) and ended up needing to do data manipulation in quite a few places; we should make it easier for themes to worry only about visual/styling issues rather than flask-admin data internals.

princerb:
https://github.com/pallets-eco/flask-admin/discussions/2647#discussioncomment-14001954

hasansezertasan:
Hey @samuelhwilliams, you've got some ideas about custom theme adoption. WDYT about #2847, would it make this task easier?

PR #2643, opened 2025-07-25
[Theme] New theme: FomanticUI [beta]

princerb:
A new theme - FomanticUI.
<details>
  <summary>[Preview]</summary>
  image
</details>

Checklist:
- [x] Core templates: base, actions, layout
- [ ] Models:
  - [x] List
  - [x] Layout
  - [x] Row actions
  - [ ] Editable columns
  - [ ] Edit
  - [ ] Details
  - [ ] Create
  - [ ] Inline forms - edit, details, create
- [x] File Manager
- [ ] Redis Console

Other:
- [ ] Leaflet integration
- [ ] Remove and organize unnecessary js/css libs

PR #2847, opened 2026-03-29
feat: replace x-editable with HTMX for inline editing

### LINKED ISSUE #1615: Replacement for X-editable?
It seems like this plugin hasn't had any development for a while. In addition, it doesn't work for Many-to-Many relationship. 

Are there other alternatives?

hasansezertasan:
## Summary

- Replace the unmaintained x-editable library with HTMX for theme-agnostic inline editing in list views
- Net result: **-702 lines** (249 added, 951 removed)
- The `column_editable_list` API is unchanged — no user-facing breaking changes

Fixes #1615

## Changes

| Component | Before | After |
|-----------|--------|-------|
| Widget | `XEditableWidget` (125 lines, 12+ field type mappings) | `HTMXEditableWidget` (30 lines, field-agnostic) |
| JS | ~70 lines x-editable init code | ~20 lines HTMX `afterSwap` handlers |
| Vendor | x-editable CSS + JS (~670 lines) | HTMX 2.0.8 min.js (14KB gzipped) |
| Endpoints | `POST /ajax/update/` → plain text | `GET /ajax/edit/` (new) + `POST /ajax/update/` → HTML fragments |

### How it works

1. User clicks an editable cell → HTMX sends `GET /ajax/edit/?pk=X&field=Y`
2. Server returns an edit form HTML fragment → HTMX swaps it into the `<td>`
3. User edits and presses Enter → HTMX sends `POST /ajax/update/`
4. Server validates, saves, returns the updated display fragment → HTMX swaps it back
5. Cancel (Escape key or button) restores the original content client-side (no round-trip)

### Why HTMX

- **Theme-agnostic**: No CSS framework dependency — works with Bootstrap 4, Bootstrap 5, Tabler, or any future theme
- **Simpler architecture**: Server renders everything, client just swaps HTML fragments
- **Less JS to maintain**: Field-type logic moves from client-side JS to server-side WTForms rendering
- **Active project**: HTMX is actively maintained vs x-editable which hasn't been updated in years

## Test plan

- [x] All existing SQLAlchemy editable list tests pass (165 passed)
- [x] All existing Peewee editable list tests pass (11 passed)
- [x] New `test_ajax_edit_endpoint` covers: valid field, non-editable field (404), non-existent record (404), no `column_editable_list` (404)
- [x] Validation errors re-render edit form with error markup
- [x] Special primary keys (strings with hyphens) work correctly
- [x] CSRF token conditional guard works with both SecureForm and BaseForm
- [x] No x-editable references remain in production code
- [x] Manual testing with the demo script below

## Manual testing

Run the "sqla_column_editable" example to test inline editing interactively.

### What to test manually

1. **Click to edit**: Click any cell in Name/Email/Active columns → should show inline input
2. **Save**: Change value, press Enter or click ✓ → value updates without page reload
3. **Cancel**: Press Escape or click ✗ → reverts to original value
4. **Boolean field**: Click Active column → should show Yes/No dropdown
5. **Validation**: Try submitting an empty required field → should show error inline


hasansezertasan:
I'd be more into showing a modal for update forms.

ElLorans:
Wow!! You say test passed but how many are testing the actual code changes?

ElLorans:
Also, this looks like a breaking change to me. If that's the case should either release this in 3.0, or provide a variable/parameter to switch.

hasansezertasan:
> Wow!! You say test passed but how many are testing the actual code changes?

Could you please provide more detailed information 🤓?

> Also, this looks like a breaking change to me. If that's the case should either release this in 3.0, or provide a variable/parameter to switch.

I taught the same. The breaking change to me seemded like "XEditableWidget", so I did a quick search for `XEditableWidget` on GitHub but couldn't find any implementor worth mentionable.

Is it possible to determine if it's a breaking change or not?

hasansezertasan:
Some LLM:

---

I did a GitHub-wide search for `XEditableWidget` imports outside of flask-admin — found zero external usage.

The backwards compatibility alias is already in place:

```python
# flask_admin/model/widgets.py:113
XEditableWidget = HTMXEditableWidget
```

So `from flask_admin.model.widgets import XEditableWidget` still works and returns `HTMXEditableWidget`.

The one theoretical breaking case: someone who subclassed `XEditableWidget` and overrode `get_kwargs()` (the old method that handled select/source field-type mappings). That method no longer exists since the new widget is field-agnostic — the server renders the proper input via WTForms. But given zero external implementations found, this seems safe.

The alias can be removed with a deprecation warning in a future major version if desired.

samuelhwilliams:
Comment from the sideline - sorry - none of this is yet me weighing in on whether I support this change or not (or have even understood it yet!).

> I did a GitHub-wide search for XEditableWidget imports outside of flask-admin — found zero external usage.

I did a non-LLM GitHub-wide search for this and I did find some external uses, eg:

- https://github.com/Drc3p0/MakerManager2/blob/ebecd2c8947d049097e8b65df8452ef0f05bae13/application/views/badges.py#L14
- https://github.com/erdal-pb/overhave/blob/19d5050b746491ef73a3c2133a56dd776e83ed00/overhave/admin/views/index/custom_page.py#L5

That said, these codebases haven't been touched in a while.

I think should recognise that this isn't a backwards-compatible change and decide what level of risk tolerance we have for doing this without a full deprecation cycle. This comment isn't meant to steer strongly in either direction. You can decide that we prioritise our own speed in merging this over breaking one or two people, or you can decide that we try to stick with a stricter deprecation policy.

Do you have thoughts on how much overhead we're looking at if we deprecate this through adding a new component and leaving the existing XEditableWidget untouched?

samialfattani:
maybe the right question is, why we are stricting to deprecate something that is already not supported, not functioning well, not compatible for future UIs ? to me,  our users would be happy if we provide them a better, stable, and permenant solution with a little cost of breaking change.
to the best of my knowledge,  `XEditableWidget` depends on JQuery which is inteded to be removed from flask-admin .

samuelhwilliams:
Because _we_ are supporting it by having it in Flask-Admin. When a project takes on a dependency it's an implicit commitment to supporting that for our users, even if it is deprecated upstream. It's not fun for users when projects make breaking changes without giving adequate warning or time to migrate.

In my opinion this is simply the cost of providing stable software for an ecosystem.

I think there's potential to say the benefits of just swapping out directly outweigh the risks/disruption for users, but that isn't a decision to make without _some_ consideration or understanding of the impact.

None of this is to say that I couldn't be convinced that just doing a straight swap here will be 'fine', so consider all of this commentary/conversation rather than edict.

hasansezertasan:
> Comment from the sideline - sorry - none of this is yet me weighing in on whether I support this change or not (or have even understood it yet!).

Actually, the issues I had with x-editable at #2444 motivated me to work on this.

I've used [skycyclone/x-editable](https://github.com/skycyclone/x-editable) over there, but that hasn't received any updates in the last 5 years either.

That got me thinking about alternatives, and I gave HTMX a try — it worked! 🥂 I did have to add some CSS to make it look a bit more polished, though.

I believe this work is a stepping stone toward better custom theme support.

> I think should recognise that this isn't a backwards-compatible change and decide what level of risk tolerance we have for doing this without a full deprecation cycle. This comment isn't meant to steer strongly in either direction. You can decide that we prioritise our own speed in merging this over breaking one or two people, or you can decide that we try to stick with a stricter deprecation policy.

After giving it more thought, I agree with you — this is not a backwards-compatible change. I think we should discuss the deprecation policy and our vision for the user interface further before moving forward.

> Do you have thoughts on how much overhead we're looking at if we deprecate this through adding a new component and leaving the existing XEditableWidget untouched?

The idea that led me to this PR: dropping Bootstrap 4 is overhead in itself.

I think we should talk about this topic thoroughly — what we want to do, where we want to go, and what we want to achieve.

ElLorans:
I think making this not a breaking change is not possible (correct me if I am wrong), but I agree the "breaking" is minor.
Here are the alternatives I see, sorted by my preferences: 

1. We add it with tabler and/or bootstrap 5 (@princerb was working on something similar in https://github.com/pallets-eco/flask-admin/pull/2643 ) as an alternative to bootstrap 4
2. We bring it in 3.0
3. Bring it in the next minor release, accept the risk of the breaking

hasansezertasan:
After thinking about this more and reading everyone's feedback, I'd like to propose **Option 1** with a twist.

Instead of tying the HTMX inline editing to a BS5/tabler theme (which depends on PR #2643), I'm proposing a **"vanilla" theme** — a dependency-free foundation that uses only semantic HTML, minimal custom CSS, and HTMX as the sole JS dependency. No jQuery, no Bootstrap, no Font Awesome.

I explored this idea through a brainstorming session with Claude Code (Opus), where I guided the design decisions and it helped me think through the architecture and write up the spec.

**Why vanilla?**

- **BYO Theme foundation** — The vanilla theme outputs clean semantic HTML that works as a starting point for *any* CSS framework: BS2-5, Tabler, Tailwind, PicoCSS, or custom. Flask-admin renders HTML, you bring the styles.
- **Testing baseline** — A minimal, predictable theme that Playwright tests can target without fighting Bootstrap's dynamic classes, animations, or JS timing.
- **Migration path clarity** — By building a fully functional theme without BS4, we get concrete answers about what breaks and what the migration looks like.
- **Proves HTMX viability** — Inline editing, modals (via native `<dialog>`), and all interactive features work with HTMX + ~175 lines of vanilla JS, replacing jQuery + Bootstrap JS + Select2 + 8 admin JS files.

To demonstrate adoptability, I'd also ship a **"picocss" theme** alongside it — extending the vanilla templates and just swapping in PicoCSS (~10KB classless CSS). If the vanilla HTML is truly semantic, PicoCSS should "just work" with minimal overrides.

**BS4 stays untouched and remains the default.** The original `XEditableWidget` and `ajax/*` endpoints are restored for BS4. The vanilla theme gets its own `HTMXEditableWidget` and new RESTful `/inline/<pk>/<field>/` endpoints. No backward-incompatible changes.

There's a full design spec behind this. Happy to share if there's interest in discussing the details.

Thoughts?

---

Of course this is just an idea, the possible output might not be exactly like that.

samuelhwilliams:
> I'm proposing a "vanilla" theme — a dependency-free foundation that uses only semantic HTML, minimal custom CSS, and HTMX as the sole JS dependency. No jQuery, no Bootstrap, no Font Awesome.

I'm really in favour of this idea. It's been in the very back of my mind (in a _very_ light way) that it would be nice if Flask-Admin had a very clear 'theming' API/interface that was well defined to support all of the actions needed for this vanilla API, and then use that. I think it would be great if themes for Flask-Admin could be published as separate packages and then just 'plugged in'. I suspect this requires quite a lot of up front thinking through and would be a big undertaking.

While working on a theme myself for some work projects I did have to mangle quite a lot of things and hit some flask-admin internals, so it's not a very clean process. Right now a lot of the functionality required for bootstrap is fairly closely integrated/coupled with flask-admin internals itself, so it'd be really great to detangle some of that.

I'd also strongly prefer that any new 'vanilla' theme we work towards is progessively enhanced, ie resilient to failures in JS (following best practice principles from eg GOV.UK: https://www.gov.uk/service-manual/technology/using-progressive-enhancement - I'm aware this is my specific context a lot of the time, but I think still a strong foundation). UX improvements should ideally be layered on top of that to provide a more full and modern experience.

ElLorans:
I agree with everything you are saying, but we already have closed PR and open PRs just to bring a new theme, and this suggestion increases the workload without bringing us further. My personal opinion is that we should push to get a bootstrap5/tabler whatever template, and then we can refactor from there.

samuelhwilliams:
That approach is fine with me!

ElLorans:
Errors are not rendered. E.g.: if the user submits a non unique field on a unique constrained column, only a console error is registered.
image

ElLorans:
Error msg fixed, I opened #2909 against master to check the rebase was correct and then vs this branch.

1. ~Do we need to change `AjaxSelect2Widget`?~ 
2. ~Do we want to try and reduce the breaking changes? I have no clue how to do that but if I restore the old test suite I can try, the issue is that it would take time for little, not guaranteed benefits~

EDIT: 
1. We do not need to change other widgets
2. Avoiding the breaking changes is not really feasible: the current API uses get_kwargs which muddles CSS classes with backend args. I do not see a way of sustaining the old API and a new one with another template not using bootstrap classes

ElLorans:
> I think making this not a breaking change is not possible (correct me if I am wrong), but I agree the "breaking" is minor. Here are the alternatives I see, sorted by my preferences:
> 
> 1. We add it with tabler and/or bootstrap 5 (@princerb was working on something similar in [[Theme] New theme: FomanticUI [beta] #2643](https://github.com/pallets-eco/flask-admin/pull/2643) ) as an alternative to bootstrap 4
> 2. We bring it in 3.0
> 3. Bring it in the next minor release, accept the risk of the breaking


> After thinking about this more and reading everyone's feedback, I'd like to propose **Option 1** with a twist.
> 
> Instead of tying the HTMX inline editing to a BS5/tabler theme (which depends on PR #2643), I'm proposing a **"vanilla" theme** — a dependency-free foundation that uses only semantic HTML, minimal custom CSS, and HTMX as the sole JS dependency. No jQuery, no Bootstrap, no Font Awesome.

After working on #2910, I have a different perspective vs my previous statements and hasanzerstan opinion.
Option1 with a twist is not possible without the breaking change. A vanilla theme would still need HTMX and maintaining the 2 different APIs is not only more work, but also means changing theme requires the users relying on Ajax and column_editable to change their code together with the template, 

The “breaking change” concern is real but modest. @samuelhwilliams found two external XEditableWidget usages, both in repos untouched for years. The ajax/update/ response format change is also exactly the kind of thing a major version bump communicates clearly.

I would propose to release 3.0 (a new major release) including:
	•	Tabler with HTMX inline editing support from day one (as I am already doing in #2910)
	•	XEditableWidget entirely removed with no alias to "fail fast" during the upgrade process
	•	A migration guide covering: the JS response format change, get_kwargs removal, and x-editable vendor removal
	•	If possible, Fomantic adopting htmx

samuelhwilliams:
If were doing large-scale work on theme support I'd be very keen to start with a bit of high level analysis. I haven't discussed this with anyone, or deeply formulated my thoughts, but I'd be keen to get to a place where flask admin themes are highly pluggable and use a clearly defined and publicly exposed+documented set of APIs.

I don't want to block progress of this if there's little appetite for the wider thinking at the moment, but I think a bit of a reset and design documented for the goals and shape of what needs to be exposed and how would be super useful. At the moment I maintain a separate flask admin theme as a python package and it does have to hack around quite a bit.

I feel like a V3 release having a really clean way of building your own themes would be a great feature for the next major release.

ElLorans:
You mentioned that back in https://github.com/pallets-eco/flask-admin/issues/2682. I can take a look, but frontend is not my specialty: you, Princerb and hasanzerstan are the ones that probably know flask admin themes the best, so I would rely on your guidance.

ElLorans:
I did a quick pass with copilot and it flags:

> ### 1. `ajax_update()` appears to use an unbound form
> 
> The new implementation creates `form = self.list_form()`, strips all fields except the edited one, validates it, and then passes that form to `update_model()`:
> 
> ```python
> if not self.update_model(form, record):
>     ...
> ```
> 
> In the previous flow, updates were performed using a form bound to the existing object. Here the form does not appear to be instantiated with `obj=record` before `update_model()` is called. Depending on the assumptions inside `update_model()`, this could affect editable fields that rely on existing object state during processing.
> 
> Could we add tests covering partial updates of more complex field types to verify that behavior remains unchanged?
> 
> ### 2. Editable fields no longer work for some valid primary keys
> 
> `HTMXEditableWidget.__call__()` now warns when the primary key contains characters outside `[\w-]` and notes that inline editing will not work for that record.
> 
> Many applications use string IDs, encoded composite keys, datastore-generated identifiers, or other key formats that may legitimately contain characters outside that set. This seems like a regression in functionality rather than just a warning condition.
> 
> Would it be possible to escape or encode identifiers instead of disabling inline editing for those records?
> 
> ### 3. `/ajax/update/` response format is now a breaking change
> 
> The changelog notes that `/ajax/update/` now returns HTML instead of plain text. While documented, the endpoint URL itself remains unchanged.
> 
> Any downstream integrations that call this endpoint directly (rather than relying on Flask-Admin's built-in JS/UI) may continue to hit the same endpoint but receive a completely different response format.
> 
> Have you considered a compatibility layer or versioned endpoint for consumers that may depend on the previous behavior?
> 
> ### 4. `_original_widgets` is stored as a class attribute
> 
> In `create_editable_list_form()`:
> 
> ```python
> class ListForm(...):
>     _original_widgets: dict[str, t.Any] = {}
> ```
> 
> The comment indicates this should be per invocation, but it is still attached to the generated class rather than an instance. I may be missing something, but this looks like a potential source of state leakage if generated form classes are reused or cached.
> 
> Could we add tests covering multiple editable list forms to ensure widget restoration remains isolated?
> 
> Overall I like the direction of the HTMX-based implementation, but I think these areas would benefit from additional regression coverage before merge.

I will analyze the issues in the coming weeks and see if they are valid. If anybody has time before that, they are welcome.

ElLorans:
I have tried to address the issues in https://github.com/pallets-eco/flask-admin/tree/feat/htmx but I have not had much time so far.
I could not solve this problem: the values if cancelled and submitted "disappear"
image

hasansezertasan:
Pushed fixes for the inline-editing issues raised above (the "cancel then submit → value disappears" report and the Copilot review).

**Root cause of the disappearing value** (reproduced in a real browser against #2931's `feat/htmx`): the edit popover is a child of the `.editable-cell` element, whose `hx-get` fires on click. Clicks inside the popover bubbled up and re-opened the editor, so `onclick="event.stopPropagation()"` was added to stop it — but that also blocked the document-delegated cancel handler, so the ✗ button did nothing. Clearing a value and clicking cancel left it stuck, and submitting then persisted the empty value (the blank Calories / `0.0` Price in the screenshot).

**Fix:** keep #2931's cleaner submit flow (`hx-target="closest .editable-cell"` + `hx-swap="outerHTML"`) but drop the `stopPropagation` hack in favor of a declarative htmx trigger filter:

```
hx-trigger="click[!target.closest('.editable-popover')]"
```

Inside the popover, clicks no longer re-trigger the GET; outside, cancel works via normal bubbling. Verified: 0 GETs on inner clicks, 1 GET on a cell click.

This also clears two of Copilot's points:
- **PK characters:** the pk is now percent-encoded in the edit URL and no longer used as a CSS selector, so records with pks containing `&`, `#`, etc. are editable (removed the warning that disabled them).
- **Duplicate `id`:** dropped the `id="editable-..."` from the `<td>`; the display `<span>` already carries it, so `hx-target` is unambiguous.

Also ported the `sqla_column_editable` example from #2931 (passes `db` instead of the deprecated `db.session`).

Browser-verified: clear→cancel restores the value and closes; edit→submit updates and stays editable; server-side validation renders inline errors without wiping the cell. Editable test suites (sqla + peewee) pass.

Credit to @ElLorans — the `outerHTML`/`closest` approach comes from #2931, and those commits are co-authored accordingly.

hasansezertasan:
Follow-up on the Copilot review (https://github.com/pallets-eco/flask-admin/pull/2847#issuecomment-4702572385) — status of the four points:

1. **Unbound form in `ajax_update`** — verified this doesn't cause data loss and added a regression test: `test_editable_partial_update_preserves_other_fields` asserts a single-field edit leaves the row's other columns (including other editable ones) untouched, across two consecutive edits. The isolated form is fine because `populate_obj` only writes the one remaining field.
2. **PK characters** — fixed: the pk is percent-encoded in the edit URL and no longer used as a CSS selector, so records with `&`, `#`, etc. in the pk are editable again (the warning that disabled them is gone).
3. **`/ajax/update/` response format** — left as an intentional breaking change, documented under *Breaking changes* in the changelog, consistent with shipping this as a major-version break. No compatibility shim.
4. **`_original_widgets` class attribute** — verified safe and added `test_editable_widgets_isolated_between_views`: two editable views restore their own input widgets (datepicker vs. text) with no cross-view leakage. `create_editable_list_form` builds a fresh `ListForm` class per view and the dict is only ever read during restoration.

Both new tests run across all three SQLA provider variants. cc @ElLorans

ElLorans:
Thanks for the great work!
`test_editable_widgets_isolated_between_views` does not totally address the issue though: ListForm can still leak if used improperly. The test can prevent regressions from our sides, but still exposes a footgun for the end users.

flask_admin/model/base.py
```diff
@@ -2697,70 +2701,158 @@ def ajax_lookup(self) -> T_RESPONSE:
...
+        edit_field.widget = original_widgets.get(field_name, type(edit_field).widget)
 
+    @expose("/ajax/edit/", methods=("GET",))
+    def ajax_edit(self) -> str:
         """
-        if not self.column_editable_list:
+        Return an edit form HTML fragment for a single editable cell.
+        Used by HTMX to swap the display state with an inline edit form.
+        """
+        if not self.can_edit or not self.column_editable_list:
```
ElLorans, line 2719: Is `if not self.can_edit` a breaking change? 
Do we currently allow editable columns through `column_editable_list` if `can_edit` is False?

flask_admin/model/base.py
```diff
@@ -2697,70 +2701,158 @@ def ajax_lookup(self) -> T_RESPONSE:
...
+        if not self.update_model(form, record):
+            return self._render_error(pk, field_name, "Failed to update record.", form)
+
+        # Recompute display value
+        record = self.get_one(pk)
+        display_value = self.get_list_value(None, record, field_name)
+
+        # Return HTMX cell replacement
+        widget = HTMXEditableWidget()
+        return widget(form[field_name], pk=pk, display_value=display_value)
```
ElLorans, line 2858: This prevents custom subclassing to have any effect.
I think we should do something like

```suggestion
        return form[field_name].widget(form[field_name], pk=pk, display_value=display_value)
```

flask_admin/templates/bootstrap4/admin/model/editable_cell_edit.html
```diff
@@ -0,0 +1,34 @@
...
+    {% if form %}
+      {% if form.csrf_token is defined and form.csrf_token %}
+      {{ form.csrf_token }}
+      {% elif csrf_token is defined and csrf_token %}
+      <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
+      {% endif %}
+      {% if form[field_name].type == 'BooleanField' %}
+      <select name="{{ field_name }}" class="form-control" style="min-width: 8em">
+        <option value="y" {{ 'selected' if form[field_name].data else '' }}>Yes</option>
+        <option value="" {{ 'selected' if not form[field_name].data else '' }}>No</option>
```
ElLorans, line 17: We must wrap the `Yes` and `No` in a transl call

ElLorans:
In order to reduce the breaking change and remove an external dependency (htmx), would it make sense to do something like
```
const resp = await fetch('./ajax/update/', { method: 'POST', body: formData });
if (resp.ok) {
  const cellHtml = await fetch(`./ajax/cell/?pk=${pk}&field=${field}`).then(r => r.text());
  cell.outerHTML = cellHtml;
} else {
  // old error contract: (message, 500)
}
```

This keeps `ajax_update`'s response equal to pre-PR behavior but keeps your deduplicated formatting logic, at the cost of one extra small GET per successful edit.

PR #2853, opened 2026-04-01
fix(peewee): support form_choices in Peewee ModelView

### LINKED ISSUE #2235: form_choices for flask_admin.contrib.peewee.ModelView doesn't work
It seems the `form_choices` does not work when using the peewee version of ModelView.

`Flask-Admin==1.6.0`

hasansezertasan:
## Summary

- Override `convert()` in `CustomModelConverter` to check `form_choices` before delegating to wtfpeewee's default conversion, returning a `Select2Field` when choices are defined (matching the SQLAlchemy contrib behavior)
- Add test for `form_choices` in the Peewee test suite
- Add `form_choices` usage to the `peewee_simple` example

Fixes #2235

## Test plan

- [x] `pytest flask_admin/tests/peeweemodel/` — all 12 tests pass
- [x] Manual verification with `peewee_simple` example — `status` field renders as a Select2 dropdown


ElLorans:
This looks solid, but why can't we use the same convert function across the different orms? Maybe we can even upstream it to ModelConverter?

flask_admin/tests/peeweemodel/test_basic.py
```diff
@@ -981,6 +981,123 @@ class Model(BaseModel):
...
+            "status": status_choices,
+            "priority": priority_choices,
+        },
+    )
+    admin.add_view(view)
+
+    form_obj = view.create_form()
+
+    # Check that select field is rendered with correct choices
+    assert type(form_obj.status).__name__ == "Select2Field"  # type: ignore[attr-defined]
```
ElLorans: Why not isinstance?
```suggestion
    assert isinstance(form_obj.status, form.Select2Field)  # type: ignore[attr-defined]
```

PR #2829, opened 2026-03-11
fix the CSRF token on delete files and directories

samialfattani:
CSRF token is not generated in the `file/list.html` page for both file and directory delete `<form>`. This PR uses `csrf_token()` to generate the token and put them in a hidden field to be submitted within the delete action.

**Why this PR:** it supports CSRF in `file/list.html` if the user useed `flask_wtf.csrf.CSRFProtect`

**Test:** This PR adds many test cases that covers all pages of FileAdmin() including list, rename, upload, ...etc. also there are some tests that include the csrf_token in the post request and assert the response to be 200 not 400.

PR #2854, opened 2026-04-01
Fix FomanticUI

samialfattani:
This PR enhances Fomantic UI  #2643  to be compatible with latest changes. 

### List of enhancements

- [x] import JQuery as a local file
- [x]  import `semantic.min.css` (with its included fonts) and `semantic.min.js` as local files (v2.9.4).
- [x] Fix the CSRF token on delete files and directories #2829 is mapped here.
- [x] Fix css class generated by column name #2730, is mapped here.
- [x] Add Menu Divider in the Fomantic  example.
- [x] add `admin_csp_nonce_attribute` where it is missing.
- [x] Fix macro `menu_icon(item)`

### Issues still need to be fixed

- [ ] Menu Divider still not appear properly since the css class is mixed up with Bootstrap4 classes.
- [ ] Inline-Editing is not supporting here. but i think #2847 is doing greate job there, and i thing it will unblock this issue.


ElLorans:
Thanks!

1. Why do we need docker to test this?
2. Can we remove the https://github.com/pallets-eco/flask-admin/pull/2829 since it's a different topic?

samialfattani:
> * Why do we need docker to test this?

what do you mean? i reviewed the source code but i couldn't find any docker refernce in the change. plus, i can run the `examples/fomanticui` normally like all other examples without need to docker.

> Can we remove the https://github.com/pallets-eco/flask-admin/pull/2829 since it's a different topic?

it was a bug in bootstrap4, i think if we are adding new UI we should avoid old mistakes. anyway the CSRF part within HTML pages and we still did not touch any *.py files.

PR #2859, opened 2026-04-14
Exclude LICENSE.txt from Dependabot pip updates (#2856)

giac41:
Dependabot was scanning .txt files in example and test directories which are not actual dependency manifests. 

The following paths are excluded from the pip ecosystem scan:
- examples/**
- flask_admin/tests/**

fixes #2856


ElLorans:
Why excluding flask_admin/tests/**? Shouldn't we exclude license, too?

giac41:
I excluded `flask_admin/tests/**` because I found several `.txt` files in test fixtures that clearly do not look like dependency manifests. However, I agree that excluding the whole directory is broader than necessary.

I checked recent Dependabot commits and did not find any that were clearly tied to `.txt` files, so I cannot confirm that these specific files have caused issues in practice.

That said, I found several `.txt` files in the repository that clearly do not appear to be dependency manifests:

- `./LICENSE.txt`
- `./examples/bootstrap4/files/d1/dummy.txt`
- `./examples/bootstrap4/static/d1/afile.txt`
- `./examples/s3/localdir/yy/zz/afile.txt`
- `./flask_admin/tests/tmp/inner/test1.txt`
- `./flask_admin/tests/fileadmin/files/dummy.txt`
- `./flask_admin/tests/sqla/files/dummy.txt`

So rather than excluding whole directories, I think an explicit `exclude-paths` list for these files would be a safer and more targeted approach.

ElLorans:
A simpler
```
 exclude-paths:
      - "*.txt"
```
would not require manually updating the list, though.

giac41:
After digging deeper, I think the problem is not just an exclusion list, but a mismatch between the current Dependabot config and the actual state of the repository.

The live `.github/dependabot.yml` still contains a `pip` entry pointing at `/requirements/`, but that directory no longer exists in the current repository. The project migrated to `uv` in June 2025 (`pyproject.toml` + `uv.lock` at repo root), and the recent public Python Dependabot PRs in this repo use `dependabot/uv/...` branches, not `dependabot/pip/...`.

Also, the relevant upstream bug is `dependabot/dependabot-core#13960`, which describes exactly a case where Dependabot treats `LICENSE.txt` as a Python requirements file in the `uv` ecosystem. 

So I’m not sure that adding `exclude-paths` under the current `pip: /requirements/` entry can actually affect root files such as `LICENSE.txt`, or the `uv` jobs / PRs.

I also checked recent public Dependabot PRs and did not find one that was clearly triggered by a specific `.txt` file, so I can’t confirm which exact file is responsible.

My suggestion would be:
1. first realign the Dependabot config with the current repo layout (replace/remove the stale `pip: /requirements/` block and add a `uv` entry in the correct location, probably `/`);
2. test on a fork and inspect the Dependabot job logs;
3. only if still needed, add conservative `exclude-paths` on the `uv` entry.

If we still want an immediate workaround, I think an explicit path list is safer than `*.txt`, but only if it is placed under the entry that actually governs the `uv` scans. Because in my opinion using `*.txt` is too coarse and would ignore all `.txt` files in the repository, including ones that could be legitimately used in the future

ElLorans:
My understanding is that the uv/pip mismatch does basically nothing: dependabot uses pip but then it install uv and uses the uv commands.
Can you trigger a dependabot job on your fork and test yourself?
You cannot see failed dependabot runs due to .txt because you don't have the permissions

giac41:
I tested it on my fork.

With `uv` at `/` and no exclusions, the Dependabot update job `uv in /. - Update #1329757521` failed.

Then I added `exclude-paths: - "LICENSE.txt"` (commit 27fcef6), and the next Dependabot update job `uv in /. - Update #1329782778` succeeded. Dependabot also opened a `python:uv` PR on the fork afterwards.

So this appears to fix the blocking updater error related to `LICENSE.txt` on my fork. The remaining failing checks are CI/test failures on the Dependabot PR itself, not the original updater failure.

PR #2881, opened 2026-04-27
fix encoding of file edit

samialfattani:
In the FileAdmin we can allow the user to edit the text files. However, the file can't save non-ascii characters such as emojis🤨. This PR is raised to enable editing files with support of UTF-8 encoding.


samuelhwilliams:
Just to check my understanding, this is only for local file editing right - not azure or s3?

Also I think the dummy file rename is not an intended part of this PR - could you remove that change please? 🙏  I think that's a side effect of some test failures, which isn't ideal, but we shouldn't bring that rename in

samialfattani:
> this is only for local file editing right - not azure or s3?

Yes 

> Also I think the dummy file rename is not an intended part of this PR

I don't know what renaming that u r refferring to ? I didn't rename any file. if you mean `examples/bootstrap4/files/dummy.txt` and `examples/bootstrap4/files/d1/dummy2.txt` ,then those are added just allow test this function in the example. i.e. the is no any unit test will be affected

ElLorans:
> I don't know what renaming that u r refferring to ? I didn't rename any file. if you mean `examples/bootstrap4/files/dummy.txt` and `examples/bootstrap4/files/d1/dummy2.txt` ,then those are added just allow test this function in the example. i.e. the is no any unit test will be affected

It's the first edit: https://github.com/pallets-eco/flask-admin/pull/2881/changes#diff-984cb167561ffd569fa92ae3e3308dc489e4ef7ebe0c4a087aa6b81c70701244

samialfattani:
> - could you remove that change please? 🙏

done

samuelhwilliams:
Looks good - thanks.

Could you include a line update under 'bug fixes' to the changelog please? 🙏

ElLorans:
> Looks good - thanks.
> 
> Could you include a line update under 'bug fixes' to the changelog please? 🙏

maybe it's a tiny bit easier if we merge https://github.com/pallets-eco/flask-admin/pull/2886 first so that we have the new section in the changelog

samialfattani:
> maybe it's a tiny bit easier if we merge #2886 first so that we have the new section in the changelog

 #2886 is merged, can we merge this one ?

samuelhwilliams:
Yep, this looks fine to get in. Please can you add a line to the changelog 🙏

samialfattani:
> Yep, this looks fine to get in. Please can you add a line to the changelog 🙏

done

samuelhwilliams (approved):

PR #2882, opened 2026-04-27
Generate URL Filter-squashed

### LINKED ISSUE #2703: Add method to easily generate a pre-filtered URL

samialfattani:
Fixes #2703

This PR adds a function to generate a URL with filter arguements including all types of filters (Like, NotLike, Greater....etc) and including both indexed (`flt2_3=xx`) and named (`flt3_email_like=@example.com`)

**Usage:**
```
    filtered_url = view.url_for(
        search="exam",
        filters=[
            filters.FilterLike(Model1.test1, "Email", url_value="@example.com"),
            filters.BooleanEqualFilter(Model1.bool_field, "active", url_value=False),
            filters.FloatGreaterFilter(Model1.test5, "Salary", url_value=15.3),
        ],
    )
```

**Output:**
`/admin/user/?search=exam&flt0_0=@example.com&flt1_21=0&flt2_16=15.3`

**Output with named parameters:**
`/admin/user/?search=exam&flt0_test1_contains=@example.com&flt1_bool_field_equals=0&flt4_test2_greater_than=15.3`

### DIFF
--- a/doc/api/mod_model_filters.rst
+++ b/doc/api/mod_model_filters.rst
@@ -1,0 +1,8 @@
+``flask_admin.model.filters``
+==============================
+
+.. automodule:: flask_admin.model.filters
+
+    .. autoclass:: BaseFilter
+        :members:
+        :exclude-members: get_url_argument
--- a/doc/api/mod_model_filters.rst
+++ b/doc/api/mod_model_filters.rst
@@ -5,4 +5,3 @@
 
     .. autoclass:: BaseFilter
         :members:
-        :exclude-members: get_url_argument
--- a/examples/bootstrap4/files/d1/dummy3.js
+++ b/examples/bootstrap4/files/d1/dummy3.js
@@ -1,0 +1,1 @@
+console.log('Hello from dummy3.js');
--- a/examples/bootstrap4/files/d1/dummy3.js
+++ b/examples/bootstrap4/files/d1/dummy3.js
@@ -1,1 +1,0 @@
-console.log('Hello from dummy3.js');
--- a/examples/pymongo_simple/main.py
+++ b/examples/pymongo_simple/main.py
@@ -112,6 +112,51 @@ if __name__ == "__main__":
         conn: MongoClient[Any] = MongoClient(mongo.get_connection_url())
         db = conn.test
 
+        # Insert 5 users
+        users = [
+            {"name": "Alice", "email": "alice@example.com", "password": "pass123"},
+            {"name": "Bob", "email": "bob@example.com", "password": "pass456"},
+            {"name": "Charlie", "email": "charlie@example.com", "password": "pass789"},
+            {"name": "Diana", "email": "diana@example.com", "password": "pass321"},
+            {"name": "Eve", "email": "eve@example.com", "password": "pass654"},
+        ]
+        user_ids = db.user.insert_many(users).inserted_ids
+
+        # Insert some tweets
+        tweets = [
+            {
+                "name": "tweet1",
+                "user_id": user_ids[0],
+                "text": "Hello World!",
+                "testie": True,
+            },
+            {
+                "name": "tweet2",
+                "user_id": user_ids[1],
+                "text": "Flask is awesome",
+                "testie": False,
+            },
+            {
+                "name": "tweet3",
+                "user_id": user_ids[0],
+                "text": "Learning MongoDB",
+                "testie": True,
+            },
+            {
+                "name": "tweet4",
+                "user_id": user_ids[2],
+                "text": "Python rocks!",
+                "testie": False,
+            },
+            {
+                "name": "tweet5",
+                "user_id": user_ids[3],
+                "text": "Admin interface demo",
+                "testie": True,
+            },
+        ]
+        db.tweet.insert_many(tweets)
+
         admin.add_view(UserView(db.user, "User"))
         admin.add_view(TweetView(db.tweet, "Tweets"))
 
--- a/flask_admin/contrib/mongoengine/filters.py
+++ b/flask_admin/contrib/mongoengine/filters.py
@@ -100,7 +100,7 @@ class FilterSmaller(BaseMongoEngineFilter):
         return lazy_gettext("smaller than")
 
 
-class FilterEmpty(BaseMongoEngineFilter, filters.BaseBooleanFilter):
+class FilterEmpty(BaseMongoEngineFilter, filters.BaseEmptyFilter):
     def apply(self, query: QuerySet, value: t.Any) -> QuerySet:
         if value == "1":
             flt = {str(self.column): None}
--- a/flask_admin/contrib/mongoengine/filters.py
+++ b/flask_admin/contrib/mongoengine/filters.py
@@ -38,7 +38,7 @@ class BaseMongoEngineFilter(filters.BaseFilter):
         :param data_type:
             Client data type
         """
-        super().__init__(name, options, data_type)
+        super().__init__(name, options, data_type, column=column)
 
         self.column = column
 
@@ -131,6 +131,9 @@ class FilterInList(BaseMongoEngineFilter):
 
     def operation(self) -> T_TRANSLATABLE:
         return lazy_gettext("in list")
+
+    def stringify(self, value: t.Any) -> str:
+        return ",".join(str(v) for v in value)
 
 
 class FilterNotInList(FilterInList):
--- a/flask_admin/contrib/peewee/filters.py
+++ b/flask_admin/contrib/peewee/filters.py
@@ -33,9 +33,9 @@ class BasePeeweeFilter(filters.BaseFilter):
         :param data_type:
             Client data type
         """
-        super().__init__(name, options, data_type)
+        super().__init__(name, options, data_type, column=column)
 
-        self.column = column
+        self.column: t.Any = column
 
 
 # Common filters
@@ -89,7 +89,7 @@ class FilterSmaller(BasePeeweeFilter):
         return lazy_gettext("smaller than")
 
 
-class FilterEmpty(BasePeeweeFilter, filters.BaseBooleanFilter):
+class FilterEmpty(BasePeeweeFilter, filters.BaseEmptyFilter):
     def apply(self, query: t.Any, value: t.Any) -> t.Any:
         if value == "1":
             return query.filter(self.column >> None)
@@ -118,6 +118,9 @@ class FilterInList(BasePeeweeFilter):
 
     def operation(self) -> T_TRANSLATABLE:
         return lazy_gettext("in list")
+
+    def stringify(self, value: t.Any) -> str:
+        return ",".join(str(v) for v in value)
 
 
 class FilterNotInList(FilterInList):
--- a/flask_admin/contrib/pymongo/filters.py
+++ b/flask_admin/contrib/pymongo/filters.py
@@ -34,7 +34,7 @@ class BasePyMongoFilter(filters.BaseFilter):
         :param data_type:
             Client data type
         """
-        super().__init__(name, options, data_type)
+        super().__init__(name, options, data_type, column=column)
 
         self.column = column
 
--- a/flask_admin/model/base.py
+++ b/flask_admin/model/base.py
@@ -1408,7 +1408,8 @@ class BaseModelView(BaseView, ActionsMixin):
             c1 = flt.column_name().lower()
             c2 = filter_arg.column_name().lower()
 
-            if c2 == c1 and type(flt) == type(filter_arg):
+            c2_is_a_c1 = issubclass(type(filter_arg), type(flt))
+            if c2 == c1 and c2_is_a_c1:
                 filter_arg_key = k
                 break
 
@@ -1419,6 +1420,28 @@ class BaseModelView(BaseView, ActionsMixin):
         search: str | None = None,
         filters: list[tuple[BaseFilter, t.Any]] | None = None,
     ) -> str:
+        """Return URL for the view with applied filters and search query.
+
+        :param search:
+            Optional search query string.
+        :param filters:
+            List of tuples containing (BaseFilter instance, filter value).
+            If duplicate filters are provided (filter type/same column), they
+            both will be included in the URL with different parameter names.
+            Both filters will be applied with an AND condition. Note that
+            the order of filters in the list does not matter.
+            Example::
+
+                filters = [
+                    (FilterLike(column='first_name'), 'John'),
+                    (FilterLike(column='first_name'), 'Jane')
+                ]
+                myview.url_for(filters=filters)
+                # Output: /admin/myview?flt0_0=John&flt1_0=Jane
+
+        :return:
+            URL string with filter and search parameters applied.
+        """
         url_args = {}
         if search:
             url_args["search"] = search
@@ -1432,7 +1455,7 @@ class BaseModelView(BaseView, ActionsMixin):
                 warnings.warn(
                     f"The filter {flt.__class__.__name__}('{flt.name}') is not found "
                     "in filter arguments, did you forget to add it in column_filters?",
-                    stacklevel=1,
+                    stacklevel=2,
                 )
                 continue
             else:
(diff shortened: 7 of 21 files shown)


examples/pymongo_simple/main.py
samuelhwilliams, line 1: The changes to this example aren't needed as part of this PR - ideally we would keep PRs small and scoped to help review. This could have gone in separately if it's just for general example behaviour unrelated to generating URL filters.

examples/bootstrap4/files/d1/dummy3.js
samuelhwilliams, line 1: What is this file for?

flask_admin/contrib/mongoengine/filters.py
```diff
@@ -21,6 +23,7 @@ def __init__(
         name: str,
         options: T_OPTIONS = None,
         data_type: T_WIDGET_TYPE = None,
+        url_value: str | None = None,
```
samuelhwilliams: I'm not sure about adding this value to all of the Filter classes - it feels like a less clear interface. One of the tiny perks of doing this, I guess, would be to add specific type hints for each Filter subclass. But we don't do that.

I think we could leave all of the filter classes untouched and specifically pass the url values as part of the `get_url` interface, something like:

```python
view.get_url(
    search="foo",
    filters=[
        (FilterEqual(Account.email, "Email"), "jane@doe.com"),
        (FilterGreaterThan(Account.age, "Age"), 30),
    ]
)
```

I think this would make a smaller changeset overall and I'm not convinced that the Filters need to understand "values" used for generating URLs - we can keep the code impact of that limited to just `get_url`?

flask_admin/model/base.py
```diff
@@ -1955,6 +2011,7 @@ def _get_list_filter_args(self) -> list[T_FILTER] | None:
 
         return None
 
+    # TODO: rename to get_view_args()
```
samuelhwilliams: We shouldn't leave todos in merged code

flask_admin/model/filters.py
```diff
@@ -22,6 +22,10 @@ def __init__(
         options: T_OPTIONS = None,
         data_type: T_WIDGET_TYPE = None,
         key_name: str | None = None,
+        column: t.Any | None = None,
+        # FIXME: to be renamed in future releases into `value`
+        # thus, `self.value` should be used in functions like clean, validate, etc.
```
samuelhwilliams: Any rename after we merge this would be a breaking public API change. We shouldn't merge things without a good attempt at naming first.

flask_admin/model/base.py
```diff
@@ -1336,7 +1336,7 @@ def get_filters(self) -> list[BaseFilter] | None:
         else:
             return None
 
-    def get_filter_arg(self, index: int, flt: BaseFilter) -> str:
+    def get_filter_arg_name(self, index: int, flt: BaseFilter) -> str:
```
samuelhwilliams: We shouldn't rename part of the public API without a really good reason. Is there one for this?

samuelhwilliams (changes requested):
I think my main issue here is on the implementation being that the URL value is baked into the Filter definition. I'm not sure this feels like the right developer experience/API.

I think we should take this back to an issue and plan out the implementation up front rather than do it on a PR.

I think I'd be more interested in something like this:

```python
    filtered_url = view.url_for(
        search="exam",
        filters=[
            (filters.FilterLike(Model1.test1, "Email"), "@example.com"),
            (filters.BooleanEqualFilter(Model1.bool_field, "active"), False),
            (filters.FloatGreaterFilter(Model1.test5, "Salary", 15.3),
        ],
    )
```

Ideally with full type support so that type checkers can statically verify that you've passed the correct argument type for the appropriate filter, ie boolean values for a Boolean filter and string values for a like filter.

examples/bootstrap4/files/d1/dummy3.js
samialfattani: sorry , i was mistakenly added

flask_admin/model/base.py
samialfattani: sorry for that, then i prefer to change it into FIXME where the current name is a bit confusing, ya?

flask_admin/model/base.py
samialfattani: i mentioned that in previouse discussion that the whole code related to filtering need a comprehensive refactoring due the the ambigouity. `get_filter_arg` actually returns a name of an argu (not the whole arg). on the other side we can find that `self._filter_args` is a `dict[str, tuple[int, BaseFilter]]`. now we end up with     `_find_filter_arg()` and `get_filter_arg()` functions,  which each returns different type in unexpected way. renaming here is just for seeking better code clarity. `get_filter_arg` returns a unique name of the filter if the naming is enabled otherwise it returns the same passed index, while the `find_filter_arg()` returns a filter object.

examples/pymongo_simple/main.py
samialfattani: simply i wanted to add some records to test the filter's js behavior. as you know filters input fields are fully created by JS and there is no way to test them using pytest. I think changes in this example accelrate testing the filter.

flask_admin/model/base.py
samuelhwilliams: Fair. I appreciate the code may be confusing, but this is part of our public API and so I think we should respect a deprecation cycle if we feel the need to rename it.

samialfattani: ok , i will just revert it for now until we have a refactoring plan

flask_admin/model/filters.py
samialfattani: ok no need to rename this

flask_admin/contrib/mongoengine/filters.py
samialfattani: i agree, i fixed the code to fit your suggestion.

samialfattani:
now this style is adapted. you can check and review

```
    filtered_url = view.url_for(
        search="exam",
        filters=[
            (filters.FilterLike(Model1.test1, "Email"), "@example.com"),
            (filters.BooleanEqualFilter(Model1.bool_field, "active"), False),
            (filters.FloatGreaterFilter(Model1.test5, "Salary", 15.3),
        ],
    )
```

flask_admin/model/filters.py
```diff
@@ -115,15 +162,33 @@ class BaseBooleanFilter(BaseFilter):
...
         )
 
     def validate(self, value: str) -> bool:
         return value in ("0", "1")
 
+    def stringify(self, value: t.Any) -> str:
+        return "1" if value else "0"
+
+
+class BaseEmptyFilter(BaseBooleanFilter):
```
ElLorans, line 196: Why is this needed?

samialfattani: it is indirectly needed in the test 

https://github.com/samialfattani/flask-admin/blame/2a1d93d64a93ad3d54480628ab0516cbb84d4292/flask_admin/tests/test_filters.py#L48-L50

while BooleanFilter cleans the "1"  into ("1" or True) yet the EmptyFilter cleans it into ("1").  that's why the test expects different cleaned from each filter. BaseEmptyFilter is needed here only to categorize all Emptyfilter classes so that it is easy to test them.

samialfattani:
@samuelhwilliams , i thought you wanted to review and confirm this PR. i think all your comments are addressed and resolved,

examples/bootstrap4/files/d1/dummy3.js
ElLorans: Can you remove it?

flask_admin/model/base.py
```diff
@@ -1392,6 +1392,64 @@ def _get_filter_groups(self) -> OrderedDict[str, list[dict[str, t.Any]]] | None:
...
+        """
+        if not self._filter_args:
+            return None
+
+        filter_arg_key = None
+        for k, (_index, filter_arg) in self._filter_args.items():
+            c1 = flt.column_name().lower()
+            c2 = filter_arg.column_name().lower()
+
+            if c2 == c1 and type(flt) == type(filter_arg):
```
ElLorans: This will fail with subclasses, shouldn't we use `isinstance()` rather than `type() == type()`?

flask_admin/model/base.py
```diff
@@ -1392,6 +1392,64 @@ def _get_filter_groups(self) -> OrderedDict[str, list[dict[str, t.Any]]] | None:
...
+        if filters is None:
+            filters = []
+
+        for i, (flt, value) in enumerate(filters):
+            found_arg = self._find_filter_arg(flt)
+            if not found_arg:
+                warnings.warn(
+                    f"The filter {flt.__class__.__name__}('{flt.name}') is not found "
+                    "in filter arguments, did you forget to add it in column_filters?",
+                    stacklevel=1,
```
ElLorans: I think this should be stacklevel 2 or 3, can you check?

flask_admin/model/base.py
```diff
@@ -1392,6 +1392,64 @@ def _get_filter_groups(self) -> OrderedDict[str, list[dict[str, t.Any]]] | None:
...
+                filter_arg_key = k
+                break
+
+        return self._filter_args[filter_arg_key] if filter_arg_key else None
+
+    def url_for(
+        self,
+        search: str | None = None,
+        filters: list[tuple[BaseFilter, t.Any]] | None = None,
+    ) -> str:
```
ElLorans, line 1422: We definitely need a docstring explaining what happens with duplicate filters and how/if order is impacting

flask_admin/tests/test_filters.py
```diff
@@ -0,0 +1,185 @@
...
+        if issubclass(FilterClass, filters.BaseTimeFilter):
+            params.append((FilterClass, "15:00:00", time(15, 0, 0)))
+            params.append((FilterClass, "03:00:00 AM", ValueError))
+
+        if issubclass(FilterClass, filters.BaseTimeBetweenFilter):
+            params.append(
+                (FilterClass, "15:00:00 to 20:00:00", [time(15, 0, 0), time(20, 0, 0)])
+            )
+
+        # FIXME: test is skipped unexpectedly, need to investigate.
```
ElLorans, line 113: do you need help in fixing this one?

flask_admin/model/base.py
```diff
@@ -1969,6 +2027,7 @@ def _get_list_filter_args(self) -> list[T_FILTER] | None:
 
         return None
 
+    # FIXME: rename to get_view_args()
```
ElLorans: Can we just remove this comment?

flask_admin/model/base.py
samialfattani: you mean using `issubclass(type(filter_arg), type(flt)` , yes sure we should, and it is changed

flask_admin/model/base.py
samialfattani: For that warning site, stacklevel 2 is the correct default, not 3.

Reasoning:

1. The warning is emitted inside BaseModelView.url_for at `base.py`.
2. stacklevel 2 points to the direct caller of url_for, which is usually the user/framework code that passed the filter.
3. stacklevel 3 would point one frame too high in most cases and is less accurate.

flask_admin/model/base.py
samialfattani: ok

flask_admin/model/base.py
samialfattani: yes sure i will write this

flask_admin/tests/test_filters.py
samialfattani: yes ,

flask_admin/model/base.py
ElLorans: I think issubclass returns False if the class is the same one, so we should use isinstance

samialfattani: issubclass returns True when both classes are the same assuming that ClassA can be a subclass of itself. using isinstance works also fine but it compares Object Vs. Type (not Type Vs. Type). check the following code:

```
class ClassA:
    pass

class ClassB(ClassA):
    pass

print(isinstance(ClassB(), ClassA))
print(isinstance(ClassA(), ClassA))
print(isinstance(ClassB(), ClassB))

print('----------')
print(issubclass(ClassB, ClassA))
print(issubclass(ClassA, ClassA))
print(issubclass(ClassB, ClassB))
```

output
```
Hello.
True
True
True
----------
True
True
True
```

examples/bootstrap4/files/d1/dummy3.js
samialfattani: ok, it is deleted now

flask_admin/model/base.py
```diff
@@ -1408,7 +1408,8 @@ def _find_filter_arg(self, flt: BaseFilter) -> tuple[int, BaseFilter] | None:
             c1 = flt.column_name().lower()
             c2 = filter_arg.column_name().lower()
 
-            if c2 == c1 and type(flt) == type(filter_arg):
+            c2_is_a_c1 = issubclass(type(filter_arg), type(flt))
```
ElLorans, line 1411: Are you sure here you need issubclass and not isinstance?

samialfattani: yes `isinstance()` will need an instrance creation to make the comparison which seems a little bit awkward

doc/api/mod_model_filters.rst
```diff
@@ -0,0 +1,8 @@
+``flask_admin.model.filters``
+==============================
+
+.. automodule:: flask_admin.model.filters
+
+    .. autoclass:: BaseFilter
+        :members:
+        :exclude-members: get_url_argument
```
ElLorans: why exclude-members?

samialfattani: sorry it was a mistake, now this line is removed, i also added some docstring for `get_url_argument`

flask_admin/model/filters.py
```diff
@@ -134,7 +199,7 @@ class BaseIntFilter(BaseFilter):
     """
 
     def clean(self, value: str) -> int:
-        return int(value)
+        return int(value) if value else 0
```
ElLorans, line 202: Is this a breaking change? Previously an empty string would raise `ValueError` (correctly signaling "invalid filter")

flask_admin/model/filters.py
```diff
@@ -115,15 +162,33 @@ class BaseBooleanFilter(BaseFilter):
...
+            (("1", lazy_gettext("Yes")), ("0", lazy_gettext("No"))),
+            data_type,
+            column,
         )
 
     def validate(self, value: str) -> bool:
         return value in ("0", "1")
 
+    def stringify(self, value: t.Any) -> str:
+        return "1" if value else "0"
```
ElLorans, line 182: Isn't this a bug? If value is "0", stringify returns "1"

flask_admin/contrib/mongoengine/filters.py
```diff
@@ -38,7 +38,7 @@ def __init__(
         :param data_type:
             Client data type
         """
-        super().__init__(name, options, data_type)
+        super().__init__(name, options, data_type, column=column)
 
         self.column = column
```
ElLorans, line 43: can't we remove this now?

flask_admin/contrib/peewee/filters.py
```diff
@@ -33,9 +33,9 @@ def __init__(
         :param data_type:
             Client data type
         """
-        super().__init__(name, options, data_type)
+        super().__init__(name, options, data_type, column=column)
 
-        self.column = column
+        self.column: t.Any = column
```
ElLorans, line 38: can't we remove this line now?

flask_admin/contrib/pymongo/filters.py
```diff
@@ -34,7 +34,7 @@ def __init__(
         :param data_type:
             Client data type
         """
-        super().__init__(name, options, data_type)
+        super().__init__(name, options, data_type, column=column)
 
         self.column = column
```
ElLorans, line 39: can't we remove this line now?

flask_admin/model/filters.py
ElLorans: But now code relying on `isinstance(flt, BaseBooleanFilter)` checks will break

PR #2883, opened 2026-04-28
Support string column paths in filters

samuelhwilliams:
When setting `column_filters` on a model, you can write dot-notated paths like 'author.account.email' to traverse relationships and reach a leaf attribute.

But when instantiating Filter classes you can't do this and need to pass a Column/InstrumentedAttribute directly. This makes it hard to traverse deeper relationships when using Filters vs using flask-admin's string resolution.

This patch makes it possible to pass string paths to Filters and handles resolving all of the required joins in the same way as passing raw strings to `column_filters`.

This change is backwards compatible.


ElLorans (approved):

flask_admin/contrib/sqla/filters.py
```diff
@@ -18,11 +19,27 @@
...
+                FilterEqual(column=Post.title, name="Title"),
+                # Or use a dotted path resolved against the view's model:
+                FilterLike(column="author.account.email", name="Author Email"),
+            ]
     """
 
     def __init__(
         self,
-        column: T_SQLALCHEMY_COLUMN | T_INSTRUMENTED_ATTRIBUTE,
+        column: T_SQLALCHEMY_COLUMN | T_INSTRUMENTED_ATTRIBUTE | str,
```
ElLorans:
```suggestion
        column: T_COLUMN,
```

flask_admin/contrib/sqla/filters.py
```diff
@@ -383,14 +468,25 @@ def clean(self, value: t.Any) -> t.Any:
 class EnumFilterNotEqual(FilterNotEqual):
     def __init__(
         self,
-        column: T_SQLALCHEMY_COLUMN,
+        column: t.Union[T_SQLALCHEMY_COLUMN, str],
```
ElLorans:
```suggestion
        column: T_COLUMN,
```

flask_admin/contrib/sqla/filters.py
```diff
@@ -400,26 +496,48 @@ def clean(self, value: t.Any) -> t.Any:
 class EnumFilterEmpty(FilterEmpty):
     def __init__(
         self,
-        column: T_SQLALCHEMY_COLUMN,
+        column: t.Union[T_SQLALCHEMY_COLUMN, str],
```
ElLorans:
```suggestion
        column: T_COLUMN,
```

flask_admin/contrib/sqla/filters.py
```diff
@@ -400,26 +496,48 @@ def clean(self, value: t.Any) -> t.Any:
...
+                    "_on_column_resolved called before column was resolved."
+                )
+            self.enum_class = self.column.type.enum_class  # type: ignore[union-attr]
+
 
 class EnumFilterInList(FilterInList):
     def __init__(
         self,
-        column: T_SQLALCHEMY_COLUMN,
+        column: t.Union[T_SQLALCHEMY_COLUMN, str],
```
ElLorans:
```suggestion
        column: T_COLUMN,
```

flask_admin/contrib/sqla/filters.py
```diff
@@ -433,14 +551,25 @@ def clean(self, value: t.Any) -> t.Any:
 class EnumFilterNotInList(FilterNotInList):
     def __init__(
         self,
-        column: T_SQLALCHEMY_COLUMN,
+        column: t.Union[T_SQLALCHEMY_COLUMN, str],
```
ElLorans:
```suggestion
        column: T_COLUMN,
```

ElLorans:
Was not expecting this to be so complex.

Thanks also for already updating the changelog.

I think we need to rename the type hints though, since I keep getting confused between T_COLUMN and its many different variations. I suggested to always use that one rather than `Union[T_SQLALCHEMY_COLUMN, str]`, but the PR can also be merged as is.

ElLorans:
LGTM, happy to merge myself

samuelhwilliams:
Thanks for the review! I just want to test a few more things on this - possibly I was over-eager on marking as ready for review (sorry).

Not expecting any fundamental changes but maybe tweaking an example app to actually dig deep through some nesting.

Also I've definitely just focused on SQLAlchemy here because I'm biased as that's all I use, but I maybe need to spend a few moments investigating if we should give feature parity to other backends (if even relevant)

ElLorans:
Hey, no worries. Let us know when ready. If you want to, you can mark the PR as draft. We'll keep receiving notifications because github is a bit silly, so keeping it as ready is fine too.

samuelhwilliams:
I think ready for a re-review @ElLorans. This PR introduces some state to Filter instances if using the dotted-string path; the Filter needs to resolve the string against the model view in order to work out the relationship chain. When using the SQLAlchemy column approach, the Filters can be defined once at the module level and reused across multiple views. When using the new string approach they can't be shared because they need to be bound to a specific model in order for SQLAlchemy to know what the base of the relationship chain is in order to resolve the column and get its data type out.

I think this is OK - it only changes developer implementation details for Filters using the new string-dotted path, so existing code remains backwards compatible.

It'd be quite a dance to keep Filters entirely stateless while supporting string-dotted paths.

ElLorans:
Oh, that's a bit of a bummer. Isn't it possible to resolve the relationship only when the filter is invoked?

samuelhwilliams:
I think not without changing how the filters are applied to also feed in the model/modelview, which would have a wider impact radius and mean recalculating the joins on every request (or adding caching somewhere else)?

examples/sqla/admin/data.py
```diff
@@ -82,6 +84,13 @@ def build_sample_db():
         ("CN", "China", 86, "CNY", "Asia/Shanghai"),
     ]
 
+    accounts = []
+    for _, first in enumerate(first_names):
```
ElLorans: I am not sure I understand why the _ if we never use it. I think we can also use add_all rather than adding Accounts one by one

ElLorans (approved):

PR #2886, opened 2026-04-28
bugfix: v2.1.0 UnboundLocalError in sqla.ModelView

### LINKED ISSUE #2885: Unbound variable error in update_model since 2.1.0
Errors in `on_model_change` cause an `UnboundLocalError` in https://github.com/pallets-eco/flask-admin/blob/v2.1.0/flask_admin/contrib/sqla/view.py#L1419 

> E   UnboundLocalError: cannot access local variable 'session' where it is not associated with a value

The `session` variable  is only created after that method is called, so if it raises an error then it doesn't exist and Python throws that error.

This works in 2.0.2, but broken in 2.1.0.

Environment:

- Python version: 3.12
- Flask version: 3.1.3
- Flask-Admin version: 2.1.0

ElLorans:
fixes #2885

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+2.1.1 [unreleased]
+------------------
+
+Bugfixes:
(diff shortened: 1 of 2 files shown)


samuelhwilliams (approved):
Nice, thank you 🙇

PR #2889, opened 2026-05-02
Ignore uv.lock files from examples

ElLorans:
uv.lock in examples are not useful for tracking since we do not use `uv run --locked main.py` but `uv run main.py` and they only cause many useless dependabot PR that are already polluting the commit history.

I ran `git rm --cached $(find examples/ -type f -name "uv.lock")` to remove them from the index


hasansezertasan (approved):
Looked good to me, thanks @ElLorans 🚀.

PR #2890, opened 2026-05-04
bump lock dependencies, remove dead ones

ElLorans:
Supersedes https://github.com/pallets-eco/flask-admin/pull/2894 and similars.

1. remove `tox<4.30` constraint after https://github.com/tox-dev/tox/issues/3602
2. Update tox code runner in tests.yaml to exclude "." since now tox strips "." from environments (e.g. py311 rather than py3.11): https://github.com/pallets-eco/flask-admin/blob/739b7712dd3192ede8f624c8738aa5820213c5b9/.github/workflows/tests.yaml
3. pin `azure-storage-blob<12.28.0` in testing since Azurite 3.35.0 only supports up to API 2025-11-05
4. remove `pyright` until we fix all the warnings and add it to the type testing matrix
5. remove `types-Pillow` since Pillow 10.3.0 introduced types but in lock we are already using 12.2.0
6. remove `types-beautifulsoup4` since bs4 introduced types in 4.13 but in lock we were using 4.14.3
7. remove pylint
8. remove most `tool.mypy.overrides`

Updating `typesheds` and `wtforms` allows to remove a few `type: ignore` but had to fix 1 possible bug and type ignore another one. Both appear to be false alarms (self.choices should not be None):

1. https://github.com/pallets-eco/flask-admin/pull/2890/changes#diff-ed0780d4a77dcbff70ec5c0d8552223de602eeabe3a2d8e6a2595a691a39c533
2. https://github.com/pallets-eco/flask-admin/pull/2890/changes#diff-3e3a15187c1f84fd2a867e515ac8063482beb0ce770ce64cc06df6075c424bbc

### DIFF
--- a/flask_admin/model/widgets.py
+++ b/flask_admin/model/widgets.py
@@ -132,8 +132,9 @@ class XEditableWidget:
         elif field.type in ["Select2Field", "SelectField"]:
             field = t.cast(Select2Field | SelectField, field)
             kwargs["data-type"] = "select2"
-            choices = [  # type:ignore[misc]
-                {"value": x, "text": y} for x, y in field.choices
+            choices = [  # type:ignore[misc, str-unpack]
+                {"value": x, "text": y}
+                for x, y in field.choices  # type: ignore[union-attr]
             ]
 
             # prepend a blank field to choices if allow_blank = True
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -82,7 +82,15 @@ jobs:
         env:
           PGPASSWORD: postgres
         run: psql -U postgres -h localhost -c 'CREATE EXTENSION hstore;' flask_admin_test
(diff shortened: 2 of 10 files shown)


flask_admin/model/widgets.py
```diff
@@ -132,9 +132,7 @@ def get_kwargs(self, field: Field, kwargs: dict[str, str]) -> dict[str, str]:
         elif field.type in ["Select2Field", "SelectField"]:
             field = t.cast(Select2Field | SelectField, field)
             kwargs["data-type"] = "select2"
-            choices = [  # type:ignore[misc]
-                {"value": x, "text": y} for x, y in field.choices
-            ]
+            choices = [{"value": x, "text": y} for x, y in field.choices or []]
```
ElLorans:
```suggestion
            choices = [{"value": x, "text": y} for x, y in field.choices or ()]
```

PR #2901, opened 2026-05-22
docs(inline): correct inline post-processing example and fix Peewee parity (#1738)

### LINKED ISSUE #1738: Incorrect documentation for post-processing inline forms
To post-process inline forms, the documentation for [flask_admin.contrib.sqla.ModelView.inline_model_form_converter](https://flask-admin.readthedocs.io/en/latest/api/mod_contrib_sqla/#flask_admin.contrib.sqla.ModelView.inline_model_form_converter) states to override `post_process` in `InlineModelConverter`. However, doing so has no effect.

I tried it with `postprocess_form` instead of `post_process`, but this did not have any effect either. What is the correct function to override for post-processing inline forms?

jbbqqf:
## Summary

The `inline_model_form_converter` docstrings in
`flask_admin.contrib.sqla.view` and `flask_admin.contrib.peewee.view`
advertise a `post_process` method on a converter subclass. That method
name is never invoked anywhere in the codebase, so following the
documented example is silently a no-op. The hook the SQLAlchemy
converter actually calls is `InlineFormAdmin.postprocess_form` (see
`flask_admin/contrib/sqla/form.py:1014,1139`); the `examples/sqla_custom_inline_forms/main.py:132`
example uses that real hook.

Fixes #1738.

## Context

- `flask_admin/contrib/sqla/view.py:213-224` (pre-change) and
  `flask_admin/contrib/peewee/view.py:112-123` (pre-change) showed:

  ```python
  class MyInlineModelConverter(InlineModelConverter):
      def post_process(self, form_class, info):
          ...
  ```

  Grepping the repo for `post_process` returns only those two docstring
  occurrences -- nothing in the code path ever calls a method by that
  name.

- The real contract is on `InlineBaseFormAdmin`:
  `flask_admin/model/form.py:120-134` defines
  `postprocess_form(self, form_class)` and the SQLA converter invokes it
  at `flask_admin/contrib/sqla/form.py:1014` and `:1139`.

- The Peewee inline converter (`flask_admin/contrib/peewee/form.py:287-350`)
  never invoked `postprocess_form` at all, so the documented pattern was
  effectively broken on the Peewee backend as well.

## Changes

- `flask_admin/contrib/sqla/view.py`, `flask_admin/contrib/peewee/view.py`:
  rewrite the `inline_model_form_converter` docstring example to use
  `InlineFormAdmin.postprocess_form` (the hook the converter actually
  invokes) and to pass the subclass through `inline_models`.
- `flask_admin/contrib/peewee/form.py`: call
  `info.postprocess_form(child_form)` inside
  `InlineModelConverter.contribute`, mirroring the long-standing SQLA
  behavior so the documented hook works on both backends.
- `flask_admin/tests/sqla/test_inlineform.py`:
  add `test_inline_form_postprocess_form_hook` that subclasses
  `InlineFormAdmin`, contributes an extra `StringField` via
  `postprocess_form`, and asserts the field lands on the generated
  inline form class. Pins the documented contract.
- `flask_admin/tests/peeweemodel/test_basic.py`:
  add a matching regression test that fails on `origin/main` (the hook
  was never called) and passes on this branch.
- `doc/changelog.rst`: add `[unreleased]` entries for the documentation
  fix and the Peewee bug fix.

## Reproduce BEFORE/AFTER yourself (copy-paste)

```bash
# --- one-time setup ---
git clone https://github.com/pallets-eco/flask-admin.git /tmp/fa-1738 && cd /tmp/fa-1738
uv venv .venv --python 3.12
source .venv/bin/activate
uv pip install -e ".[sqlalchemy,peewee]" pytest flask-sqlalchemy-lite

# --- BEFORE: origin/master, Peewee `postprocess_form` hook is never called ---
git checkout origin/master
git fetch https://github.com/jbbqqf/flask-admin.git docs/1738-inline-postprocess-form
git checkout FETCH_HEAD -- flask_admin/tests/peeweemodel/test_basic.py
pytest flask_admin/tests/peeweemodel/test_basic.py::test_inline_form_postprocess_form_hook -v
# Expected: FAILED -- "InlineFormAdmin.postprocess_form should contribute
#                      extra fields on the Peewee backend just like it
#                      does on the SQLAlchemy backend."
git checkout origin/master -- flask_admin/tests/peeweemodel/test_basic.py

# --- AFTER: this branch, both SQLA and Peewee fire the documented hook ---
git checkout FETCH_HEAD
pytest \
  flask_admin/tests/peeweemodel/test_basic.py::test_inline_form_postprocess_form_hook \
  flask_admin/tests/sqla/test_inlineform.py::test_inline_form_postprocess_form_hook \
  -v
# Expected: 4 passed, 1 skipped (SQLALite-with-session_deprecated is skipped
#           by the conftest fixture; the SQLA + Peewee cases all pass).
```

## What I ran locally

- New tests, focused:
  ```
  pytest flask_admin/tests/peeweemodel/test_basic.py::test_inline_form_postprocess_form_hook \
         flask_admin/tests/sqla/test_inlineform.py::test_inline_form_postprocess_form_hook -v
  # 4 passed, 1 skipped
  ```
- Whole inline test file:
  ```
  pytest flask_admin/tests/sqla/test_inlineform.py -q
  # 18 passed, 6 skipped
  ```
- Broader regression sweep (excluding the Postgres-only test module, which
  needs `psycopg2`):
  ```
  pytest flask_admin/tests/peeweemodel/ flask_admin/tests/sqla/ \
         flask_admin/tests/test_base.py flask_admin/tests/test_model.py \
         --ignore=flask_admin/tests/sqla/test_postgres.py -q
  # 335 passed, 81 skipped, 13 xfailed
  ```
- Lint:
  ```
  ruff check flask_admin/contrib/peewee/form.py flask_admin/contrib/peewee/view.py \
             flask_admin/contrib/sqla/view.py \
             flask_admin/tests/peeweemodel/test_basic.py \
             flask_admin/tests/sqla/test_inlineform.py
  # All checks passed!
  ```

## Edge cases

| # | Scenario | Input | Expected | Verified by |
|---|---|---|---|---|
| 1 | SQLA backend, subclass `InlineFormAdmin` and contribute `extra` field via `postprocess_form` | `inline_models = (UserInfoInlineForm(UserInfo),)` | `view._create_form_class.info.args[0]` has `extra` attribute | new `test_inline_form_postprocess_form_hook[SQLAProvider-*]` |
| 2 | Peewee backend, same subclass-and-contribute pattern | `inline_models = (Model2InlineForm(Model2),)` | generated inline form class has `extra` attribute | new `test_inline_form_postprocess_form_hook` (peewee) |
| 3 | Plain `inline_models = (Model,)` with no `InlineFormAdmin` subclass | existing behavior | continues to work unchanged | existing `test_inline_form` / `test_inline_form_required` / `test_inline_form_self` / `test_inline_form_base_class` all still pass |
| 4 | Peewee `InlineFormAdmin` returning a custom form via `get_form()` (skips the auto-generated path) | `info.get_form()` returns non-None | still passes through `postprocess_form` after assignment | code path now uniformly calls `postprocess_form` after the `if child_form is None` branch, matching the SQLA backend's structure |

## Risk / blast radius

- Docstring-only change for SQLA: no behavior change.
- Peewee `InlineModelConverter.contribute`: now calls
  `info.postprocess_form(child_form)`. The default
  `InlineFormAdmin.postprocess_form` (`flask_admin/model/form.py:134`)
  returns the form class unchanged, so existing call sites that never
  overrode the hook see no diff. Behavior changes only for users who
  *do* override the hook -- which is the previously-documented but
  broken case we're fixing.
- All non-Postgres tests pass locally; the Postgres test module fails
  to import only because `psycopg2` isn't installed in my dev env (it's
  unrelated to this diff).

## Release note

```release-note
Fix the SQLAlchemy/Peewee `inline_model_form_converter` docstring example
to point at the real `InlineFormAdmin.postprocess_form` hook, and wire
the Peewee backend to call that hook so the documented pattern actually
works there.
```

---

*PR drafted with assistance from Claude Code (Anthropic). The change was
reviewed manually against flask-admin's source -- file:line references in
the body were checked against the cloned tree, and the copy-paste
BEFORE/AFTER reproducer block is the one I used during development;
reviewers can paste it verbatim.*

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -6,6 +6,10 @@ Changelog
 
 Bugfixes:
 * Fix encoding for editing file in FileAdmin. Now it uses UTF-8 and accepts non-ASCII characters.
+* Peewee backend: invoke ``InlineFormAdmin.postprocess_form`` from the inline-model converter so it can contribute extra fields onto inline forms, matching the long-standing SQLAlchemy behavior (part of #1738).
+
+Documentation:
+* Fix the ``inline_model_form_converter`` docstrings in the SQLAlchemy and Peewee model views. They previously showed a ``post_process`` method on a converter subclass, which is never invoked anywhere in the codebase. The hook actually called by the converter is ``InlineFormAdmin.postprocess_form``; the examples now reflect that (closes #1738).
 
 Type hints:
 * Type hints added to all functions and methods (some using `typing.Any` where full typing not yet available)
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -6,7 +6,7 @@ Changelog
 
 Bugfixes:
 * Fix encoding for editing file in FileAdmin. Now it uses UTF-8 and accepts non-ASCII characters.
-* Peewee backend: invoke ``InlineFormAdmin.postprocess_form`` from the inline-model converter so it can contribute extra fields onto inline forms, matching the long-standing SQLAlchemy behavior (part of #1738).
+* Peewee backend: invoke ``InlineFormAdmin.postprocess_form`` matching SQLAlchemy behavior (#1738).
 
 Documentation:
 * Fix the ``inline_model_form_converter`` docstrings in the SQLAlchemy and Peewee model views. They previously showed a ``post_process`` method on a converter subclass, which is never invoked anywhere in the codebase. The hook actually called by the converter is ``InlineFormAdmin.postprocess_form``; the examples now reflect that (closes #1738).
--- a/flask_admin/contrib/peewee/form.py
+++ b/flask_admin/contrib/peewee/form.py
@@ -328,6 +328,11 @@ class InlineModelConverter(InlineModelConverterBase):
                 converter=converter,
             )
 
+        # Give InlineFormAdmin subclasses a chance to contribute extra fields
+        # onto the generated inline form, matching the behavior of the
+        # SQLAlchemy backend (see flask_admin/contrib/sqla/form.py).
+        child_form = info.postprocess_form(child_form)  # type: ignore[attr-defined]
+
         try:
             prop_name = reverse_field.related_name  # type: ignore[attr-defined]
         except AttributeError:
--- a/flask_admin/contrib/peewee/form.py
+++ b/flask_admin/contrib/peewee/form.py
@@ -331,7 +331,7 @@ class InlineModelConverter(InlineModelConverterBase):
         # Give InlineFormAdmin subclasses a chance to contribute extra fields
         # onto the generated inline form, matching the behavior of the
         # SQLAlchemy backend (see flask_admin/contrib/sqla/form.py).
-        child_form = info.postprocess_form(child_form)  # type: ignore[attr-defined]
+        child_form = info.postprocess_form(child_form)  # type: ignore[arg-type,assignment]
 
         try:
             prop_name = reverse_field.related_name  # type: ignore[attr-defined]
--- a/flask_admin/contrib/peewee/view.py
+++ b/flask_admin/contrib/peewee/view.py
@@ -110,16 +110,23 @@ class ModelView(BaseModelView):
 
     inline_model_form_converter: type[InlineModelConverter] = InlineModelConverter
     """
-        Inline model conversion class. If you need some kind of post-processing for
-        inline forms, you can customize behavior by doing something like this::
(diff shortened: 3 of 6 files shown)


ElLorans:
Although a fix, in practice this is a breaking change for peewee, isn't it?

PR #2892, opened 2026-05-05
Tabler UI

mariofix:
This PR adds Tabler UI into Flask-Admin, an early version.

fixes https://github.com/pallets-eco/flask-admin/issues/2443

See `tabler` folder included in examples

```py
admin = Admin(
    app,
    name="Example: Tabler",
    theme=TablerTheme(layout="condensed")
)
``` 

options allowed 
```py
layout: TablerLayout = "fluid" # "vertical", "fluid", "condensed"
# Defaults match Tabler's own defaults so existing deployments are unaffected.
theme: str = "light"  # "light" | "dark"
theme_primary: str = "blue"  # "blue" | "lime" | "azure" | "indigo" | …
theme_base: str = "gray"  # "gray" | "neutral" | "slate" | "zinc" | "stone"
theme_font: str = "sans-serif"  # "sans-serif" | "serif" | "monospace" | "comic"
theme_radius: str = "1"  # "0" | "0.5" | "1" | "1.5" | "2"
```
you can go to https://preview.tabler.io to see combinations

let me know what you think

PR #2909, opened 2026-05-24
Rebase "feat: replace x-editable with HTMX for inline editing" + other minors

ElLorans:
Rebasing https://github.com/pallets-eco/flask-admin/pull/2847 was not straightforward for me, so I am trying here before working on the original branch


ElLorans:
Test pass, example works. I would move this rebase and updates to #2909, then merge to master. We can work from there to unlock either #2892 (which looks done to me except for x editable not working) or #2444.
Finally, we can debate wether dropping XEditableWidget should happen in a minor or major release.

EDIT: not ready yet.
Errors are not rendered. E.g.: if the user submits a non unique field on a unique constrained column, only a console error is registered.
image

ElLorans:
Now err msgs appear
image

Issue #2917, opened 2026-06-14
MongoEngine `QueryAjaxModelLoader.format` crashes (HTTP 500) on unresolved `DBRef` values


hasansezertasan:
## Summary

`QueryAjaxModelLoader.format()` in `flask_admin.contrib.mongoengine.ajax`
accesses `model.pk` unconditionally. When MongoEngine cannot dereference a
`ReferenceField` (e.g. the referenced document has been deleted, or the
reference points to a different collection / database), it returns a raw
`bson.dbref.DBRef` instead of a `Document`. `DBRef` has no `.pk` attribute,
so the AJAX endpoint that powers the admin autocomplete widget crashes with
an `AttributeError`, surfacing as **HTTP 500** and breaking the edit form.

This was originally reported in #2036 (Sep 2020). The PR there bundled an
unrelated `Content-Disposition` fix, went stale, and now conflicts with
`master`. Opening a focused issue so the fix can be discussed and landed on
its own.

## Expected behavior

When an AJAX reference loader encounters a broken/unresolved reference, the
widget should degrade gracefully — either display the `DBRef` id as the
label (so the user can see *what* is broken and clear it) or return `None`
and let the form render an empty selection — not return a 500.

## Actual behavior

```
AttributeError: 'DBRef' object has no attribute 'pk'
```

…rendered as `500 Internal Server Error` from the `/admin/<view>/ajax/lookup/`
endpoint. The edit form's reference field becomes unusable until the broken
reference is cleaned up directly in the database.

## Root cause

`flask_admin/contrib/mongoengine/ajax.py:52`:

```python
def format(self, model: Document | None) -> tuple[str, str] | None:
    if not model:
        return None

    return (as_unicode(model.pk), as_unicode(model))
```

`model` here can be a `DBRef` when MongoEngine fails to dereference. There
is no type-check before `.pk` access.

## Notes on PR #2036

The original PR proposed:

```python
from bson.dbref import DBRef

def format(self, model):
    if not model:
        return None
    if not isinstance(model, DBRef):
        return (as_unicode(model.pk), as_unicode(model))
    else:
        return (as_unicode(model), as_unicode(model))
```

The shape is right, but a few things to settle before landing:

1. `as_unicode(dbref)` returns a string like `DBRef('coll', ObjectId('…'))`,
   which is not a usable id for the autocomplete value. The first tuple
   element should probably be `as_unicode(model.id)` (the `ObjectId` the
[...]

PR #2036, opened 2020-09-29
add filename when downloading FileField Content and Fix ReferenceField Reference on ajax ref

bekab95:
"download" file was downloaded without extension so this change will download file with original name and extension
@mrjoes  please review the change and merge

PR #2918, opened 2026-06-14
fix(mongoengine): set Content-Disposition on api_file_view downloads

### LINKED ISSUE #2916: MongoEngine `api_file_view` omits `Content-Disposition` — downloads have no filename or extension
## Summary

`ModelView.api_file_view()` in `flask_admin.contrib.mongoengine.view` serves
GridFS file content with `Content-Type` and `Content-Length` headers but **no
`Content-Disposition`**. Browsers therefore fall back to the URL's last path
segment (`/admin/<view>/api/file/`) when saving, producing a file named
`file` (or `file.html` after a content-sniff) with no extension — even though
GridFS knows the original filename.

This was originally reported in #2036 (Sep 2020). The PR there bundled an
unrelated `DBRef` fix, went stale, and now conflicts with `master`. Opening a
focused issue so the fix can be discussed and landed on its own.

## Expected behavior

Clicking a `FileField` download link in the admin should save the file with
its original name and extension as stored in GridFS (`data.filename`).

## Actual behavior

The response carries no `Content-Disposition` header, so the browser names
the download from the URL path:

```
HTTP/1.1 200 OK
Content-Type: image/png
Content-Length: 24576
```

→ saved as `file` (no extension), regardless of what was uploaded.

## Root cause

`flask_admin/contrib/mongoengine/view.py:727`:

```python
return Response(
    data.read(),
    content_type=data.content_type,
    headers={"Content-Length": data.length},
)
```

`data` is a `gridfs.GridOut` and exposes `.filename`, which is what should
drive `Content-Disposition`.

## Notes on PR #2036

The original PR proposed:

```python
headers={
    "Content-Length": data.length,
    "Content-disposition": "attachment; filename=%(filename)s" % {"filename": data.filename},
}
```

This works for ASCII filenames without spaces but breaks for:

- filenames containing spaces, commas, semicolons, or quotes (header parsing
  ambiguity);
- non-ASCII filenames (RFC 6266 requires `filename*=UTF-8''...` encoding);
- filenames containing CR/LF (header injection — `data.filename` ultimately
  comes from user upload).

A correct fix should either use `werkzeug.utils.send_file` (which handles
RFC 6266 encoding and range requests for free) or call
`werkzeug.datastructures.Headers.set` with the `filename` keyword, which
delegates to the same encoder.

```python
from werkzeug.utils import send_file

return send_file(
    data,                       # GridOut is a file-like object
    mimetype=data.content_type,
    download_name=data.filename,
    as_attachment=True,
)
```

`send_file` will set `Content-Length`, `Content-Type`, and an RFC-6266-safe
`Content-Disposition` for us, and supports `Range` requests — which the
current handcrafted `Response` does not.

## Reproducer

A full MRE requires a running MongoDB instance, so it's not as self-contained
as the PEP 723 scripts used elsewhere. The minimal path:

1. Define a MongoEngine `Document` with a `FileField`.
2. Register a `ModelView` for it.
3. Upload any file (e.g. `report.pdf`) through the admin.
4. Click the download link on the details view → file is saved as `file`
   instead of `report.pdf`.

The response headers can be inspected directly with `curl -I` against the
`/admin/<view>/api/file/?id=...&coll=...` URL to confirm
`Content-Disposition` is absent.

## Environment

- `mongoengine` (any recent version).
- Any modern browser, or `curl`.

## Related

- #2036 — original report and bundled PR (will be closed in favour of two
  focused PRs).

---

> Disclosure: I drafted this issue with help from Claude Code while triaging
> stale PR #2036; the references above were verified manually.

hasansezertasan:
## Summary

Closes #2916.

`ModelView.api_file_view()` in `flask_admin.contrib.mongoengine.view` served GridFS file content with `Content-Type` and `Content-Length` headers but **no `Content-Disposition`**. Browsers therefore named the saved file from the URL's last path segment (`/admin/<view>/api/file/`) — producing `file` (or `file.html` after a content-sniff) with no extension, even though GridFS already knows the original filename.

This PR switches the handler to `flask.send_file` with `download_name=data.filename`:

```python
return send_file(
    data,
    mimetype=data.content_type,
    download_name=data.filename,
    as_attachment=False,
)
```

The response now carries `Content-Disposition: inline; filename="<original>"`, which:

- preserves the **existing inline-render behavior** for images/PDFs (no behavioral regression for users who clicked a download link expecting the browser to display the file);
- supplies the original filename and extension when the user does choose "Save as";
- handles RFC 6266 encoding for non-ASCII filenames automatically (the original PR #2036's handcrafted `%(filename)s` interpolation did not);
- supports `Range` requests for free.

## Attribution

Originally reported and patched by [@bekab95](https://github.com/bekab95) in #2036 (Sep 2020). That PR bundled this change with an unrelated `DBRef` fix, went stale, and now conflicts with `master`. I've split the report into two focused issues — #2916 (this PR) and #2917 — and credited @bekab95 in the commit (`Co-authored-by`) and the changelog entry.

A companion PR for #2917 will follow.

## Test plan

- [x] Added `test_api_file_view_sets_content_disposition` in `flask_admin/tests/mongoengine/test_basic.py`. The test uploads a `report.txt` blob through a `FileField`, hits `/admin/<view>/api/file/?id=...`, and asserts `report.txt` appears in `Content-Disposition`.
- [x] Full MongoEngine test file passes locally against `mongo:7`.
- [x] Existing tests still pass (no behavioral change beyond the new header).

## Notes

`pyproject.toml` gains one warning filter for `GridOut property 'contentType' is deprecated and will be removed in PyMongo 5.0`. That deprecation is raised by `data.content_type` — which the pre-existing code already used; this PR simply added the first test that exercises that path. Migrating off the deprecated property belongs in a separate change.

---

> Disclosure: I drafted this PR with help from Claude Code while triaging stale PR #2036; the diff and the test were verified manually against a live MongoDB instance.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -7,7 +7,7 @@ Changelog
 Bugfixes:
 * Fix encoding for editing file in FileAdmin. Now it uses UTF-8 and accepts non-ASCII characters.
 * SQLAlchemy backend: ``conv_ARRAY`` now infers the array element's ``python_type`` and passes it through as the ``Select2TagsField`` ``coerce`` callable. Saving a Postgres ``ARRAY(Integer)`` / ``ARRAY(Float)`` column no longer fails with ``column "x" is of type integer[] but expression is of type text[]`` (closes #1724).
-* MongoEngine backend: ``api_file_view`` now emits a ``Content-Disposition`` header carrying the original GridFS filename, so downloaded files keep their name and extension instead of being saved as ``file`` (closes #2916). Originally reported and patched by `@bekab95 <https://github.com/bekab95>`_ in #2036.
+* MongoEngine fileadmin backend: downloading GridFS files now preserves their name and extension instead of saving them as ``file`` (closes #2916).
 
 Type hints:
 * Type hints added to all functions and methods (some using `typing.Any` where full typing not yet available)
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -7,6 +7,7 @@ Changelog
 Bugfixes:
 * Fix encoding for editing file in FileAdmin. Now it uses UTF-8 and accepts non-ASCII characters.
 * SQLAlchemy backend: ``conv_ARRAY`` now infers the array element's ``python_type`` and passes it through as the ``Select2TagsField`` ``coerce`` callable. Saving a Postgres ``ARRAY(Integer)`` / ``ARRAY(Float)`` column no longer fails with ``column "x" is of type integer[] but expression is of type text[]`` (closes #1724).
+* MongoEngine backend: ``api_file_view`` now emits a ``Content-Disposition`` header carrying the original GridFS filename, so downloaded files keep their name and extension instead of being saved as ``file`` (closes #2916). Originally reported and patched by `@bekab95 <https://github.com/bekab95>`_ in #2036.
 
 Type hints:
 * Type hints added to all functions and methods (some using `typing.Any` where full typing not yet available)
--- a/flask_admin/contrib/mongoengine/view.py
+++ b/flask_admin/contrib/mongoengine/view.py
@@ -9,6 +9,7 @@ from flask import abort
 from flask import flash
 from flask import request
 from flask import Response
+from flask import send_file
 from mongoengine import Document
 from mongoengine import QuerySet
 from mongoengine.connection import get_db
@@ -724,10 +725,11 @@ class ModelView(BaseModelView):
         if not data:
             abort(404)
 
-        return Response(
-            data.read(),
-            content_type=data.content_type,
-            headers={"Content-Length": data.length},  # type: ignore[arg-type]
+        return send_file(
+            data,
+            mimetype=data.content_type,
+            download_name=data.filename,
+            as_attachment=False,
         )
 
     # Default model actions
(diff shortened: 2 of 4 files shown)


doc/changelog.rst
```diff
@@ -7,6 +7,7 @@ Changelog
 Bugfixes:
 * Fix encoding for editing file in FileAdmin. Now it uses UTF-8 and accepts non-ASCII characters.
 * SQLAlchemy backend: ``conv_ARRAY`` now infers the array element's ``python_type`` and passes it through as the ``Select2TagsField`` ``coerce`` callable. Saving a Postgres ``ARRAY(Integer)`` / ``ARRAY(Float)`` column no longer fails with ``column "x" is of type integer[] but expression is of type text[]`` (closes #1724).
+* MongoEngine backend: ``api_file_view`` now emits a ``Content-Disposition`` header carrying the original GridFS filename, so downloaded files keep their name and extension instead of being saved as ``file`` (closes #2916). Originally reported and patched by `@bekab95 <https://github.com/bekab95>`_ in #2036.
```
ElLorans:
```suggestion
* MongoEngine fileadmin backend: downloading GridFS files now preserves their name and extension instead of saving them as ``file`` (closes #2916).
```

ElLorans (approved):

ElLorans:
I would simplify the changelog. If GridOut property 'contentType' is deprecated and will be removed in PyMongo 5.0, do we need to start working on the fix?

PR #2919, opened 2026-06-14
fix(mongoengine): handle unresolved DBRef in QueryAjaxModelLoader.format

### LINKED ISSUE #2917: MongoEngine `QueryAjaxModelLoader.format` crashes (HTTP 500) on unresolved `DBRef` values

hasansezertasan:
## Summary

Closes #2917.

`QueryAjaxModelLoader.format()` in `flask_admin.contrib.mongoengine.ajax` accessed `model.pk` unconditionally. When MongoEngine cannot dereference a `ReferenceField` (target document deleted, or pointing at a different collection / database), it returns a raw `bson.dbref.DBRef` instead of a `Document`. `DBRef` has no `.pk` attribute, so the AJAX endpoint backing the admin autocomplete widget crashed with:

```
AttributeError: 'DBRef' object has no attribute 'pk'
```

…rendered as HTTP 500 from `/admin/<view>/ajax/lookup/`, making the edit form unusable until the broken reference was cleaned up directly in the database.

## Change

```python
def format(self, model):
    if not model:
        return None
    if isinstance(model, DBRef):
        return (as_unicode(model.id), f"(missing: {model.collection}/{model.id})")
    return (as_unicode(model.pk), as_unicode(model))
```

Two design notes:

1. **The posted-back value is `model.id`, not `as_unicode(dbref)`.** The original PR #2036 used `as_unicode(model)`, which would emit `DBRef('coll', ObjectId('…'))` — not a value the persistence layer can resolve. Using the underlying `ObjectId` keeps the form round-trippable: if the user leaves the broken reference in place, nothing worse happens; if they pick a new value, the new id replaces it cleanly.
2. **The label is `(missing: <collection>/<id>)`** so users see *what* is broken rather than the raw `DBRef` repr leaking through to the UI.

## Attribution

Originally reported and patched by [@bekab95](https://github.com/bekab95) in #2036 (Sep 2020). That PR bundled this change with an unrelated `Content-Disposition` fix, went stale, and now conflicts with `master`. I've split the report into two focused issues — #2916 (companion PR #2918) and #2917 (this PR) — and credited @bekab95 in the commit (`Co-authored-by`) and the changelog entry.

## Test plan

- [x] Added `test_query_ajax_model_loader_format_handles_dbref` in `flask_admin/tests/mongoengine/test_basic.py`. The test constructs a `DBRef` pointing at a non-existent document and asserts `format()` returns a 2-tuple with the `ObjectId` as the value and a `"missing"` marker in the label — instead of raising `AttributeError`.
- [x] Full MongoEngine test file passes locally.
- [x] No behavioral change for the happy path (well-resolved `Document` values still go through the original branch).

---

> Disclosure: I drafted this PR with help from Claude Code while triaging stale PR #2036; the diff and the test were verified manually.


doc/changelog.rst
```diff
@@ -7,6 +7,7 @@ Changelog
 Bugfixes:
 * Fix encoding for editing file in FileAdmin. Now it uses UTF-8 and accepts non-ASCII characters.
 * SQLAlchemy backend: ``conv_ARRAY`` now infers the array element's ``python_type`` and passes it through as the ``Select2TagsField`` ``coerce`` callable. Saving a Postgres ``ARRAY(Integer)`` / ``ARRAY(Float)`` column no longer fails with ``column "x" is of type integer[] but expression is of type text[]`` (closes #1724).
+* MongoEngine backend: ``QueryAjaxModelLoader.format`` now handles unresolved ``DBRef`` values gracefully instead of crashing with ``AttributeError: 'DBRef' object has no attribute 'pk'`` (which surfaced as HTTP 500 from the autocomplete endpoint when a ``ReferenceField`` target was deleted). Broken references are shown as ``(missing: <collection>/<id>)`` so users can clear them (closes #2917). Originally reported and patched by `@bekab95 <https://github.com/bekab95>`_ in #2036.
```
ElLorans:
```suggestion
* MongoEngine backend: ``QueryAjaxModelLoader.format`` now handles unresolved ``DBRef`` values gracefully instead of crashing with ``AttributeError: 'DBRef' object has no attribute 'pk'`` (which surfaced as HTTP 500 from the autocomplete endpoint when a ``ReferenceField`` target was deleted). Broken references are shown as ``(missing: <collection>/<id>)`` so users can clear them (closes #2917).
```

ElLorans (approved):

flask_admin/tests/mongoengine/test_basic.py
```diff
@@ -147,6 +149,28 @@ def test_model(app: Flask, db: t.Any, admin: Admin) -> None:
...
+    missing_id = ObjectId()
+    dbref = DBRef(collection="ref_target", id=missing_id)
+
+    result = loader.format(dbref)
+
+    assert result is not None
+    value, label = result
+    assert value == str(missing_id)
+    assert "missing" in label
+    assert str(missing_id) in label
```
ElLorans, line 173:
```suggestion
    assert str(missing_id) in label


```

PR #2922, opened 2026-06-18
fix(filters): BaseTimeBetweenFilter.validate() returns False on invalid input

Noethix55555:
`BaseTimeBetweenFilter.validate()` in `flask_admin/model/filters.py` had a bare `raise` in its `except ValueError` block, followed by unreachable `return False`. This caused an unhandled `ValueError` (HTTP 500) whenever a malformed time range string — e.g. missing the `" to "` separator, garbage values like `notavalidtime`, or out-of-range times — was submitted via the URL query string.

The two analogous filters in the same file already handle this correctly:
- `BaseDateBetweenFilter.validate()` — `return False` on `ValueError`
- `BaseDateTimeBetweenFilter.validate()` — `return False` on `ValueError`

**Fix:** remove the `raise` so `ValueError` results in `return False`, matching the existing pattern.

A test covering valid and invalid inputs is added to `flask_admin/tests/test_model.py`.

### DIFF
--- a/flask_admin/tests/test_model.py
+++ b/flask_admin/tests/test_model.py
@@ -914,3 +914,26 @@ def test_form_submit(app: Flask, admin: Admin, url: str, age: int, msg: str) -> 
     assert rv.status_code == 200
     data = rv.data.decode("utf-8")
     assert all([part in data for part in msg.split("|")])
+
+
+def test_time_between_filter_validate_invalid_input() -> None:
+    """BaseTimeBetweenFilter.validate() must return False for bad input, not raise."""
+    flt = filters.BaseTimeBetweenFilter("time_col")
+
+    # Valid range returns True
+    assert flt.validate("08:00:00 to 17:00:00") is True
+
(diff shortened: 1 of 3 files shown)


ElLorans:
Thanks! Can you add a line to the changelog, too?

ElLorans:
Thanks, this looks correct and valuable. I'll test it in the next few weeks and check no code actually expects a ValueError to be returned

flask_admin/tests/test_model.py
```diff
@@ -914,3 +914,26 @@ def test_form_submit(app: Flask, admin: Admin, url: str, age: int, msg: str) ->
     assert rv.status_code == 200
     data = rv.data.decode("utf-8")
     assert all([part in data for part in msg.split("|")])
+
+
+def test_time_between_filter_validate_invalid_input() -> None:
```
ElLorans:
```suggestion
def test_base_time_between_filter_validate() -> None:
```

PR #2926, opened 2026-06-19
feat(fileadmin): enhance modal view response on invalid inputs

### LINKED ISSUE #2925: Wrong html response on invalid input with FileAdmin (Modal view)
If the user enter an invalid name while renaming a file on a modal view, then the error message should be presented into the regular view (not the modal one) which is not the case currently. instead, it shows the error in the modal page which doesn't have a styling!!

 here is what i got when i tried to enter an invalid name: 

Image

this is happening becaues the `<form>` submits the value into the backend with a parameter `modal=True` , so the response would be returned as a modal !

samialfattani:
Fixes #2925 

image

### DIFF
--- a/flask_admin/templates/bootstrap4/admin/file/modals/form.html
+++ b/flask_admin/templates/bootstrap4/admin/file/modals/form.html
@@ -7,14 +7,19 @@
     {% block header %}<h3>{{ header_text }}</h3>{% endblock %}
     <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
   </div>
-  <div class="modal-body">
-    {% block fa_form %}
 
-      {# to exclude 'modal' arg  #}
-      {% set p = url_for(request.endpoint, path=path)  %}
-      {{ lib.render_form(form, dir_url, action=p ) }}
-    {% endblock %}
-  </div>
+  {% block fa_form %}
+
+    {% call lib.form_tag(action=url_for(request.endpoint, path=path)) %}
+      <div class="modal-body">
+          {{ lib.render_form_fields(form) }}
+      </div>
+      <div class="modal-footer">
+            {{ lib.render_form_buttons(dir_url, extra=None, is_modal=True) }}
+      </div>
+    {% endcall %}
+  {% endblock %}
+
 {% endblock %}
 
 {% block tail %}
--- a/flask_admin/templates/bootstrap4/admin/file/modals/form.html
+++ b/flask_admin/templates/bootstrap4/admin/file/modals/form.html
@@ -10,7 +10,7 @@
   <div class="modal-body">
     {% block fa_form %}
 
-      {# to execlude 'modal' arg  #}
+      {# to exclude 'modal' arg  #}
       {% set p = url_for(request.endpoint, path=path)  %}
       {{ lib.render_form(form, dir_url, action=p ) }}
     {% endblock %}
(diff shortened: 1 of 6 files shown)


flask_admin/templates/bootstrap4/admin/file/modals/form.html
```diff
@@ -9,11 +9,16 @@
   </div>
   <div class="modal-body">
     {% block fa_form %}
-      {{ lib.render_form(form, dir_url, action=request.url, is_modal=True) }}
+
+      {# to execlude 'modal' arg  #}
```
ElLorans: Exclude*

ElLorans:
```suggestion
      {# to exclude 'modal' arg  #}
```

flask_admin/templates/bootstrap4/admin/file/modals/form.html
```diff
@@ -9,11 +9,16 @@
   </div>
   <div class="modal-body">
     {% block fa_form %}
-      {{ lib.render_form(form, dir_url, action=request.url, is_modal=True) }}
+
+      {# to exclude 'modal' arg  #}
```
ElLorans: why excluding the modal arg? dropping `is_modal` will change other rendering behavior in the render_form macro (CSS classes, hidden fields, etc.) for all modal forms (upload/mkdir/rename/edit), not just the rename-with-invalid-name case the changelog mentions

flask_admin/tests/fileadmin/test_fileadmin.py
ElLorans, line 1: I do not understand what the tests are for: I see no post request

flask_admin/tests/fileadmin/test_fileadmin.py
```diff
@@ -160,6 +165,64 @@ class MyFileAdmin(fileadmin_class):  # type: ignore[valid-type, misc]
...
+            admin.add_view(view)
+
+            client = app.test_client()
+
+            rv = client.get(f"/admin/mymodalfileadmin{url}")
+            assert rv.status_code == 200
+            data = rv.data.decode("utf-8")
+
+            for ex in expected:
+                assert ex in data
```
ElLorans: this is too fragile: we should check for equality, not `in`

flask_admin/templates/bootstrap4/admin/file/modals/form.html
samialfattani: yes it will change the render form, actually this is the same approach that is used with create/edit modal forms. anyway i will change the code to match the ones that are used in create/edit modals.

flask_admin/tests/fileadmin/test_fileadmin.py
samialfattani: in those tests every test case has multiple expected texts and all of them should be `in data` to pass the whole test case. they are not fragile because the passing depends on different expected outputs

flask_admin/tests/fileadmin/test_fileadmin.py
samialfattani: sorry i forget to add post request, i will add them though. but also the get request is helpful to make sure that Modal content is being retrieved as it is expected.

flask_admin/templates/bootstrap4/admin/file/modals/form.html
ElLorans: If there is a reason, we can change it, but it is nice to explain why! and it is always useful to understand why you came up with some solutions

samialfattani: **Explanation:** just like any other modal forms in flask-admin, they always submit the form to its target without `is_modal` parameter which was not the case here. if the `is_moadl` didn't excluded then the response will be like the issue #2925. having `is_modal` in the target url is the root cause of of #2925. that's why it has to be excluded

PR #2934, opened 2026-07-06
invert sorting arrows in list view

### LINKED ISSUE #2933: Sorting arrows are inversed in list view
I noticed that sorting arrows in admin list view instead of indicating current sort state indicate the opposite. If you sort by column in descending order you get arrow pointing UP, and if order is ascending then arrow points DOWN. It seems rather counter intuitive and I guess that usually the arrows indicate the current state of applied sorting. 

I checked excel, js datatables, django admin to be sure and all everywhere down pointing arrow is used for descending, and up pointing for ascending sorting. Wrote about this in official discord channel to discussed and was told to open the issue. I found the place that needs to be changed to fix this, so I will add pull request addressing this this issue. 

Here is the place in templates code that chooses the direction of the arrows (similar in model and file admins)

```
{% if admin_view.is_sortable(c) %}
    {% if sort_column == column %}
        <a href="{{ sort_url(column, True) }}" title="{{ _gettext('Sort by %(name)s', name=name) }}">
            {{ name }}
            {% if sort_desc %}
                <span class="fa fa-chevron-up glyphicon glyphicon-chevron-up"></span>
            {% else %}
                <span class="fa fa-chevron-down glyphicon glyphicon-chevron-down"></span>
            {% endif %}
        </a>
    {% else %}
        <a href="{{ sort_url(column) }}" title="{{ _gettext('Sort by %(name)s', name=name) }}">{{ name }}</a>
    {% endif %}
{% else %}

```

Tikrong:
I found that in admin list view the direction of sorting arrow is inversed and opennen issue about it after asking in official discord channel. In short: If you sort by column in descending order you get arrow pointing UP, and if order is ascending then arrow points DOWN. Usually its done the other way around so I proposed to fix this.

There are 2 places in code that set the sorting in list view, in model/list.html and file/list.html, and I swaped the icons for descending and ascending order there.

fixes #2933

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -7,7 +7,7 @@ Changelog
 Bugfixes:
 * Fix encoding for editing file in FileAdmin. Now it uses UTF-8 and accepts non-ASCII characters.
 * SQLAlchemy backend: ``conv_ARRAY`` now infers the array element's ``python_type`` and passes it through as the ``Select2TagsField`` ``coerce`` callable. Saving a Postgres ``ARRAY(Integer)`` / ``ARRAY(Float)`` column no longer fails with ``column "x" is of type integer[] but expression is of type text[]`` (closes #1724).
-* Fix sorting arrow direction in admin list view. Now it reflects the current sorting state. (closes #2933).
+* Fix sorting arrow direction in admin list view. Now it reflects the current sorting state (closes #2933).
 
 Type hints:
 * Type hints added to all functions and methods (some using `typing.Any` where full typing not yet available)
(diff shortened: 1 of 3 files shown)


ElLorans (approved):

ElLorans:
Thanks for the PR! Can you add a line to the changelog under [unreleased]?

hasansezertasan (approved):
LGTM

Tikrong:
> Thanks for the PR! Can you add a line to the changelog under [unreleased]?

Done, added a line about the fix. Thank you for quick feedback.

PR #2935, opened 2026-07-07
fix(mongoengine): stop reading deprecated GridOut.contentType

### LINKED ISSUE #2920: chore(mongoengine): migrate off deprecated GridOut.contentType ahead of PyMongo 5.0
## Context

PyMongo deprecated the `GridOut.contentType` property and will remove it in PyMongo 5.0:

> `GridOut property 'contentType' is deprecated and will be removed in PyMongo 5.0`

Although flask-admin's code reads `data.content_type` (snake_case), the underlying GridFS `GridOut` object still resolves this through the deprecated `contentType` accessor, so the `DeprecationWarning` fires on every read.

Flask-admin's test suite runs pytest with `filterwarnings = ["error", ...]`, which promotes every warning to an error. PR #2918 added a `default:GridOut property 'contentType' is deprecated...:DeprecationWarning` entry to `pyproject.toml`, which downgrades this specific warning back to the default (print-once) behavior so the new MongoEngine test could pass. The warning itself is not silenced — it is just exempted from the strict-errors policy. This issue tracks the actual migration off the deprecated property.

## Affected call sites (read path — `GridOut` / `GridFSProxy`)

- `flask_admin/contrib/mongoengine/view.py` — `api_file_view` reads `data.content_type` where `data = fs.get(...)` is a raw `GridOut`
- `flask_admin/contrib/mongoengine/widgets.py` — `MongoFileInput` reads `field.data.content_type` where `field.data` is a `GridFSProxy`
- `flask_admin/contrib/mongoengine/typefmt.py` — `grid_formatter` reads `value.content_type` where `value` is a `GridFSProxy`/`GridOut`

## Not affected

- `flask_admin/contrib/mongoengine/fields.py` — this reads `content_type` from a Werkzeug `FileStorage` (the uploaded file) on the *write* path and passes it into `GridFSProxy.put`/`replace`. Not a `GridOut` attribute access; unrelated to the PyMongo deprecation.

## Proposed approach

1. Two of the three read sites go through MongoEngine's `GridFSProxy` rather than raw PyMongo. Investigate whether MongoEngine already exposes (or plans to expose) a non-deprecated accessor; if so, use it directly. If not, fall back to reading the content type from the GridFS document's `metadata` dict (PyMongo's recommended location in 5.0) with a `mimetypes.guess_type(filename)` fallback when missing.
2. Centralize the lookup in a small helper (e.g. `_gridfs_content_type(data)`) so all three read-side call sites share one implementation and the raw-`GridOut` site and `GridFSProxy` sites stay consistent.
3. Remove the `default:GridOut property 'contentType'...` entry from `pyproject.toml` once the migration lands and verify the suite still passes under `filterwarnings = ["error", ...]`.
4. Verify the test added in #2918 (`test_api_file_view_sets_content_disposition`) still asserts the correct `Content-Type` (`text/plain`) after the migration.

## Out of scope

- Behavioral changes to `api_file_view` beyond the content-type source.
- Any unrelated PyMongo or MongoEngine deprecations.

## References

- PR #2918 (introduced the warning filter)
- PyMongo deprecation notice: GridOut `contentType` → removed in 5.0

---

> Disclosure: I drafted this issue with help from Claude Code while reviewing feedback on PR #2918; the affected call sites were verified manually against the current `master`.

hasansezertasan:
## Summary

Closes #2920.

PyMongo deprecated ``GridOut.contentType`` and will remove it in PyMongo 5.0. Flask-admin's MongoEngine backend reads it in three places on the file read path:

- ``ModelView.api_file_view`` — GridFS download response
- ``MongoFileInput`` — form widget rendering
- ``grid_formatter`` (typefmt) — list/detail formatter rendering

This PR routes all three through a new ``gridfs_content_type()`` helper in ``flask_admin/contrib/mongoengine/helpers.py`` that resolves the MIME type without touching the deprecated property:

1. ``data.metadata["content_type"]`` — the PyMongo 5.0-compatible location.
2. ``data._file["contentType"]`` — legacy fallback for existing deployments whose documents still store the content type at the deprecated top-level field. Reaching into the raw BSON doc avoids the deprecation warning while preserving behavior for data written before this change.
3. ``mimetypes.guess_type(data.filename)`` — best-effort from filename.
4. ``None`` if no source resolves.

The write path (``MongoFileField`` in ``fields.py``) is intentionally out of scope — it reads ``content_type`` from a Werkzeug ``FileStorage``, not from a ``GridOut``, and is unrelated to the PyMongo deprecation.

## Coordination with #2918

PR #2918 (unmerged) adds a ``default:GridOut property 'contentType'...`` entry to ``pyproject.toml``'s ``filterwarnings`` so a new test can pass under the suite's strict ``[""error""]`` policy. Whichever PR lands second should drop that entry:

- If #2918 lands first: a follow-up commit here should remove the ``filterwarnings`` line.
- If this lands first: #2918 should drop that entry from its diff.

## Test plan

- [x] Added ``flask_admin/tests/mongoengine/test_helpers.py`` with 7 unit tests covering the full resolution order against lightweight ``types.SimpleNamespace`` fakes — no live MongoDB needed. All pass.
- [x] Existing MongoEngine tests remain unchanged (behavior is preserved for both legacy and forward-compatible data layouts).

## Notes

Once PyMongo 5.0 ships and ``GridOut.contentType`` is removed, ``_legacy_content_type`` becomes dead code and can be deleted in a follow-up.

---

> Disclosure: I drafted this PR with help from Claude Code while addressing the deprecation flagged during review of #2918; the resolution order was verified against PyMongo 4.17 and MongoEngine 0.29.3, and the unit tests were run locally.

### DIFF
--- a/flask_admin/contrib/mongoengine/typefmt.py
+++ b/flask_admin/contrib/mongoengine/typefmt.py
@@ -11,6 +11,7 @@ from flask_admin.model.typefmt import BASE_FORMATTERS
 from flask_admin.model.typefmt import list_formatter
 
 from . import helpers
+from .helpers import gridfs_content_type
 
 if t.TYPE_CHECKING:
     from . import ModelView
@@ -34,7 +35,7 @@ def grid_formatter(view: ModelView, value: GridOut) -> Markup | str:
             "url": view.get_url(".api_file_view", **args),
             "name": escape(value.name),
             "size": value.length // 1024,
-            "content_type": escape(value.content_type),
+            "content_type": escape(gridfs_content_type(value) or ""),
         }
     )
 
--- a/flask_admin/contrib/mongoengine/typefmt.py
+++ b/flask_admin/contrib/mongoengine/typefmt.py
@@ -34,7 +34,7 @@ def grid_formatter(view: ModelView, value: GridOut) -> Markup | str:
             "url": view.get_url(".api_file_view", **args),
             "name": escape(value.name),
             "size": value.length // 1024,
-            "content_type": escape(gridfs_content_type(value) or ""),
+            "content_type": escape(helpers.gridfs_content_type(value) or ""),
         }
     )
 
--- a/flask_admin/contrib/mongoengine/typefmt.py
+++ b/flask_admin/contrib/mongoengine/typefmt.py
@@ -11,7 +11,6 @@ from flask_admin.model.typefmt import BASE_FORMATTERS
 from flask_admin.model.typefmt import list_formatter
 
 from . import helpers
-from .helpers import gridfs_content_type
 
 if t.TYPE_CHECKING:
     from . import ModelView
--- a/flask_admin/contrib/mongoengine/widgets.py
+++ b/flask_admin/contrib/mongoengine/widgets.py
@@ -10,7 +10,6 @@ from flask_admin.helpers import get_url
 
 from ..._types import _T_MONGOENGINE_FIELD_PROTOCOL
 from . import helpers
-from .helpers import gridfs_content_type
 
 
 class MongoFileInput:
--- a/flask_admin/contrib/mongoengine/widgets.py
+++ b/flask_admin/contrib/mongoengine/widgets.py
@@ -10,6 +10,7 @@ from flask_admin.helpers import get_url
 
 from ..._types import _T_MONGOENGINE_FIELD_PROTOCOL
 from . import helpers
+from .helpers import gridfs_content_type
 
 
 class MongoFileInput:
@@ -33,7 +34,7 @@ class MongoFileInput:
 
             placeholder = self.template % {
                 "name": escape(data.name),
-                "content_type": escape(data.content_type),
+                "content_type": escape(gridfs_content_type(data) or ""),
                 "size": data.length // 1024,
                 "marker": f"_{field.name}-delete",
             }
--- a/flask_admin/contrib/mongoengine/widgets.py
+++ b/flask_admin/contrib/mongoengine/widgets.py
@@ -34,7 +34,7 @@ class MongoFileInput:
 
             placeholder = self.template % {
                 "name": escape(data.name),
-                "content_type": escape(gridfs_content_type(data) or ""),
+                "content_type": escape(helpers.gridfs_content_type(data) or ""),
                 "size": data.length // 1024,
                 "marker": f"_{field.name}-delete",
             }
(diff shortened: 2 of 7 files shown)


ElLorans:
Can you do rebase -i to remove commits now merged with #2920 ?

hasansezertasan:
@ElLorans Done — rebased onto `master`, dropping the merge commit so the PR is now a single self-contained commit (`f22994cb`).

While resolving the rebase, two changes were needed:

- **`view.py`**: #2918 already switched `api_file_view` to `send_file(..., download_name=...)` (the #2916 Content-Disposition fix). I kept that structure and only routed the MIME type through the new helper — `mimetype=gridfs_content_type(data)` — instead of `data.content_type`. Note `GridOut.content_type` (snake_case) is also deprecated (it maps to the same `contentType` descriptor), so the helper is still needed on the download path.
- **`pyproject.toml`**: per the coordination note, since #2918 landed first I removed the `GridOut.contentType` `filterwarnings` entry. No runtime code hits the deprecated property anymore, so it's dead. Verified against a live MongoDB with the suite's warnings-as-error policy: the api_file_view regression test, the 7 helper tests, and the full mongoengine suite all pass with no deprecation warning.

flask_admin/contrib/mongoengine/typefmt.py
```diff
@@ -11,6 +11,7 @@
 from flask_admin.model.typefmt import list_formatter
 
 from . import helpers
+from .helpers import gridfs_content_type
```
ElLorans:
```suggestion
```

flask_admin/contrib/mongoengine/typefmt.py
```diff
@@ -34,7 +35,7 @@ def grid_formatter(view: ModelView, value: GridOut) -> Markup | str:
             "url": view.get_url(".api_file_view", **args),
             "name": escape(value.name),
             "size": value.length // 1024,
-            "content_type": escape(value.content_type),
+            "content_type": escape(gridfs_content_type(value) or ""),
```
ElLorans:
```suggestion
            "content_type": escape(helpers.gridfs_content_type(value) or ""),
```

flask_admin/contrib/mongoengine/widgets.py
```diff
@@ -33,7 +34,7 @@ def __call__(self, field: _T_MONGOENGINE_FIELD_PROTOCOL, **kwargs: t.Any) -> Mar
 
             placeholder = self.template % {
                 "name": escape(data.name),
-                "content_type": escape(data.content_type),
+                "content_type": escape(gridfs_content_type(data) or ""),
```
ElLorans:
```suggestion
                "content_type": escape(helpers.gridfs_content_type(data) or ""),
```

flask_admin/contrib/mongoengine/widgets.py
```diff
@@ -10,6 +10,7 @@
 
 from ..._types import _T_MONGOENGINE_FIELD_PROTOCOL
 from . import helpers
+from .helpers import gridfs_content_type
```
ElLorans:
```suggestion
```

ElLorans:
Lokks good. I am adding 4 suggestions to consolidate the helper imports but it is a style choice more than anything else.

Currently In `typefmt.py` and `widgets.py` you're importing both

```
from . import helpers
from .helpers import gridfs_content_type
```

we can remove the second line and change the function call from `gridfs_content_type` to `helpers.gridfs_content_type`

ElLorans (approved):

hasansezertasan:
Thanks @ElLorans, I did what you recommended. Looks better 🤓.

PR #2939, opened 2026-07-30
fix(peewee): return None from get_one() when no record matches

### LINKED ISSUE #2924: bug(peewee): `get_one()` raises uncaught `DoesNotExist` (HTTP 500) for any absent or malformed `id`
# bug(peewee): `get_one()` raises uncaught `DoesNotExist` (HTTP 500) for any absent or malformed `id`

## Summary

`ModelView.get_one()` in `flask_admin.contrib.peewee.view` calls
`self.model.get(**{self._primary_key: id})` for single-PK models. Peewee's
`Model.get()` raises `DoesNotExist` whenever the query matches no row, and
`get_one` does not catch it. Because `details_view` / `edit_view` /
`delete_view` in `flask_admin/model/base.py` call `get_one(id)` with no
exception guard (they expect `None` for "not found"), the exception surfaces
as an unhandled **HTTP 500**.

This is the peewee analogue of #2915 (sqla), found while auditing the other
contribs for the same failure mode. It is **broader** than the sqla bug: in
peewee, even a *well-formed but absent* id (`?id=99999`) returns 500, not just
a malformed one (`?id=1,2`). By contrast the mongoengine and pymongo contribs
handle both cases gracefully (they translate backend-specific errors / misses
into a `None` return, which renders the existing "Record does not exist."
flash).

## Expected behavior

A malformed or absent `id` should produce the existing "Record does not
exist." flash + redirect to the list view — not an unhandled `500`.

## Actual behavior

```
peewee.WidgetDoesNotExist: <Model: Widget> instance matching query does not exist:
SQL: SELECT "t1"."id", "t1"."name" FROM "widget" AS "t1" WHERE ("t1"."id" = ?) LIMIT ? OFFSET ?
Params: [99999, 1, 0]
```

…rendered as a `500 Internal Server Error` response.

## Minimal reproducer (PEP 723)

Save as `repro.py` and run with `uv run repro.py`. No venv setup needed.

```python
# /// script
# requires-python = ">=3.10"
# dependencies = [
#     "flask-admin[peewee]",
# ]
# ///
"""MRE for flask-admin peewee get_one() crash: any missing/malformed `id` -> 500."""

import peewee
from flask import Flask
from flask_admin import Admin
from flask_admin.contrib.peewee import ModelView

db = peewee.SqliteDatabase(":memory:")


class Widget(peewee.Model):
    name = peewee.CharField()

    class Meta:
        database = db


app = Flask(__name__)
app.config["SECRET_KEY"] = "x"
app.config["WTF_CSRF_ENABLED"] = False


class WidgetView(ModelView):
    can_view_details = True


admin = Admin(app)
admin.add_view(WidgetView(Widget, endpoint="widget"))

db.connect()
db.create_tables([Widget])
Widget.create(name="alpha")
Widget.create(name="beta")

client = app.test_client()

for label, query in [
    ("id=1 (valid)", "id=1"),
    ("id=1,2 (malformed for single-PK model)", "id=1,2"),
    ("id=99999 (valid but absent)", "id=99999"),
]:
    rv = client.get(f"/admin/widget/details/?{query}")
    print(f"=== {label} ===")
    print("status:", rv.status_code)
    print()
```

Output:

```
=== id=1 (valid) ===
status: 200

=== id=1,2 (malformed for single-PK model) ===
status: 500

=== id=99999 (valid but absent) ===
status: 500
```

## Notes

- Root cause is shared with #2915: `details_view`/`edit_view`/`delete_view`
  delegate malformed/absent-id handling entirely to each backend's `get_one`,
  and only the two mongo backends currently translate failures into `None`.
- A fix could either make `peewee.ModelView.get_one` return `None` on
  `DoesNotExist` (and a malformed id), or centralize the guard in
  `model/base.py` so all backends behave consistently. Worth deciding
  alongside #2915.

timyjsong:
`ModelView.get_one()` in the peewee backend calls `self.model.get(...)`, which raises peewee's `DoesNotExist` when no row matches. Nothing catches it, so requesting a record that is absent, or whose `id` cannot match the primary key, returns an unhandled HTTP 500.

Against the existing peewee test models:

```
GET /admin/model2/details/?url=%2Fadmin%2Fmodel2%2F&id=999999    -> 500
GET /admin/model2/edit/?url=%2Fadmin%2Fmodel2%2F&id=999999       -> 500
GET /admin/model2/details/?url=%2Fadmin%2Fmodel2%2F&id=notanint  -> 500
```

```
flask_admin/model/base.py:2467          details_view -> model = self.get_one(id)
flask_admin/contrib/peewee/view.py:532  get_one      -> self.model.get(**{pk: id})
peewee.py:7508                          raise self.model.DoesNotExist
```

`BaseModelView.get_one()` is annotated `-> T_ORM_MODEL | None`, and every call site in `flask_admin/model/base.py` already handles `None`: `edit_view` (2410), `details_view` (2467) and `delete_view` (2501) flash "Record does not exist." and redirect, and the ajax list update (2741) returns that message directly. The sqla and mongoengine backends return `None`; peewee is the only one that lets the exception escape.

**Fix:** catch `DoesNotExist` in `get_one` and return `None`, matching the documented contract and the other backends. The three requests above now redirect, and following the redirect renders "Record does not exist.".

The module-level `peewee.DoesNotExist` is caught rather than `self.model.DoesNotExist` because each model's `DoesNotExist` subclasses it and the module-level name is present in `types-peewee`, `self.model.DoesNotExist` fails `mypy --strict` with `"type[Model]" has no attribute "DoesNotExist"`.

`test_get_one_missing_record` in `flask_admin/tests/peeweemodel/test_basic.py` covers the details, edit and delete views, the flashed message, and the malformed-id case. It fails on `master` with `Model2DoesNotExist` and passes with this change.

Two notes on scope, both deliberate:

- On sqlite a malformed `id` against an integer primary key simply matches no rows, so `DoesNotExist` covers it. On Postgres the driver would raise a `DataError` instead, which this does not catch. I left that alone rather than importing driver-specific exceptions into the backend, but happy to widen it if you would prefer.
- `QueryAjaxModelLoader.get_one` in `flask_admin/contrib/peewee/ajax.py` has the same shape (`self.model.get(...)` with nothing catching `DoesNotExist`), while its caller `AjaxSelectField._get_data` checks `if model is not None` and the sqla/mongoengine loaders both return `None`. That is the autocomplete endpoint rather than the views this issue is about, so I have left it out of this PR, happy to send it separately.

Closes #2924.


hasansezertasan (approved):
LGTM

hasansezertasan:
Opened #2940 to track the identical `DoesNotExist` gap in `QueryAjaxModelLoader.get_one` (`peewee/ajax.py:57`), as noted in the scope section here. Approving this PR as-is.

PR #2948, opened 2026-08-14
fix(peewee): return None from ajax get_one() when no record matches

### LINKED ISSUE #2940: fix(peewee): QueryAjaxModelLoader.get_one() raises DoesNotExist for a stale/absent pk
### Summary

`QueryAjaxModelLoader.get_one()` in the peewee backend calls `self.model.get(...)`, which raises peewee's `DoesNotExist` when no row matches. Its caller, `AjaxSelectField._get_data`, guards with `if model is not None` — expecting `None`, not an exception — so a form referencing a since-deleted (or otherwise unmatched) pk crashes instead of rendering with no selection.

This is the autocomplete/ajax-select twin of the view-layer bug fixed in #2939 / #2924. That PR fixed `ModelView.get_one()`; this one was left deliberately out of scope there.

### Location

```
flask_admin/contrib/peewee/ajax.py:57   get_one -> self.model.get(**{self.pk: pk})
flask_admin/model/fields.py:195          AjaxSelectField._get_data -> model = self.loader.get_one(...)
flask_admin/model/fields.py:197          if model is not None:   # expects None on miss
```

### Other backends

Both sibling loaders already return `None` on a miss, so peewee is the only one that lets the exception escape:

- sqla — `flask_admin/contrib/sqla/ajax.py:90`: `session.get(self.model, pk)` returns `None`
- mongoengine — `flask_admin/contrib/mongoengine/ajax.py:69`: `self.model.objects.filter(pk=pk).first()` returns `None`

### Proposed fix

Catch the module-level `peewee.DoesNotExist` in `QueryAjaxModelLoader.get_one` and return `None`, mirroring the fix in #2939:

```python
def get_one(self, pk: t.Any) -> t.Any:
    try:
        return self.model.get(**{self.pk: pk})
    except DoesNotExist:
        return None
```

(Module-level `DoesNotExist` rather than `self.model.DoesNotExist` — each model subclasses it, and the module-level name is what passes `mypy --strict`.)

### Scope note

As with #2939, a malformed pk against an integer primary key raises no error on sqlite (`DoesNotExist` covers it) but would raise a driver-specific `DataError` on Postgres. That's left out here too rather than importing driver exceptions into the backend.

DSeaStar:
## Summary

`QueryAjaxModelLoader.get_one()` in the peewee backend called `self.model.get(...)`, which raises `DoesNotExist` when no row matches. `AjaxSelectField._get_data` already guards with `if model is not None`, so a form that still holds a deleted (or otherwise unmatched) pk crashed instead of rendering with no selection.

The sqla and mongoengine loaders already return `None` on a miss, and the same peewee pattern was fixed for `ModelView.get_one()` in #2939. This applies that change to the ajax loader left out of that PR.

```python
def get_one(self, pk):
    try:
        return self.model.get(**{self.pk: pk})
    except DoesNotExist:
        return None
```

Closes #2940.

## Test plan

- [x] `test_ajax_fk` now asserts `loader.get_one(999999) is None`
- [x] Creating a form and feeding a missing pk no longer raises; the field data stays `None`
- [x] Full peewee suite: `pytest flask_admin/tests/peeweemodel/` (13 passed)

### DIFF
--- a/flask_admin/tests/peeweemodel/test_basic.py
+++ b/flask_admin/tests/peeweemodel/test_basic.py
@@ -1077,6 +1077,15 @@ def test_ajax_fk(app: Flask, db: peewee.SqliteDatabase, admin: Admin) -> None:
     mdl = loader.get_one(model.id)  # type: ignore[attr-defined]
     assert mdl.test1 == model.test1
 
+    # a missing pk is not an error: get_one returns None so AjaxSelectField
+    # can render with no selection rather than raising peewee's DoesNotExist
+    assert loader.get_one(999999) is None
+
+    with app.test_request_context("/admin/view/"):
+        form = view.create_form()
+        form.model1.process_formdata([as_unicode(999999)])  # type: ignore[attr-defined]
+        assert form.model1.data is None  # type: ignore[attr-defined]
+
     items = loader.get_list("fir")
     assert len(items) == 1
     assert items[0].id == model.id  # type: ignore[attr-defined, union-attr]
--- a/flask_admin/contrib/peewee/ajax.py
+++ b/flask_admin/contrib/peewee/ajax.py
@@ -1,5 +1,7 @@
 import typing as t
 
+from peewee import DoesNotExist
+
 from flask_admin._compat import as_unicode
 from flask_admin._compat import string_types
 from flask_admin.model.ajax import AjaxModelLoader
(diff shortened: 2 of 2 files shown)


flask_admin/tests/peeweemodel/test_basic.py
```diff
@@ -1077,6 +1077,15 @@ class Model2(BaseModel):
     mdl = loader.get_one(model.id)  # type: ignore[attr-defined]
     assert mdl.test1 == model.test1
 
+    # a missing pk is not an error: get_one returns None so AjaxSelectField
+    # can render with no selection rather than raising peewee's DoesNotExist
+    assert loader.get_one(999999) is None
+
+    with app.test_request_context("/admin/view/"):
```
ElLorans, line 1084: Why do we need this part of the test?

PR #2952, opened 2026-09-12
fix(sqla): coerce primary key values to column types in get_query_for_ids

### LINKED ISSUE #2951: Bulk delete fails with PostgreSQL integer primary key when using psycopg 3
Bulk deletion of records from a flask_admin.contrib.sqla.ModelView fails when using psycopg3 with PostgreSQL and an integer primary key.

The same application works correctly when using psycopg2. The problem appears immediately after switching PG driver from psycopg2 to psycopg.

Single record deletion works perfectly but when deleting multiple records, the application crashes.

The issue occurs across multiple models, so it does not seem to be specific to any one single model.

Sharing my sample model:

```
class BRS(db.Model):
    id: Mapped[IntPK]
    uiic_regional_code: Mapped[str]
    uiic_office_code: Mapped[str]
    financial_year: Mapped[str]
    month: Mapped[str]
```
    
where IntPK is an annotated type as below:

`IntPK = Annotated[int, mapped_column(primary_key=True)]`

## Steps to reproduce
1. Configure flask application with flask-admin with sqlalchemy, flask-sqlalchemy and Postgres.
2. Use an integer PK.
3. Use psycopg 3 as the PG driver.
4. Open a sqla.ModelView list page.
5. Select multiple records.
6. Choose Delete from the bulk actions.
7. Confirm deletion.

## Expected behavior

The selected records should be deleted successfully, as they are when using psycopg2.

## Actual behavior:

The bulk delete fails with:

```
psycopg.errors.UndefinedFunction: operator does not exist:
integer = character varying

LINE 3:
WHERE brs.id IN ($1::VARCHAR, $2::VARCHAR)

HINT:
No operator matches the given name and argument types.
You might need to add explicit type casts.

```
 
## Environment:

Flask-Admin: 2.2.0
SQLAlchemy: 2.0.52
psycopg: 3.3.4
PostgreSQL: 16.15
Python: 3.12
Flask: 3.1.3

## Workaround: 

I am able to get bulk delete to work by converting the ID to primary key data type as below but I am not sure if it should be handled by flask-admin or psycopg itself?

```
@action(
    "delete",
    lazy_gettext("Delete"),
    lazy_gettext("Are you sure you want to delete selected records?"),
)
def action_delete(self, ids: tuple[str, ...]) -> None:
    try:
        pk_column = self.model.__mapper__.primary_key[0]

        python_type = pk_column.type.python_type
        ids = tuple(python_type(value) for value in ids)
        query = tools.get_query_for_ids(
            self.get_query(),
            self.model,
            ids,
        )
```

Grove-ovo:
Fixes #2951

### Problem
In the SQLAlchemy backend, bulk delete passes primary key IDs submitted from HTTP form actions as strings into `get_query_for_ids()`. When running on PostgreSQL with strict drivers like `psycopg` 3, filtering an integer primary key column with string IDs fails with:
```
operator does not exist: integer = character varying
```

### Solution
Coerce string primary key values to their column's `python_type` in `get_query_for_ids()` before constructing the filter. This ensures:
- Values match the column's expected type when binding parameters.
- Boolean primary keys correctly parse boolean strings without misinterpreting `"False"`.
- Custom `TypeDecorator` primary keys preserve their existing bind-parameter processing.
- Composite primary keys validate field counts against model columns.

### Testing & CI
- Consolidated standard primary key bulk delete tests across Integer, String, and UUID into a single parameterized behavioral test.
- Added behavioral tests for Boolean PK, custom TypeDecorator PK, and composite PK with mismatched field count validation.
- Added `psycopg[binary]` to test dependencies in `pyproject.toml` and updated `uv.lock` so PostgreSQL + `psycopg` 3 is automatically tested in CI on SQLAlchemy 2.
- Verified that on PostgreSQL with `psycopg` 3, the bulk delete test fails without this fix and passes with it.

### DIFF
--- a/flask_admin/contrib/sqla/tools.py
+++ b/flask_admin/contrib/sqla/tools.py
@@ -127,6 +127,35 @@ def tuple_operator_in(
         return None
 
 
+def _coerce_pk_value(col: t.Any, val: t.Any) -> t.Any:
+    """
+    Coerce string value from HTTP request into the column's native python_type.
+    Falls back gracefully if python_type is not implemented or coercion fails.
+    """
+    try:
+        type_ = getattr(col, "type", None)
+        try:
+            python_type = getattr(type_, "python_type", None)
+        except NotImplementedError:
+            python_type = None
+
+        if python_type is not None:
+            if isinstance(val, python_type):
+                return val
+            if python_type is bool:
+                if isinstance(val, str):
+                    v = val.strip().lower()
+                    if v in ("true", "1", "t", "yes", "y"):
+                        return True
+                    elif v in ("false", "0", "f", "no", "n"):
+                        return False
+                    return val
+            return python_type(val)
+    except (NotImplementedError, AttributeError, ValueError, TypeError):
+        return val
+    return val
+
+
 def get_query_for_ids(
     modelquery: t.Any, model: type[T_SQLALCHEMY_MODEL], ids: tuple[str, ...]
 ) -> t.Any:
@@ -143,8 +172,22 @@ def get_query_for_ids(
         # Get model primary key property references
         model_pk = [getattr(model, name) for name in get_primary_key(model)]
 
+        coerced_decoded_ids = []
+        for key_tuple in decoded_ids:
+            if len(key_tuple) != len(model_pk):
+                raise ValueError(
+                    f"Mismatch between expected primary key count ({len(model_pk)}) "
+                    f"and input key fields ({len(key_tuple)})."
+                )
+            coerced_decoded_ids.append(
+                tuple(
+                    _coerce_pk_value(col, val)
+                    for col, val in zip(model_pk, key_tuple, strict=True)
+                )
+            )
+
         try:
-            query = modelquery.filter(tuple_(*model_pk).in_(decoded_ids))
+            query = modelquery.filter(tuple_(*model_pk).in_(coerced_decoded_ids))
             # Only the execution of the query will tell us, if the tuple_
             # operator really works
             query.all()
@@ -152,7 +195,7 @@ def get_query_for_ids(
             query = modelquery.filter(
                 tuple_operator_in(
                     model_pk,
-                    decoded_ids,  # type: ignore[arg-type]
+                    coerced_decoded_ids,  # type: ignore[arg-type]
                 )
             )
     else:
@@ -160,7 +203,8 @@ def get_query_for_ids(
             model,
             get_primary_key(model),  # type: ignore[arg-type]
         )
-        query = modelquery.filter(model_pk.in_(ids))
+        coerced_ids = [_coerce_pk_value(model_pk, v) for v in ids]
+        query = modelquery.filter(model_pk.in_(coerced_ids))
 
     return query
 
--- a/flask_admin/tests/sqla/test_basic.py
+++ b/flask_admin/tests/sqla/test_basic.py
@@ -2547,7 +2547,57 @@ def test_multiple_delete(
         assert sqla_db_ext.db.session.query(M1).count() == 0
 
 
-def test_multiple_delete_uuid_pk(
+_u1, _u2, _u3 = uuid.uuid4(), uuid.uuid4(), uuid.uuid4()
+
+
+@pytest.mark.parametrize(
+    ("col_type", "records", "delete_ids", "expected_remaining_id"),
+    [
+        (Integer, [1, 2, 3], ["1", "2"], 3),
+        (String(50), ["s1", "s2", "s3"], ["s1", "s2"], "s3"),
+        (UUIDType(binary=False), [_u1, _u2, _u3], [str(_u1), str(_u2)], _u3),
+    ],
+)
+def test_multiple_delete_standard_pks(
+    app: Flask,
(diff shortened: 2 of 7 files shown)


flask_admin/contrib/sqla/tools.py
```diff
@@ -127,6 +127,35 @@ def tuple_operator_in(
...
 
 
+def _coerce_pk_value(col: t.Any, val: t.Any) -> t.Any:
+    """
+    Coerce string value from HTTP request into the column's native python_type.
+    Falls back gracefully if python_type is not implemented or coercion fails.
+    """
+    try:
+        type_ = getattr(col, "type", None)
+        try:
```
ElLorans: Isn't this try-except block already covered? Looks it can be eliminated

Grove-ovo:
@ElLorans

flask_admin/contrib/sqla/tools.py
Grove-ovo: Removed in 79c38d1f. The outer handler already catches `NotImplementedError` and returns the original value, so the inner try/except was redundant. The existing behavioural tests pass. Thanks for pointing this out.

ElLorans: You can remove the getattr call, the if guard, and the final return, too.

Grove-ovo: Thanks for the suggestion! Simplified in 71c9dd6e: removed the `getattr` calls, the redundant guard, and the final return.

flask_admin/tests/sqla/test_basic.py
```diff
@@ -2539,12 +2540,165 @@ def test_multiple_delete(
...
+            name = Column(String(50))
+
+        sqla_db_ext.create_all()
+
+        sqla_db_ext.db.session.add_all(
+            [DynamicModel(id=val, name=f"name_{val}") for val in records]
+        )
+        sqla_db_ext.db.session.commit()
+
+        param = skip_or_return_session_or_db(sqla_db_ext, session_or_db)
```
ElLorans: we can move this up a few LoC to save time for the skip case

flask_admin/tests/sqla/test_basic.py
```diff
@@ -2539,12 +2540,165 @@ def test_multiple_delete(
...
+
+        sqla_db_ext.db.session.add_all(
+            [
+                BoolModel(id=False, name="f_row"),
+                BoolModel(id=True, name="t_row"),
+            ]
+        )
+        sqla_db_ext.db.session.commit()
+
+        param = skip_or_return_session_or_db(sqla_db_ext, session_or_db)
```
ElLorans: We can move this up a few LoCs to save time in the skip case

Grove-ovo: Thanks! Moved the skip checks before model and database setup in 12f63161, including the same pattern in three other tests, to avoid unnecessary setup for unsupported combinations. All related regression tests pass.

ElLorans:
Sorry for the back and forth, I write down when I think of something, feel free to take a few days before replying/acting upon.
1. If the problem is in strict drivers like psycopg3, why don't we test only in test_postgres? The other tests should pass even in master, don't they?
2. The checks for psycopg look unnecessary. Why do we need them?

Grove-ovo:
@ElLorans Thanks for the feedback! I've narrowed the tests and simplified the checks:

- Kept the PostgreSQL/psycopg 3 integer regression, which fails with the original query implementation and passes with the fix.
- Removed the additional general Integer/String/UUID test and the duplicate default-driver integer test to reduce overlapping coverage.
- Moved the boolean and TypeDecorator cases to test_postgres.py. These protect against coercion selecting the wrong boolean record or interfering with custom ID binding.
- Kept focused composite-key coverage in test_postgres.py because this PR also changes that query path. It checks that unselected records remain and malformed IDs cannot be truncated into a valid key.
- Removed the psycopg import check: missing a declared test dependency should fail rather than skip.
- Replaced dialect probing with the existing HAS_SQLALCHEMY_2 flag, since the project still tests SQLAlchemy 1.4, which does not support psycopg 3.

The production fix and existing test cases are unchanged by this revision. The related regression suite passes, and I verified the expected skips in an actual SQLAlchemy 1.4.18 environment.

ElLorans:
Other 2 questions:
1. why do we change the query path?
2. can we stop checking for psycopg3 and just upgrade the dependencies in uv.lock? The tests must pass on both psycopg2 and 3, so no need for that
3. Do we really need to convert "t", "yes", "y" to True? When would that be needed?

Grove-ovo:
Thanks for the review! Updated in 9aab7ec2:

1. The composite-key changes were unnecessary for this issue. I've restored the original query path and limited coercion to single-column primary keys.
2. Removed the driver-specific checks and fixture. The tests now reuse the existing PostgreSQL fixture, with psycopg 3 declared in the test dependencies and locked in uv.lock.
3. Removed the boolean aliases. Conversion now handles only "True" and "False", which the admin interface generates.

I also reverted the workflow override after it affected unrelated tests. All 12 existing CI checks pass with the default driver, psycopg2. I verified the related tests locally with both drivers, including confirming that the integer regression fails without the fix under psycopg 3.

The PR description now reflects the current scope and validation.

PR #2957, opened 2026-09-21
Support building with flit_core v4

### LINKED ISSUE #2956: Support building with flit_core v4?
Hi, thanks for all the work keeping Flask-Admin going under pallets-eco — the 2.x releases have been great to package downstream.

I ran into a small snag while building `v2.2.1` against the `flit_core` 4.x that distros now ship (Arch has 4.1.0), and I wanted to check whether lifting the v4 cap is on the roadmap.

## What I hit

`pyproject.toml` still pins the backend:

```toml
[build-system]
requires = ["flit_core<4"]
build-backend = "flit_core.buildapi"
```

So an offline / `--no-isolation` build with the system backend fails twice: first `python-build` declines (`wanted: <4, found: 4.1.0`), and past that, `flit_core` 4 raises:

```
flit_core.config.ConfigError: Unrecognised key(s) in [project] table: author
```

That comes from line 7, which uses the singular key:

```toml
author = [{ name = "Flask-Admin team" }]   # PEP 621 wants `authors`
```

Older `flit_core` tolerated it; v4 validates strictly.

## Good news

It looks like a two-line fix. With just `author` → `authors`, I built the `v2.2.1` sdist into a wheel cleanly under `flit_core 4.1.0` — nothing else in `[project]` trips validation, and `[tool.flit.module]` (just the `flask_admin` import name) is untouched by v4's removal of `[tool.flit.metadata]`. That lines up with flit's v4 changelog note that `[project]`-table projects "can simply change the `flit_core <4` constraint to `<5`".

```diff
-author = [{ name = "Flask-Admin team" }]
+authors = [{ name = "Flask-Admin team" }]
```

```diff
 [build-system]
-requires = ["flit_core<4"]
+requires = ["flit_core<5"]
 build-backend = "flit_core.buildapi"
```

I also noticed Dependabot PR #2947 (open) already moves the dev dependency to `<5`, so the `[build-system]` pin looks like the last holdout.

## Why I'm asking

I maintain the Arch AUR `python-flask-admin` package, where this currently needs a `sed` + `--skip-dependency-check` workaround in `build()` — perfectly livable, but I'd love to drop it once upstream supports v4.

Totally understand if you'd want CI to exercise flit 4 first (I see the `exclude-newer` pin in the uv config). Happy to put the above up as a PR if that would help — just say the word. Thanks again!

stiermid:
Fixes #2956.

Distros now ship flit_core 4.x, but the build-system pin capped at <4, and the singular `author` key in [project] (invalid per PEP 621, tolerated by v3) makes flit_core v4 raise `ConfigError: Unrecognised key(s) in [project] table: author`.

Changes:
- `pyproject.toml`: `author` -> `authors`, `flit_core<4` -> `flit_core<5` (per the flit v4 changelog guidance for [project]-table projects).
- `doc/changelog.rst`: [unreleased] entry.

Verification:
- `python -m build --no-isolation` with system flit_core 4.1.0: builds sdist + wheel successfully.
- Isolated `python -m build`: resolves flit-core==4.1.0 and succeeds.
- `twine check`: PASSED on both artifacts; Author metadata correct.
- Local pytest (excluding service-backed azure/mongo/postgres/redis/s3/geoa tests, which need live infra): 365 passed, 81 skipped, 13 xfailed.
- No new pytest test added: no runtime code changed, and no unit test can cover pyproject metadata; CI's wheel builds exercise this.


samuelhwilliams (approved):
Thank you!

